PR #48, opened 2023-05-31
Authentication sigv4

mlt180:
The authentication flow is the following.
Added 2 flags on running the cmd -login, -password, which are the root user login and password. The authentication is the s3 SigV4 authentication. In the authentication middlware it's used SignHTTP method from aws sdk v4 to calculate the signature and compare. After validating the signature user grants access to all the resources.

### DIFF
--- a/cmd/versitygw/main.go
+++ b/cmd/versitygw/main.go
@@ -144,7 +144,7 @@ func runGateway(be backend.Backend) error {
 		opts = append(opts, s3api.WithTLS(cert))
 	}
 
-	srv, err := s3api.New(app, be, port, middlewares.AdminUser{AdminAccess: adminAccess, AdminSecret: adminSecret}, opts...)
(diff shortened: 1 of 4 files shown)


s3api/utils/utils.go
```diff
@@ -32,3 +38,50 @@ func GetUserMetaData(headers *fasthttp.RequestHeader) (metadata map[string]strin
...
+	if *loginPtr == "" || *passwordPtr == "" {
+		os.Exit(3)
+	}
+
+	rootUser = RootUser{
+		Login:    *loginPtr,
+		Password: *passwordPtr,
+	}
+	return
+}
```
benmcclelland: Lets just use the ones that the cli is already providing instead of using new flags here.
I don't think GetRootUserCreds() is needed at all since the cli already has these.

PR #68, opened 2023-06-08
feat: Added admin CLI, created api endpoint for creating new user, cr…

mlt180:
I created an admin CLI for admin related actions(users CRUD...).  (All CRUD operations are not ready yet, will add them further).
Created api endpoint for creating new users.
Created IAM config json file for users storage.
Changed the authentication middleware to verify the user from storage.
Removed root user authentication flow and added the root user creds to admin cli.

### DIFF
--- a/backend/auth/iam.go
+++ b/backend/auth/iam.go
@@ -18,58 +18,97 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"sync"
 
 	"github.com/versity/versitygw/s3err"
 )
 
 type Account struct {
-	Access string `json:"access"`
 	Secret string `json:"secret"`
 	Role   string `json:"role"`
 	Region string `json:"region"`
 }
 
 type IAMConfig struct {
-	AccessAccounts []Account `json:"accessAccounts"`
+	AccessAccounts map[string]Account `json:"accessAccounts"`
+}
+
+type AccountsCache struct {
+	mu       sync.Mutex
+	Accounts map[string]Account
+}
+
+func (c *AccountsCache) getAccount(access string) *Account {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	acc, ok := c.Accounts[access]
+	if !ok {
+		return nil
+	}
+
+	return &acc
+}
+
+func (c *AccountsCache) updateAccounts() error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	var data IAMConfig
+
+	file, err := os.ReadFile("users.json")
+	if err != nil {
+		return fmt.Errorf("error reading config file: %w", err)
+	}
+
+	if err := json.Unmarshal(file, &data); err != nil {
+		return fmt.Errorf("error parsing the data: %w", err)
+	}
+
+	c.Accounts = data.AccessAccounts
+
+	return nil
 }
 
 type IAMService interface {
 	GetIAMConfig() (*IAMConfig, error)
-	CreateAccount(account *Account) error
+	CreateAccount(access string, account *Account) error
 	GetUserAccount(access string) *Account
 }
 
-type IAMServiceUnsupported struct{}
+type IAMServiceUnsupported struct {
+	accCache *AccountsCache
+}
 
 var _ IAMService = &IAMServiceUnsupported{}
 
 func New() IAMService {
-	return &IAMServiceUnsupported{}
+	return &IAMServiceUnsupported{accCache: &AccountsCache{Accounts: map[string]Account{}}}
 }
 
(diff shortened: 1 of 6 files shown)


s3api/middlewares/authentication.go
```diff
@@ -77,9 +72,9 @@ func VerifyV4Signature(config AdminConfig, iam auth.IAMService, debug bool) fibe
 		}
 		signedHdrs := strings.Split(signHdrKv[1], ";")
 
-		secret, ok := acct.getAcctSecret(creds[0])
-		if !ok {
-			return controllers.Responce[any](ctx, nil, s3err.GetAPIError(s3err.ErrInvalidAccessKeyID))
+		userAccount := iam.GetUserAccount(creds[0])
+		if userAccount == nil {
+			return controllers.Responce[any](ctx, nil, s3err.GetAPIError(s3err.ErrAccessDenied))
```
benmcclelland: the error for an unknown account is s3err.ErrInvalidAccessKeyID

cmd/versitygw/admin.go
```diff
@@ -0,0 +1,147 @@
...
+	if role != "admin" && role != "user" {
+		return fmt.Errorf("invalid input parameter for role")
+	}
+
+	req, err := http.NewRequest(http.MethodPost, fmt.Sprintf("http://localhost:7070/create-user?access=%v&secret=%v&role=%v&region=%v", access, secret, role, region), nil)
+	if err != nil {
+		return fmt.Errorf("failed to send the request: %w", err)
+	}
+
+	fmt.Printf("%v\n", req)
```
benmcclelland: leftover debugging?

cmd/versitygw/admin.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to send the request: %w", err)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("%s", body)
```
benmcclelland, line 142: more debug?

cmd/versitygw/main.go
```diff
@@ -155,12 +134,7 @@ func runGateway(be backend.Backend) error {
 		opts = append(opts, s3api.WithDebug())
 	}
 
-	srv, err := s3api.New(app, be, port,
-		middlewares.AdminConfig{
-			AdminAccess: adminAccess,
-			AdminSecret: adminSecret,
-			Region:      region,
-		}, auth.IAMServiceUnsupported{}, opts...)
+	srv, err := s3api.New(app, be, port, auth.IAMServiceUnsupported{}, opts...)
```
benmcclelland: I think this will break the world to lose the admin account and only pass in auth.IAMServiceUnsupported here

cmd/versitygw/main.go
```diff
@@ -25,14 +25,10 @@ import (
...
 	"github.com/versity/versitygw/backend/auth"
 	"github.com/versity/versitygw/s3api"
-	"github.com/versity/versitygw/s3api/middlewares"
 )
 
 var (
 	port              string
-	adminAccess       string
-	adminSecret       string
-	region            string
```
benmcclelland: If we lose the main admin user, how do we create new accounts? also, I think we should keep the main admin user so that no IAM config is needed if people want to run in single user (admin only) mode.

backend/auth/iam.go
```diff
@@ -35,3 +50,60 @@ func New() IAMService {
 func (IAMServiceUnsupported) GetIAMConfig() (*IAMConfig, error) {
 	return nil, s3err.GetAPIError(s3err.ErrNotImplemented)
 }
+
+func (s IAMServiceUnsupported) CreateAccount(account *Account) error {
+	var data IAMConfig
+
+	fmt.Printf("%+v\n", account)
```
benmcclelland: debugging?

backend/auth/iam.go
```diff
@@ -14,14 +14,29 @@
...
+type Account struct {
+	Access string `json:"access"`
+	Secret string `json:"secret"`
+	Role   string `json:"role"`
+	Region string `json:"region"`
+}
 
 type IAMConfig struct {
-	AccessAccounts map[string]string
+	AccessAccounts []Account `json:"accessAccounts"`
```
benmcclelland: I think this would be better as a map since the requests rate could be high.  We would note that this should be cached on the backend too for the same reason.

cmd/versitygw/admin.go
mlt180: This is for printing out the error or success messages

benmcclelland: oh, right, i forgot that some of these are interactive

PR #69, opened 2023-06-09
Ben/dir objects

benmcclelland:


backend/walk.go
```diff
@@ -76,8 +83,35 @@ func Walk(fileSystem fs.FS, prefix, delimiter, marker string, max int) (WalkResu
 				return fs.SkipDir
 			}
 
-			// TODO: special case handling if directory is empty
-			// and was "PUT" explicitly
+			// TODO: can we do better here rather than a second readdir
+			// per directory?
```
mlt180, line 87: I think it's okay for now, but leave the comment here, we can think of something better further

s3api/controllers/base.go
```diff
@@ -252,8 +256,8 @@ func (c S3ApiController) PutActions(ctx *fiber.Ctx) error {
 	metadata := utils.GetUserMetaData(&ctx.Request().Header)
 
 	res, err := c.be.PutObject(&s3.PutObjectInput{
-		Bucket:        &dstBucket,
-		Key:           &dstKeyStart,
+		Bucket:        &bucket,
+		Key:           &keyStart,
```
mlt180, line 260: Looks like the pipeline fails on line 370. Double check?

benmcclelland: these are the govulncheck for the stdlib, will be fixed automatically in newer go version once actions picks that up

PR #72, opened 2023-06-10
Ben/posix multipart

benmcclelland:


s3api/controllers/base.go
```diff
@@ -242,18 +221,16 @@ func (c S3ApiController) PutActions(ctx *fiber.Ctx) error {
 	}
 
 	if copySource != "" {
+		// TODO: copy object
+		_, _, _, _ = copySrcIfMatch, copySrcIfNoneMatch,
+			copySrcModifSince, copySrcUnmodifSince
```
mlt180: Copy object action is handled, WDYM?

PR #101, opened 2023-06-19
Ben/auth iam

benmcclelland:


backend/auth/acl.go
```diff
@@ -141,12 +141,15 @@ func UpdateACL(input *s3.PutBucketAclInput, acl ACL, iam IAMConfig) error {
...
 }
 
-func checkIfAccountsExist(accs []string, iam IAMConfig) ([]string, error) {
+func checkIfAccountsExist(accs []string, iam IAMService) ([]string, error) {
 	result := []string{}
 
 	for _, acc := range accs {
-		_, ok := iam.AccessAccounts[acc]
-		if !ok {
+		_, err := iam.GetUserAccount(acc)
```
mlt180, line 152: Reading the IAM configuration file for each account is an expensive operation. Instead I'd recommend implementing an IAM method, which will take []string and check if they exist

jonaustin09 (dismissed):

backend/auth/acl.go
benmcclelland: Let me think about this one for a new ticket.  This was in the original code that was refactored too.  The benefit about tossing this to the backend would be to avoid grabbing the lock multiple times and validating the data crc multiple times.  But I want to be careful about premature optimization here since we are probably only ever talking about a handful of accounts here. Also, the more we push to the backend, the more we require each IAM service to implement.

jonaustin09 (approved):

PR #129, opened 2023-07-10
feat: Set up audit logging basic structure, set up webhook logger, bu…

jonaustin09:
Set up audit logging interface.
Implement audit logging with webhook url.
Bug fix in posix DeleteObject function. Not existing object didn't return error.

### DIFF
--- a/s3api/controllers/base.go
+++ b/s3api/controllers/base.go
@@ -17,7 +17,6 @@ package controllers
 import (
 	"bytes"
 	"encoding/xml"
-	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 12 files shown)


s3api/controllers/base.go
```diff
@@ -765,27 +773,35 @@ func (c S3ApiController) CreateActions(ctx *fiber.Ctx) error {
...
+	Logger s3log.AuditLogger
+	Meta   s3log.LogMeta
 }
 
-func SendResponse(ctx *fiber.Ctx, err error) error {
+func SendResponse(ctx *fiber.Ctx, err error, l *LogOptions) error {
 	if err != nil {
+		if l.Logger != nil {
+			l.Logger.Log(ctx, err, nil, l.Meta)
+		}
```
benmcclelland: move this up one line, and then you can remove the success logger call as well

s3api/controllers/base.go
```diff
@@ -800,15 +816,21 @@ func SendResponse(ctx *fiber.Ctx, err error) error {
...
 	return nil
 }
 
-func SendXMLResponse(ctx *fiber.Ctx, resp any, err error) error {
+func SendXMLResponse(ctx *fiber.Ctx, resp any, err error, l *LogOptions) error {
 	if err != nil {
-		fmt.Println(err)
+		if l.Logger != nil {
+			l.Logger.Log(ctx, err, nil, l.Meta)
+		}
```
benmcclelland, line 829: same here, move up one line, remove success case logger

s3api/controllers/base_test.go
```diff
@@ -1359,6 +1362,8 @@ func TestS3ApiController_CreateActions(t *testing.T) {
 	for _, tt := range tests {
 		resp, err := tt.app.Test(tt.args.req)
 
+		fmt.Println(tt.name)
+
```
benmcclelland: extra output still in here?

s3api/controllers/base.go
mlt180: Here we need the logger separated for error and success cases to get the exact body size

benmcclelland: oh i see, yes, thats fine

benmcclelland (approved):

PR #167, opened 2023-07-27
Issue 156

jonaustin09:
Added bucket name validation on bucket creation:
Here is the [link](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html), where you can find the naming rules.

### DIFF
--- a/s3api/utils/utils.go
+++ b/s3api/utils/utils.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"github.com/gofiber/fiber/v2"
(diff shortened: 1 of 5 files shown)


s3api/utils/utils.go
```diff
@@ -83,6 +84,22 @@ func SetResponseHeaders(ctx *fiber.Ctx, headers []CustomHeader) {
...
+	}
+	// Checks to contain only digits, lowercase letters, dot, hyphen.
+	// Checks to start and end with only digits and lowercase letters.
+	if !regexp.MustCompile(`^[a-z0-9][a-z0-9.-]+[a-z0-9]$`).MatchString(bucket) {
+		return false
+	}
+	// Checks not to be a valid IP address
+	if regexp.MustCompile(`^(?:[0-9]{1,3}\.){3}[0-9]{1,3}$`).MatchString(bucket) {
+		return false
+	}
```
benmcclelland: put regexp.MustCompile(...) into a global variable, and then use var.MatchString(bucket) in here.  this will speed up the check to only compile the regexp once.

benmcclelland (changes requested):
move regexp compile to global variable

benmcclelland (approved):

PR #169, opened 2023-07-27
Issue 154

jonaustin09:
Changed error from InvalidRequest to InvalidRange for GetObject invalid range error

### DIFF
--- a/s3err/s3err.go
+++ b/s3err/s3err.go
@@ -106,9 +106,9 @@ const (
 	ErrNotImplemented
 	ErrPreconditionFailed
 	ErrInvalidObjectState
-
 	ErrExistingObjectIsDirectory
 	ErrObjectParentIsFile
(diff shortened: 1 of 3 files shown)


s3err/s3err.go
```diff
@@ -106,9 +106,9 @@ const (
 	ErrNotImplemented
 	ErrPreconditionFailed
 	ErrInvalidObjectState
-
 	ErrExistingObjectIsDirectory
 	ErrObjectParentIsFile
```
benmcclelland, line 113: these were separated out because they are non-standard AWS errors.  lets keep those separated just so we keep track of these. they probably could have used a comment indicating why they were separate.

benmcclelland (changes requested):
keep non-standard aws errors separated out

benmcclelland (approved):

PR #173, opened 2023-07-31
Issu 168

jonaustin09:


backend/posix/posix.go
```diff
@@ -946,7 +946,7 @@ func (p *Posix) PutObject(ctx context.Context, po *s3.PutObjectInput) (string, e
 
 	err = f.link()
 	if err != nil {
-		return "", fmt.Errorf("link object in namespace: %w", err)
+		return "", s3err.GetAPIError(s3err.ErrExistingObjectIsDirectory)
```
benmcclelland, line 949: I wonder if there are other errors that could happen here? Maybe we should specifically check for EISDIR (is this the error thats happening here in the failure case?)

benmcclelland (changes requested):

backend/posix/posix.go
mlt180: Here we can't detect the exact error type, as f.link() returns only some custom errors depending on OS. The other option could be walking through the implementations for all the platforms and change the error type there

benmcclelland: ok, we can add this for now, and decide if we need to handle more errors later

benmcclelland (approved):

PR #265, opened 2023-09-21
CopyObject metadata

jonaustin09:
CopyObject copies the object with new metadata provided. Added a comparison step for metadata, if a user tries to copy the object to itself, it is allowed if the metadata has been changed.

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -449,6 +449,20 @@ func loadUserMetaData(path string, m map[string]string) (contentType, contentEnc
 	return
 }
 
+func compareUserMetadata(meta1, meta2 map[string]string) bool {
+	if len(meta1) != len(meta2) {
+		return false
+	}
+
+	for key, val := range meta1 {
+		if meta2[key] != val {
+			return false
+		}
+	}
+
+	return true
+}
+
 func isValidMeta(val string) bool {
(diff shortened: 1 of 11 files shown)


backend/posix/posix.go
```diff
@@ -1294,12 +1304,20 @@ func (p *Posix) CopyObject(ctx context.Context, input *s3.CopyObjectInput) (*s3.
...
 
-	etag, err := p.PutObject(ctx, &s3.PutObjectInput{Bucket: &dstBucket, Key: &dstObject, Body: f, ContentLength: fInfo.Size()})
+	meta := make(map[string]string)
+	loadUserMetaData(objPath, meta)
+
+	dstObjdPath := filepath.Join(dstBucket, dstObject)
+	if dstObjdPath == objPath && compareUserMetadata(meta, input.Metadata) {
+		return &s3.CopyObjectOutput{}, s3err.GetAPIError(s3err.ErrInvalidCopyDest)
+	}
+
```
benmcclelland, line 1323: We still need a check here if source and dest objects are the same.  We could have a request that is only updating metadata. And in that case we want to set the new metadata but not do any data copying.

benmcclelland (changes requested):

benmcclelland (approved):

PR #269, opened 2023-09-27
IAM account interface cleanup

jonaustin09:
Add a new type `InternalAcct` for internal IAM storage. Didn't change IAMService interface to make it more usable for other IAM options(LDAP...)


benmcclelland:
I think all of the relevant cleanup here was done in #270

PR #291, opened 2023-10-27
S3 select event streaming

jonaustin09:
1. Created a struct which handles event streaming and message constructing.
2. The struct handles Continuation event by periodically sending continuation message on every 1 second.
3.  It sends Progress event on every 500 milliseconds if it's enabled by client.
4. It handles Record Event with SendRecord method.
5. FinishWithError method is responsible for error handling: sending an error in a proper way and closing the streaming.
6. Finish method is responsible for sending Stats and End events and closing the streaming.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -20,6 +20,7 @@ import (
 	"io"
 
 	"github.com/aws/aws-sdk-go-v2/service/s3"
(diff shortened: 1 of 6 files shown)


benmcclelland:
closing this pr, we will be revisiting this with a different implementation

PR #296, opened 2023-11-15
feat: Closes #290, implemented request body stream reading for PutObj…

### LINKED ISSUE #290: Investigate enable StreamRequestBody for put/put-part
Instead of buffering the full request in memory before handing off to the backend, we should investigate enabling StreamRequestBody to start sending data to the backend before the request processing completes for larger data puts.

I think this would mean that we would need something that could do the content checksums as data streams too, and either error on the last chunk of data or set the context error that the backend or frontend would have to check before final response. 

The benefit here would be that the backend could start processing the data stream for a large incoming data put, and we could hopefully not have to keep all of the dat in memory at one time.

jonaustin09:
Implemented request body streaming for PutObject and UploadPart actions

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -37,6 +37,7 @@ import (
 	"github.com/pkg/xattr"
 	"github.com/versity/versitygw/auth"
 	"github.com/versity/versitygw/backend"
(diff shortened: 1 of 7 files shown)


benmcclelland:
closing this, and contiuing work in #331

PR #298, opened 2023-11-20
S3 proxy comparison tests

jonaustin09:
Added s3 proxy comparison tests on
1. Upload
2. Download
3. Throughput
benchmarking tests


benmcclelland:
closing this for now, as it doesnt quite do what we need for the perf compare. will revisit if we need.

PR #301, opened 2023-11-27
Dockerize the application

jonaustin09:
Dockerized the application. Created Dockerfiles for dev and prod environments, created a docker compose config file to run the s3 and proxy servers in dev environments with live code update and container recreation. Added commands in Makefile to run s3, proxy and both servers as docker containers


Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+COPY go.mod ./
+RUN go mod download
+
+COPY ./ ./
+
+WORKDIR /app/cmd/versitygw
+RUN go build -o versitygw
+
+RUN mkdir /tmp/vgw
```
kfox1111: Layering here will mean all the stuff done and removed will still be in the final image.

Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
```
kfox1111, line 14: Can probably fix it with a multistage build like:
```suggestion
FROM alpine:latest
COPY --from=0 /app/versitygw /app/versitygw
```

mlt180: I made the changes you mentioned by adding multistage build and the final image size has significantly decreased.
Screenshot 2023-11-27 at 14 10 34

benmcclelland:
It looks like this is just doing a mkdir /tmp/vgw and hosting that via posix gateway from within the container. Wouldn't the more useful case here be to pass in an external volume to be hosted by the container gateway?

kfox1111:
> It looks like this is just doing a mkdir /tmp/vgw and hosting that via posix gateway from within the container. Wouldn't the more useful case here be to pass in an external volume to be hosted by the container gateway?

The Dockerfile stuff is only used to build the container. Persisting the data at runtime is done by mounting a volume in at runtime.

benmcclelland:
> > It looks like this is just doing a mkdir /tmp/vgw and hosting that via posix gateway from within the container. Wouldn't the more useful case here be to pass in an external volume to be hosted by the container gateway?
> 
> The Dockerfile stuff is only used to build the container. Persisting the data at runtime is done by mounting a volume in at runtime.

I meant in the docker compose, we have this:
```
    command: ["sh", "-c", CompileDaemon -build="go build -C ./cmd/versitygw -o versitygw" -command="./cmd/versitygw/versitygw -p :$POSIX_PORT -a $ACCESS_KEY_ID -s $SECRET_ACCESS_KEY --iam-dir /tmp/vgw posix /tmp/vgw"]
```
is hard coding /tmp/vgw really what we want here? I would expect that people would want to run versitygw in the container, but would want the persistent data on some other storage? We have other env vars here, so maybe it could just be another env var to say where the mapped volume is mounted to?

jonaustin09:
> > > It looks like this is just doing a mkdir /tmp/vgw and hosting that via posix gateway from within the container. Wouldn't the more useful case here be to pass in an external volume to be hosted by the container gateway?
> > 
> > 
> > The Dockerfile stuff is only used to build the container. Persisting the data at runtime is done by mounting a volume in at runtime.
> 
> I meant in the docker compose, we have this:
> 
> ```
>     command: ["sh", "-c", CompileDaemon -build="go build -C ./cmd/versitygw -o versitygw" -command="./cmd/versitygw/versitygw -p :$POSIX_PORT -a $ACCESS_KEY_ID -s $SECRET_ACCESS_KEY --iam-dir /tmp/vgw posix /tmp/vgw"]
> ```
> 
> is hard coding /tmp/vgw really what we want here? I would expect that people would want to run versitygw in the container, but would want the persistent data on some other storage? We have other env vars here, so maybe it could just be another env var to say where the mapped volume is mounted to?



> > > It looks like this is just doing a mkdir /tmp/vgw and hosting that via posix gateway from within the container. Wouldn't the more useful case here be to pass in an external volume to be hosted by the container gateway?
> > 
> > 
> > The Dockerfile stuff is only used to build the container. Persisting the data at runtime is done by mounting a volume in at runtime.
> 
> I meant in the docker compose, we have this:
> 
> ```
>     command: ["sh", "-c", CompileDaemon -build="go build -C ./cmd/versitygw -o versitygw" -command="./cmd/versitygw/versitygw -p :$POSIX_PORT -a $ACCESS_KEY_ID -s $SECRET_ACCESS_KEY --iam-dir /tmp/vgw posix /tmp/vgw"]
> ```
> 
> is hard coding /tmp/vgw really what we want here? I would expect that people would want to run versitygw in the container, but would want the persistent data on some other storage? We have other env vars here, so maybe it could just be another env var to say where the mapped volume is mounted to?

I can make the setup directory configurable from env vars

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+    command: ["sh", "-c", CompileDaemon -build="go build -C ./cmd/versitygw -o versitygw" -command="./cmd/versitygw/versitygw -p :$POSIX_PORT -a $ACCESS_KEY_ID -s $SECRET_ACCESS_KEY --iam-dir /tmp/vgw posix /tmp/vgw"]
+  proxy:
+    build: 
+      context: .
+      dockerfile: ./Dockerfile.dev
+    volumes:
+      - ./:/app
+    ports:
+      - "${PROXY_PORT}:${PROXY_PORT}"
+    command: ["sh", "-c", CompileDaemon -build="go build -C ./cmd/versitygw -o versitygw" -command="./cmd/versitygw/versitygw -a $ACCESS_KEY_ID -s $SECRET_ACCESS_KEY -p :$PROXY_PORT s3 --endpoint http://posix:$POSIX_PORT"]
```
kfox1111, line 23: Tangential question as I don't see it in the docs. Whats the difference between proxy and posix? Can you scale both up independently?

benmcclelland: s3proxy is missing in the docs right now. We were working on finalizing functionality before documenting and having people use it.  The backend storage for the gateway is an s3 server for the s3proxy (vs a filesystem for posix).  The original use case was for us to be able to compare performance of direct to s3 storage vs going through the gateway. But there has been interest in using the s3 backend to manage local IAM accounts and have a single backend storage account. This way a site admin could have a single account on an s3 service, and then manage access with local IAM accounts.

kfox1111: Ah. Gotcha. Thanks.

PR #324, opened 2023-12-01
Admin CLI actions in S3 Proxy

jonaustin09:


backend/backend.go
```diff
@@ -69,7 +70,10 @@ type Backend interface {
 	DeleteObjectTagging(_ context.Context, bucket, object string) error
 
 	// non AWS actions
-	ChangeBucketOwner(_ context.Context, bucket, newOwner string) error
+	CreateUser(_ context.Context, iam auth.IAMService, user auth.Account) error
+	DeleteUser(_ context.Context, iam auth.IAMService, access string) error
+	ListUsers(_ context.Context, iam auth.IAMService) ([]auth.Account, error)
```
benmcclelland, line 75: we dont want the backend service to know about user management. this should only be handled by the iam service.

benmcclelland (changes requested):

PR #326, opened 2023-12-04
Ben/select object content

benmcclelland:


s3select/message-handler.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	if mh.getProgress != nil {
+		bytesScanned, bytesProcessed := mh.getProgress()
+
+		err := mh.write(genStatsMessage(bytesScanned, bytesProcessed, mh.bytesReturned))
+		if err != nil {
+			return err
+		}
+	}
```
mlt180: Stats event has to be independent from progress. It's always sent before End message no matter progress is enabled or not

benmcclelland: we only collect bytesScanned, bytesProcessed if progress enabled. is it sufficient to report these as 0 and only have valid bytesReturned if progress disabled?

alternatively, I guess we can accept bytesScanned, bytesProcessed as args to Finish() if these are required.

benmcclelland: ok, looking at the spec, I think it is best to change Finish() to add in bytesScanned, bytesProcessed so that backend is forced to provide these

jonaustin09 (approved):

PR #331, opened 2023-12-09
Ben/req body streaming

benmcclelland:


s3api/utils/auth-reader.go
```diff
@@ -0,0 +1,254 @@
...
+		if debug {
+			options.LogSigning = true
+			options.Logger = logging.NewStandardLogger(os.Stderr)
+		}
+	})
+	if signErr != nil {
+		return s3err.GetAPIError(s3err.ErrInternalError)
+	}
+
+	genAuth, err := ParseAuthorization(req.Header.Get("Authorization"))
```
mlt180, line 142: Do we need to use ParseAuthorization method here, if we only need to get the signature from the signed request and compare?

benmcclelland: we can store the signature from the initial check, this was just a way to validate the signature without having to store it anywhere

benmcclelland: CheckValidSignature is currently also called from the non-streaming requests to try to not have duplicated code around

jonaustin09 (approved):

PR #338, opened 2023-12-13
S3 proxy ListParts max parts issue

jonaustin09:
Fixed max-parts property issue in ListParts action to pass nil if the property is not provided
Added .vscode in gitignore(didn't want to create another PR for this)


s3api/controllers/base.go
```diff
@@ -92,7 +92,7 @@ func (c S3ApiController) GetActions(ctx *fiber.Ctx) error {
 	}
 
 	if uploadId != "" {
-		if maxParts < 0 || (maxParts == 0 && ctx.Query("max-parts") != "") {
+		if (maxParts < 0 && ctx.Request().URI().QueryArgs().Has("max-parts")) || maxParts > 2147483647 {
```
benmcclelland: better to use math.MaxInt32 instead of 2147483647 so its clear where this number comes from

mlt180: I removed this check: it's unnecessary. If max-parts is larger than math.MaxInt32 it will result the variable value to be -1 and it will return the error

benmcclelland (approved):

PR #343, opened 2023-12-16
chore: clarify s3 backend struct with better name

benmcclelland:


benmcclelland:
no logic change here, just cleaning up symbol names

PR #371, opened 2024-01-10
fix: cleanup backend ACLs

benmcclelland:
This adds the default ACL to the CreateBucket backend method so that the backend doesn't need to know how to construct and ACL.

This also moves the s3proxy ACLs to a tag key/value because the gateway ACLs are not the same accounts as the backend s3 server. TODO: we may need to mask this tag key/value if we add support for the Get/PutBucketTagging API.

### DIFF
--- a/backend/s3proxy/s3.go
+++ b/backend/s3proxy/s3.go
@@ -17,6 +17,7 @@ package s3proxy
 import (
 	"context"
 	"crypto/sha256"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
@@ -32,11 +33,12 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/aws/smithy-go"
-	"github.com/versity/versitygw/auth"
 	"github.com/versity/versitygw/backend"
 	"github.com/versity/versitygw/s3err"
 	"github.com/versity/versitygw/s3response"
 )
+
+const aclKey string = "versitygwAcl"
 
 type S3Proxy struct {
 	backend.BackendUnsupported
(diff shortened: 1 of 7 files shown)


backend/s3proxy/s3.go
```diff
@@ -342,53 +343,56 @@ func (s *S3Proxy) DeleteObjects(ctx context.Context, input *s3.DeleteObjectsInpu
...
+	return []byte{}, nil
 }
 
 func (s *S3Proxy) PutBucketAcl(ctx context.Context, bucket string, data []byte) error {
-	acl, err := auth.ParseACL(data)
+	tagout, err := s.client.GetBucketTagging(ctx, &s3.GetBucketTaggingInput{
+		Bucket: &bucket,
+	})
+	err = handleError(err)
 	if err != nil {
```
mlt180: But these 2 actions are actually not supported in our gateway: GetBucketTagging, PutBucketTagging.

s3api/controllers/base.go
```diff
@@ -407,10 +408,16 @@ func (c S3ApiController) PutBucketActions(ctx *fiber.Ctx) error {
...
 	}
 
-	err := c.be.CreateBucket(ctx.Context(), &s3.CreateBucketInput{
+	defACL := auth.ACL{ACL: "private", Owner: acct.Access, Grantees: []auth.Grantee{}}
+	jsonACL, err := json.Marshal(defACL)
+	if err != nil {
+		return SendResponse(ctx, fmt.Errorf("marshal acl: %w", err), &MetaOpts{Logger: c.logger, Action: "CreateBucket", BucketOwner: acct.Access})
+	}
+
+	err = c.be.CreateBucket(ctx.Context(), &s3.CreateBucketInput{
```
mlt180, line 417: I think we have to support here to add bucket ACL on bucket creation, because the CLI and SDKs have this feature to define bucket ACL on creation. Maybe this could be a separate task?

jonaustin09 (approved):

s3api/controllers/base.go
benmcclelland: yeah, lets add a new issue for this

backend/s3proxy/s3.go
benmcclelland: I think this means that we cant use our gateway as a backend for s3proxy yet.  we will need to think about if we want to support this stacking.

mlt180: I think we can add these 2 actions support in posix. That wouldn't be a big deal for us

PR #379, opened 2024-01-19
Fix/azure list objects

jonaustin09:


backend/azure/azure.go
```diff
@@ -301,13 +320,28 @@ func (az *Azure) ListObjectsV2(ctx context.Context, input *s3.ListObjectsV2Input
...
 			return nil, azureErrToS3Err(err)
 		}
 		for _, v := range resp.Segment.BlobItems {
+			if nextMarker == nil && *resp.NextMarker != "" {
+				nextMarker = resp.NextMarker
+				isTruncated = true
+			}
+			if len(objects) >= int(maxKeys) {
+				breakFlag = true
+				break
```
benmcclelland: Use the label break instead of breakFlag, for example:
```
Loop:
for outer() {
    for inner() {
        break Loop
    }
}
```

benmcclelland (changes requested):

benmcclelland (approved):

PR #380, opened 2024-01-21
feat: add chunked upload support

benmcclelland:
As described in
https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html this adds support for reading from a chunked upload encoded request body. The chunked reader modifies the data stream to remove the chunk encoding while validating the chunk signatures in line. This allows the upper layers to get just the object data stream.

### DIFF
--- a/s3api/middlewares/chunk.go
+++ b/s3api/middlewares/chunk.go
@@ -1,4 +1,4 @@
-// Copyright 2023 Versity Software
+// Copyright 2024 Versity Software
 // This file is licensed under the Apache License, Version 2.0
 // (the "License"); you may not use this file except in compliance
 // with the License.  You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


s3api/middlewares/chunk.go
```diff
@@ -0,0 +1,78 @@
...
+			return sendResponse(ctx, err, logger)
+		}
+
+		account, err := acct.getAccount(authData.Access)
+		if err == auth.ErrNoSuchUser {
+			return sendResponse(ctx, s3err.GetAPIError(s3err.ErrInvalidAccessKeyID), logger)
+		}
+		if err != nil {
+			return sendResponse(ctx, err, logger)
+		}
```
mlt180: Instead of calling getAccount method, which is an underlying IAM call, you can get the user account from fiber context.

s3api/middlewares/chunk.go
```diff
@@ -0,0 +1,78 @@
...
+
+		amzdate := ctx.Get("X-Amz-Date")
+		if amzdate == "" {
+			return sendResponse(ctx, s3err.GetAPIError(s3err.ErrMissingDateHeader), logger)
+		}
```
mlt180: This is a double check, it gets checked in the previous middleware(authentication).

s3api/middlewares/chunk.go
```diff
@@ -0,0 +1,78 @@
...
+
+		date, err := time.Parse(iso8601Format, amzdate)
+		if err != nil {
+			return sendResponse(ctx, s3err.GetAPIError(s3err.ErrMalformedDate), logger)
+		}
```
mlt180: The same here, this method can't return an error

PR #411, opened 2024-02-21
IAM integration tests

jonaustin09:
1. Added IAM integration tests
2. Added iam and posix tests in github pipelines

### DIFF
--- a/cmd/versitygw/admin.go
+++ b/cmd/versitygw/admin.go
@@ -289,11 +289,14 @@ func listUsers(ctx *cli.Context) error {
 	}
 	defer resp.Body.Close()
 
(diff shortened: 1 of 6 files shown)


cmd/versitygw/admin.go
```diff
@@ -289,11 +289,15 @@ func listUsers(ctx *cli.Context) error {
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode >= 400 {
+		fmt.Printf("%s", body)
+		return nil
```
benmcclelland: dont we want to return an error here so that the program exits with an error status?

benmcclelland (approved):

PR #427, opened 2024-02-29
Add additional environment variables for configuration properties

static-moonlight:
- New environment variables for all parameters which didn't have any
- All new environment variables have the prefix `VGW_`
- Already existing environment variables have been left unchanged


benmcclelland:
This looks good to me. We probably don't want to break any current env vars, but the list is to allow multiple for the same setting. So we could add some more if we want to have the VGW_ prefix for all. I think the boolean values can be set to true, but I'll double check that when updating the docs.

Thanks for the contribution! I'll get this merged when tests pass.

PR #449, opened 2024-03-12
fix: remove namespace restrictions on tag xml input

### LINKED ISSUE #447: Attempt to set bucket tags with mc tool returns "Invalid request."
**Describe the bug**
When running the command `mc --insecure tag set <alias>/<bucket> <key>=<value>` via the versitygw app, something similar to the following is returned:
`mc: <ERROR> Failed to set tags for https://127.0.0.1:7070/<bucket>. Invalid Request.`

This error is not seen when using **mc** to communicate with S3 directly, or when the same command is run via **aws cli**.

**To Reproduce**
1.  Run versitygw
2. Create a bucket
3. Create an mc alias pointing to the bucket
4. Attempt to set tag with the command above

**Expected behavior**
Tags are set, then are able to be retrieved with **tag list**, or the corresponding **aws cli** command.

**Server Version**
 output of
```
./versitygw -version
uname -a
```
```
Version  : v0.15
Build    : c9c05b4
BuildTime: 2024-03-08_07:57:11PM
Darwin Lukes-MacBook-Pro.local 23.2.0 Darwin Kernel Version 23.2.0: Wed Nov 15 21:54:51 PST 2023; root:xnu-10002.61.3~2/RELEASE_ARM64_T6030 arm64
```

**Additional context**
```
mc version RELEASE.2024-03-03T00-13-08Z (commit-id=98af07b69ce564bec48c5a9edc6d080679ee1c13)
Runtime: go1.21.7 darwin/arm64
Copyright (c) 2015-2024 MinIO, Inc.
License GNU AGPLv3 <https://www.gnu.org/licenses/agpl-3.0.html>
```

benmcclelland:
Fixes #447. Previously we required XML namespace and got these errors with this input:
DEBUG:  <Tagging><TagSet><Tag><Key>mykey</Key><Value>myvalue</Value></Tag></TagSet></Tagging>
DEBUG: expected element <Tagging> in name space http://s3.amazonaws.com/doc/2006-03-01/ but have no name space

### DIFF
--- a/s3response/s3response.go
+++ b/s3response/s3response.go
@@ -112,6 +112,10 @@ type Tagging struct {
 	TagSet  TagSet   `xml:"TagSet"`
 }
 
+type TaggingInput struct {
+	TagSet TagSet `xml:"TagSet"`
+}
+
 type DeleteObjects struct {
 	Objects []types.ObjectIdentifier `xml:"Object"`
 }
--- a/s3api/controllers/base.go
+++ b/s3api/controllers/base.go
@@ -615,7 +615,7 @@ func (c S3ApiController) PutBucketActions(ctx *fiber.Ctx) error {
 	if ctx.Request().URI().QueryArgs().Has("tagging") {
 		parsedAcl := ctx.Locals("parsedAcl").(auth.ACL)
 
-		var bucketTagging s3response.Tagging
+		var bucketTagging s3response.TaggingInput
 		err := xml.Unmarshal(ctx.Body(), &bucketTagging)
 		if err != nil {
 			return SendResponse(ctx, s3err.GetAPIError(s3err.ErrInvalidRequest),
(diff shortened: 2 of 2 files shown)


s3response/s3response.go
```diff
@@ -112,6 +112,10 @@ type Tagging struct {
 	TagSet  TagSet   `xml:"TagSet"`
 }
 
+type TaggingInput struct {
+	TagSet TagSet `xml:"TagSet"`
+}
+
 type DeleteObjects struct {
```
mlt180, line 119: Maybe we better remove Tagging struct above and instead start using this new TaggingInput?

benmcclelland: I think we want to keep with the
```
XMLName xml.Name `xml:"http://s3.amazonaws.com/doc/2006-03-01/ Tagging" json:"-"`
```
for the responses to keep in sync with the aws response
the bug in this one was when the client sent the xml tags without this namespace
the official request syntax also does show the namespace https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketTagging.html
```
PUT /?tagging HTTP/1.1
Host: Bucket.s3.amazonaws.com
Content-MD5: ContentMD5
x-amz-sdk-checksum-algorithm: ChecksumAlgorithm
x-amz-expected-bucket-owner: ExpectedBucketOwner
<Tagging xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
   <TagSet>
      <Tag>
         <Key>string</Key>
         <Value>string</Value>
      </Tag>
   </TagSet>
</Tagging>
```
but I guess not all clients will strictly follow this

jonaustin09 (approved):

PR #468, opened 2024-03-18
fix: require root credentials be set to start gateway

### LINKED ISSUE #466: command does not check and error if root credentials not defined
**Describe the bug**
Command should exit if the root access and secret key are not defined.

**To Reproduce**
```
% ./versitygw posix /tmp/gw
No IAM service configured, enabling single account mode

 ┌───────────────────────────────────────────────────┐ 
 │                     versitygw                     │ 
 │                   Fiber v2.52.2                   │ 
 │               http://127.0.0.1:7070               │ 
 │       (bound on host 0.0.0.0 and port 7070)       │ 
 │                                                   │ 
 │ Handlers ............ 27  Processes ........... 1 │ 
 │ Prefork ....... Disabled  PID ............. 85847 │ 
 └───────────────────────────────────────────────────┘ 

```

**Expected behavior**
Command should exit with error if these are not defined:
```
   --access value, -a value                root user access key [$ROOT_ACCESS_KEY_ID, $ROOT_ACCESS_KEY]
   --secret value, -s value                root user secret access key [$ROOT_SECRET_ACCESS_KEY, $ROOT_SECRET_KEY]
```

**Server Version**
```
% ./versitygw -v           
Version  : v0.16
Build    : f0a00b4
BuildTime: 2024-03-18_08:33:19PM```

### PR DESCRIPTION
**Pull request by benmcclelland**
Fixes #466

### DIFF
--- a/cmd/versitygw/main.go
+++ b/cmd/versitygw/main.go
@@ -369,6 +369,10 @@ func initFlags() []cli.Flag {
 }
 
 func runGateway(ctx context.Context, be backend.Backend) error {
+	if rootUserAccess == "" || rootUserSecret == "" {
+		return fmt.Errorf("root user access and secret key must be provided")
+	}
+
 	app := fiber.New(fiber.Config{
 		AppName:           "versitygw",
 		ServerHeader:      "VERSITYGW",

### CODE REVIEW

File: cmd/versitygw/main.go [hunk 1]
```diff
@@ -369,6 +369,10 @@ func initFlags() []cli.Flag {
 }
 
 func runGateway(ctx context.Context, be backend.Backend) error {
+	if rootUserAccess == "" || rootUserSecret == "" {
+		return fmt.Errorf("root user access and secret key must be provided")
+	}
+
```
mlt180, line 375: Maybe to keep things simple and avoid additional checks we can set `Required` property to true in cli.Flag like below:
```
		&cli.StringFlag{
			Name:        "access",
			Usage:       "root user access key",
			EnvVars:     []string{"ROOT_ACCESS_KEY_ID", "ROOT_ACCESS_KEY"},
			Aliases:     []string{"a"},
			Required:    true,
			Destination: &rootUserAccess,
		},
		&cli.StringFlag{
			Name:        "secret",
			Usage:       "root user secret access key",
			EnvVars:     []string{"ROOT_SECRET_ACCESS_KEY", "ROOT_SECRET_KEY"},
			Aliases:     []string{"s"},
			Required:    true,
			Destination: &rootUserSecret,
		},
```

benmcclelland: The problem with this approach is that we would get an error when running the version option:
```
 % ./versitygw -v
2024/03/19 08:13:38 Required flags "access, secret" not set
```
I dont think this is what we want.

mlt180: Yes you are correct. There are other cases where we would get error like running the test command
```
./versitgw test -a access -s secret -e http://127.0.0.1:7070 full-flow
```

jonaustin09 (approved):

PR #489, opened 2024-04-02
feat: implement read only option for posix/scoutfs

### LINKED ISSUE #484: gateway read only mode
It might be nice to have a read only option for posix/scoutfs backends such that no operations are permitted that would modify the filesystem. This would make it a bit safe to host read only data sets over s3 using the gateway.

benmcclelland:
Fixes #484. When read only option selected, backend will return AccessDenied for any operations that would update data or metadata in the backend. This include creating and deleting objects and buckets as well as updating object metadata.

This will still allow admin to change bucket owner since read only is mainly intended for s3 api only.


benmcclelland:
After more consideration, the default read only should be handled in the frontend code so that it can apply to any backend. If a backend needs specific read only handling in the future, then we can add a special option for that when it happens.

PR #516, opened 2024-04-12
WORM protection with S3 object locks

### LINKED ISSUE #490: add support for object lock functionality
Object Lock is a way to prevent removal or modification of objects before a certain time stamp. See details:
https://aws.amazon.com/s3/features/object-lock/
https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock.html

**Additional context**
We probably will want this enforced at the frontend api layer, so might need a way to store and retrieve retention times in the backend.

Maybe this should be an option to enable? or would we want this on all the time? Configuration and behavior should match AWS experience as closely as possible since common backup utilities expect this behavior from s3 already.

jonaustin09:
Closes #490. Added S3 object lock functionality without object versioning support. 
Added the following s3 actions support in posix:
1. PutObjectLockConfiguration 
2. GetObjectLockConfiguration
3. PutObjectRetention
4. GetObjectRetention
5. PutObjectLegalHold
6. GetObjectLegalHold

Added groups of integration tests for each action and another group testing the whole WORM protection flow.


backend/posix/posix.go
```diff
@@ -2011,6 +2014,263 @@ func (p *Posix) DeleteBucketPolicy(ctx context.Context, bucket string) error {
...
+
+		config.DefaultRetention = retentation
+		now := time.Now()
+		config.CreatedAt = &now
+	}
+
+	configParsed, err := json.Marshal(config)
+	if err != nil {
+		return fmt.Errorf("parse object lock config: %w", err)
+	}
```
benmcclelland: would this logic be better up in the frontend so that we dont need to copy this to all backends? it seems like we could just have
```go
func (p *Posix) PutObjectLockConfiguration(_ context.Context, input []bytes) error
```
for a more opaque storage of this. The backend can be responsible for setting/getting the attributes, but wouldnt the frontend be the best place for the rule parsing?

backend/backend.go
```diff
@@ -81,6 +82,14 @@ type Backend interface {
 	PutObjectTagging(_ context.Context, bucket, object string, tags map[string]string) error
 	DeleteObjectTagging(_ context.Context, bucket, object string) error
 
+	// object lock operations
+	PutObjectLockConfiguration(context.Context, *s3.PutObjectLockConfigurationInput) error
+	GetObjectLockConfiguration(_ context.Context, bucket string) ([]byte, error)
+	PutObjectRetention(context.Context, *s3.PutObjectRetentionInput) error
+	GetObjectRetention(_ context.Context, bucket, object, versionId string) (*types.ObjectLockRetention, error)
+	PutObjectLegalHold(context.Context, *s3.PutObjectLegalHoldInput) error
+	GetObjectLegalHold(_ context.Context, bucket, object, versionId string) (*types.ObjectLockLegalHold, error)
```
benmcclelland: would this be better?
```go
 	PutObjectLockConfiguration(context.Context, []byte) error
 	GetObjectLockConfiguration(_ context.Context, bucket string) ([]byte, error)
 	PutObjectRetention(context.Context, []byte) error
 	GetObjectRetention(_ context.Context, bucket, object, versionId string) ([]byte, error)
 	PutObjectLegalHold(context.Context, []byte) error
 	GetObjectLegalHold(_ context.Context, bucket, object, versionId string) ([]byte, error)
 ```

benmcclelland (approved):

benmcclelland (approved):

PR #663, opened 2024-07-05
Test cmdline acl correction

lrm25:


tests/util_policy.sh
```diff
@@ -47,3 +47,12 @@ setup_policy_with_single_statement() {
...
 }
+
+#get_principal() {
+#  if [[ $DIRECT == "true" ]]; then
+#  fi
+#}
+
+#compare_principal() {
+
+#}
```
benmcclelland: are these some leftovers that need to be cleaned up?

benmcclelland (approved):

PR #673, opened 2024-07-13
feat: add option to allow symlinked directories as buckets

### LINKED ISSUE #644: Posix backend: follow symlinks when listing buckets
I have a unique situation where all of the buckets I'd like to serve via the posix backend are symlinks to various locations on my linux machine.

Consider the following directory structure:
```
var/
└── s3-buckets/
    ├── bucket-1/  -->  /mnt/foo/bucket-1
    ├── bucket-2/  -->  /mnt/bar/bucket-2
    └── bucket-3/  -->  /mnt/baz/bucket-3


mnt/
├── foo/
│   └── bucket-1/
├── bar/
│   └── bucket-2/
└── baz/
    └── bucket-3/
```

Basically I'm trying to flatten out the things in `/mnt`. If I start versitygw with the parameters `./versitygw posix /var/s3-buckets` nothing is returned when listing buckets. I believe that is because of the following:

https://github.com/versity/versitygw/blob/63c9e7503973ad2f49590cfc3303d894ff62ec2f/backend/posix/posix.go#L131-L135

After changing the above to the following it seems to return buckets and allow me to upload. (Far from an exhaustive test)

```go
	for _, entry := range entries {
		fi, err := os.Stat(entry.Name())
		if err != nil {
			// skip entries returning errors
			continue
		}

		if !fi.IsDir() {
			// buckets must be a directory
			continue
		}
```

This works because `os.Stat()` will follow an items symlink unlike what comes back on a `DirEntry` item.

If there's another solution to how I can flatten out the things in `/mnt` without modifying the source structure I'm all ears. Interested to get y'alls thoughts. Thanks.

benmcclelland:
This adds the ability to treat symlinks to directories at the top level gateway directory as buckets the same as normal directories.

This could be a potential security issue allowing traversal into other filesystems within the system, so is defaulted to off. This can be enabled when specifically needed for both posix and scoutfs backend systems.

Fixes #644

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -61,6 +61,10 @@ type Posix struct {
 	// used to determine if chowning is needed
 	euid int
 	egid int
+
+	// bucketlinks is a flag to enable symlinks to directories at the top
+	// level gateway directory to be treated as buckets the same as directories
+	bucketlinks bool
 }
 
 var _ backend.Backend = &Posix{}
@@ -87,8 +91,9 @@ const (
 )
 
 type PosixOpts struct {
-	ChownUID bool
-	ChownGID bool
+	ChownUID    bool
+	ChownGID    bool
+	BucketLinks bool
 }
 
 func New(rootdir string, meta meta.MetadataStorer, opts PosixOpts) (*Posix, error) {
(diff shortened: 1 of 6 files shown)


chaowang-versity (approved):
Should a test be added for following the simlink?

jonaustin09 (approved):

benmcclelland:
> Should a test be added for following the simlink?

yes, trying to decide if this is worth a unit test or if we should just add this in integrations tests

PR #710, opened 2024-08-01
fix: set default content type to binary/octet-stream

### LINKED ISSUE #697: content-type should default to "binary/octet-stream" if not set on upload
**Describe the bug**
Uploading an object without setting the content type explicitly results in head-object returning:
```
    "ContentType": "text/plain; charset=utf-8",
```

When it should return:
```
```
to match AWS behavior

**To Reproduce**
```
# aws s3 mb s3://test
# aws s3 cp ./testfile s3://test/testfile
# aws s3api head-object --bucket test --key testfile
{
    "Restore": "",
    "LastModified": "2024-07-29T14:44:23+00:00",
    "ContentLength": 4862,
    "ETag": "\"2aa6c33b2de8bc9a735111cf3bccf071\"",
    "ContentType": "text/plain; charset=utf-8",
    "Metadata": {},
    "StorageClass": ""
}
```

**Expected behavior**
expected
```
```

**Server Version**
main build
```
Version  : v1.0.4
Build    : 55c7109
BuildTime: 2024-07-30_05:00:18PM
```

**Additional context**
```
# aws --version
aws-cli/2.13.11 Python/3.11.4 Darwin/23.5.0 exe/x86_64 prompt/off
```

benmcclelland:
AWS uses binary/octet-stream for the default content type if the client doesn't specify the content type. Change the default for the gateway to match this behavior.

Fixes #697

### DIFF
--- a/s3api/controllers/base.go
+++ b/s3api/controllers/base.go
@@ -50,7 +50,8 @@ type S3ApiController struct {
 }
 
 const (
-	iso8601Format = "20060102T150405Z"
+	iso8601Format      = "20060102T150405Z"
+	defaultContentType = "binary/octet-stream"
 )
 
 func New(be backend.Backend, iam auth.IAMService, logger s3log.AuditLogger, evs s3event.S3EventSender, mm *metrics.Manager, debug bool, readonly bool) S3ApiController {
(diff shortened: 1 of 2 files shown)


benmcclelland:
> It would be great if we added this case to the integration tests?

ah yes, good call. i'll add on for this

PR #721, opened 2024-08-06
Test cmdline multiple principals

lrm25:


tests/commands/put_bucket_acl.sh
```diff
@@ -14,6 +16,40 @@ put_bucket_acl_s3api() {
   return 0
 }
 
+reset_bucket_acl() {
+  if [[ $# -ne 1 ]]; then
+    log 2 "'reset_bucket_acl' requires bucket name"
+    return 1
+  fi
+  acl_file="acl_file"
+  if ! create_test_files "$acl_file"; then
```
anodelman: I'd like to see us move towards using the run/assert workflow provided in bats - which then also removes a bunch of raw calls to exit.

tests/teardown_static.sh
anodelman, line 1: Wondering why this isn't in a "teardown()" in a bats test file - or is this called after all testing?

tests/env.sh
```diff
@@ -135,6 +135,23 @@ check_versity_vars() {
...
+  fi
+  if [[ -z "$USERNAME_TWO" ]]; then
+    log 2 "missing USERNAME_ONE parameter"
+    return 1
+  fi
+  if [[ -z "$PASSWORD_TWO" ]]; then
+    log 2 "missing PASSWORD_ONE parameter"
+    return 1
+  fi
+  export USERNAME_ONE PASSWORD_ONE USERNAME_TWO PASSWORD_TWO
```
anodelman: Since these already exist (as you are checking if they are set) I don't think that you need to re-export.

anodelman (changes requested):
Some code style comments added.

tests/teardown_static.sh
lrm25: That's a utility script that I added for when using a static bucket configuration (i.e. not removing the buckets between tests).  Right now, it's called manually by me to add and remove the buckets before running things in this config.

tests/commands/put_bucket_acl.sh
lrm25: Added some assert_failure calls to start in the latest test

tests/commands/put_bucket_acl.sh
```diff
@@ -14,6 +16,38 @@ put_bucket_acl_s3api() {
   return 0
 }
 
+reset_bucket_acl() {
+  if [[ $# -ne 1 ]]; then
```
anodelman: Calls like this can all be formed as asserts.

tests/env.sh
```diff
@@ -135,6 +120,22 @@ check_versity_vars() {
 }
 
 check_user_vars() {
+  if [[ -z "$USERNAME_ONE" ]]; then
```
anodelman: Again, form as an assert.

anodelman (changes requested):
I'd like to see the 'if SOMETHING return 1" formed as asserts.

tests/test_aws.sh
```diff
@@ -729,14 +690,19 @@ EOF
...
-  setup_policy_with_single_statement "$test_file_folder/$policy_file" "dummy" "$effect" "$principal" "$action" "$resource" || fail "failed to set up policy"
-  create_multipart_upload "$BUCKET_ONE_NAME" "$test_file" || fail "error creating multipart upload"
+  get_bucket_policy "s3api" "$BUCKET_ONE_NAME" || fail "error getting bucket policy"
+  log 5 "BUCKET POLICY: $bucket_policy"
+  get_bucket_acl "s3api" "$BUCKET_ONE_NAME" || fail "error getting bucket ACL"
+  log 5 "ACL: $acl"
+  setup_policy_with_single_statement "$test_file_folder/$policy_file" "dummy" "$effect" "$principal" "$action" "$resource"
+  assert_success "failed to set up policy"
+  create_multipart_upload "$BUCKET_ONE_NAME" "$test_file"
+  assert_success "failed to create multipart upload"
```
anodelman, line 703: For this assert to fire you need to wrap the 'create_multipart_upload' in a run - otherwise it will use the last run call's output/status and check that.

anodelman (approved):
Just one last nit.  Thanks for all your work on this!

benmcclelland (approved):

PR #737, opened 2024-08-14
Test cmdline policy get tagging

lrm25:


tests/commands/delete_bucket.sh
```diff
@@ -1,12 +1,12 @@
...
 
-# delete an AWS bucket
 # param:  bucket name
-# return 0 for success, 1 for failure
+# fail if params are bad, or bucket exists and user is unable to delete bucket
 delete_bucket() {
+  log 6 "delete_bucket"
   record_command "delete-bucket" "client:$1"
   if [ $# -ne 2 ]; then
-    log 2 "delete bucket missing command type, bucket name"
```
anodelman, line 9: This could be an assert.

tests/commands/delete_bucket.sh
```diff
@@ -32,10 +31,9 @@ delete_bucket() {
   if [ $exit_code -ne 0 ]; then
```
anodelman, line 31: More assert possibilities here.

tests/commands/get_bucket_tagging.sh
```diff
@@ -29,4 +25,28 @@ get_bucket_tagging() {
     return 1
   fi
   export tags
-}
\ No newline at end of file
+}
+
+get_bucket_tagging_with_user() {
+  log 6 "get_bucket_tagging_with_user"
+  if [ $# -ne 3 ]; then
```
anodelman, line 32: assert?

tests/commands/get_object_lock_configuration.sh
```diff
@@ -6,11 +6,13 @@ get_object_lock_configuration() {
     log 2 "'get object lock configuration' command missing bucket name"
     return 1
   fi
-  lock_config=$(aws --no-verify-ssl s3api get-object-lock-configuration --bucket "$1") || local get_result=$?
-  if [[ $get_result -ne 0 ]]; then
+  if ! lock_config=$(aws --no-verify-ssl s3api get-object-lock-configuration --bucket "$1" 2>&1); then
     log 2 "error obtaining lock config: $lock_config"
+    get_object_lock_config_err=$lock_config
+    export get_object_lock_config_err
     return 1
```
anodelman, line 13: Still think that we can do asserts instead of return 1.

anodelman (changes requested):
Can you do a pass seeing where  the check-exit-code-then-return-1 with asserts?

tests/commands/get_object_lock_configuration.sh
```diff
@@ -6,11 +6,13 @@ get_object_lock_configuration() {
...
   fi
+  lock_config=$(echo "$lock_config" | grep -v "InsecureRequestWarning")
   export lock_config
```
anodelman: I don't believe that these export calls are necessary - variable creation is global by default and should be available after the function is called.

anodelman (changes requested):
Just some question about the need for the export calls for variables in functions.

anodelman (approved):
Looks good.

PR #743, opened 2024-08-20
fix: azure backend errors with storage url and container metadata

### LINKED ISSUE #736: [Bug] - "The requested URI does not represent any resource on the server" when using root credentials to PutObject with Azure backend
**Describe the bug**
PutObject request fails with "The requested URI does not represent any resource on the server" even when using root access.

**To Reproduce**
1. Start gateway with azure backend:
 `docker run -p 7070:7070 versity/versitygw --access s3a --secret s3s azure -a <az-account-name> -k <az-account-key>`
3. Create blob container on Azure Blob Storage manually
4. Export root credentials as env:
```
export AWS_ACCESS_KEY_ID=s3a
export AWS_SECRET_ACCESS_KEY=s3s
```
5. Verify gateway is successfully connecting to storage:
`aws --endpoint-url http://127.0.0.1:7070 s3 ls s3://`. 
Container(s) are listed successfully
7. Try to upload object: 
`aws --endpoint-url http://127.0.0.1:7070 s3 cp localfile s3://<container-from-output-above>/test.txt`
8. Request fails with 400 with following response body - `<Error><Code>InvalidUri</Code><Message>400 The requested URI does not represent any resource on the server.</Message><Resource></Resource><RequestId></RequestId><HostId></HostId></Error>`

**Expected behavior**
Object to be successfully uploaded

**Server Version**
```
Version  : v1.0.5
Build    : adbc8140ed85e6b5b240600b49fcd60fe2282307
BuildTime: 2024-08-06T17:53:24Z
```

**Additional context**
<!-- Describe s3 client and version if applicable.

benmcclelland:
The container URL was adding an extra / when the service URL ended with /. This results in an invalid container URI error from Azure.

This also fixes the container metadata keys to not include "-" that the Azure service disallows. The metadata values are base64 encoded to prevent any special chacter handling on Azure service side.

The content type is set for uploads consistent with S3 expectations.

Fixes #736

### DIFF
--- a/backend/azure/azure.go
+++ b/backend/azure/azure.go
@@ -30,6 +30,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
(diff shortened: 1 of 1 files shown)


benmcclelland:
There are still several test failures dealing with multipart uploads since the azure blob service doesnt have a corresponding create multipart upload. For azure blob, it just starts uploading parts without initiating the multipart upload. We usually set some metadata on create multipart upload, so we need to figure out how to translate this to azure.

benmcclelland:
Its also interesting that some of the failures fixed with this PR could only be seen when interacting with the real azure service. So azurite doesnt appear to be a perfect proxy for testing azure.

jonaustin09 (approved):

PR #757, opened 2024-08-24
fix: put file corruption with chunked transfer

### LINKED ISSUE #753: [Bug] - Objects put with mc are corrupt
**Describe the bug**

Objects put with `mc` are stored corrupt. This is the case for small and large objects.

No apparent issue with GetObject. The download matches what lands in storage in size and content.

On put (non-multipart), the etag is correct. The reflected size by list-objects is incorrect.

I haven't analyzed the multipart case, but it appears the corruption is per-part. What is happening for single puts is occurring for each part. Resulting etag is incorrect in addition to the reflected size being incorrect.

Here is an example for a small file. The file named and containing "hello" is get and put and hexdumped for comparison.  Captures will be attached.
```
$ mc cp --debug hello vers/test1/hello1 >capture.upload 2>&1
$ mc cp --debug vers/test1/hello1 hello1 >capture.dl 2>&1
$ hexdump hello
0000000 6568 6c6c 0a6f
0000006
$ hexdump hello1
0000000 6568 6c6c 0a6f 0000 0000 0000 0000 0000
0000010 0000 0000 0000 0000 0000 0000 0000 0000
*
00000b0 0000
00000b2

$ mc ls --json vers/test1/hello1
{
 "status": "success",
 "type": "file",
 "lastModified": "2024-08-22T15:19:13.724-06:00",
 "size": 178,
 "key": "hello1",
 "etag": "b1946ac92492d2347c6235b4d2611184",
 "url": "http://localhost:7070/test1/hello1",
 "versionOrdinal": 1
}
```
[capture.upload.txt](https://github.com/user-attachments/files/16719878/capture.upload.txt)
[capture.dl.txt](https://github.com/user-attachments/files/16719877/capture.dl.txt)

From the captured upload debug:
```
mc: <DEBUG> PUT /test1/hello1 HTTP/1.1
Host: localhost:7070
User-Agent: MinIO (linux; amd64) minio-go/v7.0.73 mc/RELEASE.2024-07-26T13-08-44Z
Content-Length: 178
Accept-Encoding: zstd,gzip
Authorization: AWS4-HMAC-SHA256 Credential=user/20240822/us-east-1/s3/aws4_request,SignedHeaders=host;x-amz-content-sha256;x-amz-date;x-amz-decoded-content-length,Signature=**REDACTED**
Content-Type: application/octet-stream
X-Amz-Content-Sha256: STREAMING-AWS4-HMAC-SHA256-PAYLOAD
X-Amz-Date: 20240822T211913Z
X-Amz-Decoded-Content-Length: 6

mc: <DEBUG> HTTP/1.1 200 OK
Connection: close
Content-Length: 0
Date: Thu, 22 Aug 2024 21:19:13 GMT
Etag: b1946ac92492d2347c6235b4d2611184
Server: VERSITYGW

mc: <DEBUG> Response Time:  755.591µs
```

The stored file ends up being 178 bytes. I point out that this is the value from `Content-Length`. The actual size of the object is 6 bytes, value from `X-Amz-Decoded-Content-Length`.

**To Reproduce**

```
# mc cp testfile <versity>/<bucket>/testfile
```
Or:
```
# mc cp --disable-multipart testfile <versity>/<bucket>/testfile
```

The original and downloaded file will differ.

Client version is:
```
MinIO (linux; amd64) minio-go/v7.0.73 mc/RELEASE.2024-07-26T13-08-44Z
```

**Expected behavior**

The object on get should be the correct size and contain the correct data.

**Server Version**

Posix backend with ext4 filesystem.

```
# versitygw --version && uname -a
Version  : 1.0.5
Build    : adbc8140ed85e6b5b240600b49fcd60fe2282307
BuildTime: 2024-08-06T17:50:34Z
Linux 9ac166f12a05 4.18.0-305.10.2.x6.3.000.52.x86_64 #1 SMP Fri Jun 17 08:34:03 MDT 2022 x86_64 x86_64 x86_64 GNU/Linux
```

**Additional context**

Testing was performed on versitygw running in a rootful podman container.
This report is assuming that `mc` is well behaved. Positive evidence is that the single object put etag matches and content up to the padding is correct.

benmcclelland:
When on linux with O_TMPFILE support, we issue and fallocate for the expected object size ax an optimization for the underlying filesystem to allocate the full file all ate once. With the chunked transfer encoding, the final object size is recoded in the X-Amz-Decoded-Content-Length instead of the standard ContentLength which includes the chunk encoding in the payload.

We were incorrectly using the content length to fallocate the file which would cause the filesystem to pad out any unwritten length to this size with 0s.

The fix here is to make sure we pass the X-Amz-Decoded-Content-Length as the object size to the backend for all PUTs.

Fixes #753

### DIFF
--- a/s3api/controllers/base.go
+++ b/s3api/controllers/base.go
@@ -1544,11 +1544,19 @@ func (c S3ApiController) PutActions(ctx *fiber.Ctx) error {
 	granWrite := ctx.Get("X-Amz-Grant-Write")
 	grantWriteACP := ctx.Get("X-Amz-Grant-Write-Acp")
 
-	// Other headers
+	// Content Length
 	contentLengthStr := ctx.Get("Content-Length")
 	if contentLengthStr == "" {
 		contentLengthStr = "0"
 	}
+	// Use decoded content length if available because the
+	// middleware will decode the chunked transfer encoding
+	decodedLength := ctx.Get("X-Amz-Decoded-Content-Length")
+	if decodedLength != "" {
+		contentLengthStr = decodedLength
+	}
+
+	// Other headers
 	bucketOwner := ctx.Get("X-Amz-Expected-Bucket-Owner")
 	storageClass := ctx.Get("X-Amz-Storage-Class")
 


hnicke:
I as well had observed trailing null bytes on uploaded files.
kopia failed to when reading the initialized repository, because when if fetched the json blobs from the repo, the json parser failed (trailing null bytes are, according to spec, not valid).
Can confirm this patch fixes it.

Issue #773, opened 2024-08-29
[Bug] - CopyObject fails with error "the specified bucket does not exist" if versitygw is configured with an IAM backend


cvubrugier:
**Describe the bug**

With WinSCP and Cyberduck, renaming or duplicating an object fails with error "the specified bucket does not exist" if versitygw is configured with an IAM backend.

**To Reproduce**

1. Start versitygw with an IAM backend (for instance IAM Internal).
1. Create a user.
1. As admin, create a bucket.
1. Change ownership of the bucket to the new user.
1. Browse the bucket with WinSCP or Cyberduck as a user.
1. Upload a file.
1. Try to rename or duplicate the file. The operation fails with error "the specified bucket does not exist".

**Expected behavior**

The user should be able to rename or duplicate a file with WinSCP or Cyberduck.

**Server and client versions**

* versitygw v1.0.5 or latest revision of the "main" branch (commit 923ee5f)
* WinSCP 6.3.4
* Cyberduck 8.7.3

**Bug analysis**

To rename or duplicate an object, WinSCP and Cyberduck send a [CopyObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html) S3 command that contains a `x-amz-copy-source` header that starts with '/'.

The function [VerifyObjectCopyAccess](https://github.com/versity/versitygw/blob/main/auth/acl.go#L407) contains the following code:

    srcBucket, srcObject, found := strings.Cut(copySource, "/")

If `copySource` starts with '/', then `srcBucket` is set to an empty string. Later, an error is returned because bucket "" does not exist.

The same issue was fixed in the Posix and Azure backends by the following commit:

 * 5e484f2 fix: Fixed CopySource parsing to handle the values starting with '/' in CopyObject action in posix and azure backends.

However, the issue still exist in `VerifyObjectCopyAccess`.

I will provide a fix for this issue.

cvubrugier:
<img width="406" alt="cyberduck-copy-object-error" src="https://github.com/user-attachments/assets/d69a5114-692f-4e70-aeb4-f0a872992aeb">
<img width="386" alt="winscp-copy-object-error" src="https://github.com/user-attachments/assets/fe6ae4c7-31a5-401a-b95e-9e84e26feec3">

benmcclelland:
fix PR merged

PR #774, opened 2024-08-29
fix: handle "x-amz-copy-source" header starting with '/' in s3api

cvubrugier:
The "x-amz-copy-source" header may start with '/' as observed with WinSCP. However, '/' is also the separator between the bucket and the object path in "x-amz-copy-source".

Consider the following code in VerifyObjectCopyAccess():

    srcBucket, srcObject, found := strings.Cut(copySource, "/")

If `copySource` starts with '/', then `srcBucket` is set to an empty string. Later, an error is returned because bucket "" does not exist.

This issue was fixed in the Posix and Azure backends by the following commit:

 * 5e484f2 fix: Fixed CopySource parsing to handle the values starting with '/' in CopyObject action in posix and azure backends.

But the issue was not fixed in `VerifyObjectCopyAccess`.

This commit sanitizes "x-amz-copy-source" right after the header is extracted in `s3api/controllers/base.go`. This ensures that the `CopySource` argument passed to the backend functions UploadPartCopy() and CopyObject() does not start with '/'. Since the backends no longer need to strip away any leading '/' in `CopySource`, the parts of commit 5e484f2 modifying the Posix and Azure backends are reverted.

Fixes issue #773.

### DIFF
--- a/backend/azure/azure.go
+++ b/backend/azure/azure.go
@@ -692,12 +692,7 @@ func (az *Azure) CopyObject(ctx context.Context, input *s3.CopyObjectInput) (*s3
 		return nil, err
 	}
 
-	cpSrc := *input.CopySource
-	if cpSrc[0] == '/' {
-		cpSrc = cpSrc[1:]
-	}
-
-	if strings.Join([]string{*input.Bucket, *input.Key}, "/") == cpSrc && isMetaSame(mdmap, input.Metadata) {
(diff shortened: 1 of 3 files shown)


versity-github:

jonaustin09 (approved):

benmcclelland (approved):

PR #777, opened 2024-08-29
test: add better file comparisons

lrm25:


anodelman (changes requested):
I see a lot of use of create_test_file - with uses an assert.  In some places it is wrapped in a run call (run create_test_file) which would result in the assert being eaten.  In my own code, if I have a multiline function then I ensure that it only uses exits/returns instead of asserts - then I use run/assert at the top level.  Let's go for that pattern here.  Same for copy_file_locally & move_file_locally.  

In this way, run/assert should only show up in bats test files.  Utilities and other code should use return/exit.

I know that we've had some confusion around this and I've had to do a review on my own bats testing code.  Thanks for rolling with the changes as we both get to an understanding as to how run/assert/exit all interact.

lrm25:
> I see a lot of use of create_test_file - which uses an assert. In some places it is wrapped in a run call (run create_test_file) which would result in the assert being eaten. In my own code, if I have a multiline function then I ensure that it only uses exits/returns instead of asserts - then I use run/assert at the top level. Let's go for that pattern here. Same for copy_file_locally & move_file_locally.
> 
> In this way, run/assert should only show up in bats test files. Utilities and other code should use return/exit.
> 
> I know that we've had some confusion around this and I've had to do a review on my own bats testing code. Thanks for rolling with the changes as we both get to an understanding as to how run/assert/exit all interact.

So, just to clarify, use run/assert at the top level in all of the tests?  That's the pattern we want to go with?

anodelman:
Right - everything that is a multiline function uses return/exit and then have the runs at the top level.  That also makes a division between test code (that uses bats-ism) and utility code (that uses bash-isms).

anodelman (changes requested):
There's a lot of ' || fail "message" ' in the test files - I think that those should be run/asserts since they are at the top level.

lrm25:
> There's a lot of ' || fail "message" ' in the test files - I think that those should be run/asserts since they are at the top level.

Should those all be changed now, or gradually over the following commits?

anodelman (approved):
Looks good

PR #785, opened 2024-09-05
fix: DeleteObject not working as expected when using `--bucketlinks` (#784)

### LINKED ISSUE #784: [Bug] - DeleteObject not working as expected when using `--bucketlinks`
**Describe the bug**
When using versitygw with the posix backend and `--bucketlinks`, the bucket's symlink gets destroyed when deleting an object at the root of that bucket.

To add some more context, I am using symlinks with versitygw for reasons I outlined in this [previous issue](https://github.com/versity/versitygw/issues/644).

**To Reproduce**

```bash
# Create test bucket
mkdir bucket-1

# Add random file
touch bucket-1/test.txt

# Create directory to house buckets
mkdir buckets

# Symlink bucket-1 to buckets/bucket-1
ln -s ${PWD}/bucket-1 ${PWD}/buckets/bucket-1

# Start versitygw
./versitygw --access myaccess --secret mysecret posix --bucketlinks ${PWD}/buckets

# Sanity check
# Ensure versity started and we can see our bucket + object
aws --endpoint-url http://127.0.0.1:7070 s3 ls s3://bucket-1

# Delete bucket-1/test.txt
# This invokes the bug
aws --endpoint-url http://127.0.0.1:7070 s3 rm s3://bucket-1/test.txt

# Another ls
aws --endpoint-url http://127.0.0.1:7070 s3 ls s3://bucket-1

# No symlinks in buckets anymore
ls -l buckets
```

**Expected behavior**

The symlink `buckets/bucket-1` should not be removed

**Server Version**

```
Version  : 1.0.6
Build    : 9cfc2c7b0836fc786d302ee94e3d482d1f9f64fc
BuildTime: 2024-08-29T20:00:30Z
Linux <hostname> 6.2.0-31-generic #31~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Aug 16 13:45:26 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

**Additional context**

Clients used to test:
aws-cli 2.17.44
@aws-sdk/client-s3 3.614.0 (javascript SDK)

kyleaupton:
This fixes #784.

`removeParents()` gets called when `posix.DeleteObject()` is called. It will iterate through each parent directory of an object path that is being deleted, and it'll also delete that parent unless:
1. The parent is the bucket
2. The parent has a valid etag
3. The parent is not empty (or if `os.Remove()` fails for another reason)

I believe the logic used to check if the parent is the bucket is off.

Currently this is the condition:

```go
parent := filepath.Dir(objPath)

if parent == string(filepath.Separator) {
	// stop removing parents if we hit the bucket directory.
	break
}
```

Because `objPath` doesn't start with a `/` (relative style), it is impossible for `parent` to ever equal `/`. Instead I think we should test for `.`

Referenced: https://pkg.go.dev/path/filepath#example-Dir

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -1571,7 +1571,7 @@ func (p *Posix) removeParents(bucket, object string) error {
 	for {
 		parent := filepath.Dir(objPath)
 
-		if parent == string(filepath.Separator) {
+		if parent == "." {
 			// stop removing parents if we hit the bucket directory.
 			break
 		}


versity-github:

benmcclelland (approved):

PR #786, opened 2024-09-06
fix: set content type on put object

### LINKED ISSUE #783: versity does not store object content type
**Describe the bug**
Using the cli s3api functionality, if we use 'put-object' to push an object into versity with content-type text/stream and then inspect the object with 'head-object', the content type is reported as binary/octet-stream. AWS s3 reports the object with the content-type that we specified. 

**To Reproduce**
DEV03 [/mmfs1/data/TISRC] :>vgaws s3api put-object --key test --body file100.txt --bucket tonyitest --content-type text/plain
{
    "ETag": "86c27a184101d828becca6ec4b65f3b8"
}
DEV03 [/mmfs1/data/TISRC] :>vgaws s3api head-object --key test --buck tonyitest
{
    "Restore": "",
    "LastModified": "Thu, 05 Sep 2024 15:20:19 GMT",
    "ContentLength": 10,
    "ETag": "86c27a184101d828becca6ec4b65f3b8",
    "ContentType": "binary/octet-stream",   ##<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
    "Metadata": {},
    "StorageClass": "STANDARD"
}

**Expected behavior**
Object should be reported with correct content-type, as aws s3 does.

**Server Version**
Cloned and built on 4th september 2024

**Additional context**

benmcclelland:
This fixes put object with setting a content type. If no content type is set, then we will return a default content type for following requests. Mutli-part upload appears to be ok.

Fixes #783

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -86,6 +86,9 @@ const (
 	objectRetentionKey  = "object-retention"
 	objectLegalHoldKey  = "object-legal-hold"
 
+	// this is the media type for directories in AWS and Nextcloud
+	dirContentType = "application/x-directory"
+
 	doFalloc   = true
 	skipFalloc = false
 )
@@ -474,7 +477,8 @@ func (p *Posix) CreateMultipartUpload(ctx context.Context, mpu *s3.CreateMultipa
 	}
 
 	// set content-type
-	if *mpu.ContentType != "" {
+	ctype := getString(mpu.ContentType)
+	if ctype != "" {
 		err := p.meta.StoreAttribute(bucket, filepath.Join(objdir, uploadID),
 			contentTypeHdr, []byte(*mpu.ContentType))
 		if err != nil {
@@ -482,6 +486,19 @@ func (p *Posix) CreateMultipartUpload(ctx context.Context, mpu *s3.CreateMultipa
 			os.RemoveAll(filepath.Join(tmppath, uploadID))
 			os.Remove(tmppath)
 			return s3response.InitiateMultipartUploadResult{}, fmt.Errorf("set content-type: %w", err)
+		}
+	}
+
+	// set content-encoding
+	cenc := getString(mpu.ContentEncoding)
+	if cenc != "" {
+		err := p.meta.StoreAttribute(bucket, filepath.Join(objdir, uploadID), contentEncHdr,
+			[]byte(*mpu.ContentEncoding))
+		if err != nil {
+			// cleanup object if returning error
+			os.RemoveAll(filepath.Join(tmppath, uploadID))
+			os.Remove(tmppath)
+			return s3response.InitiateMultipartUploadResult{}, fmt.Errorf("set content-encoding: %w", err)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


backend/posix/posix.go
```diff
@@ -485,6 +486,19 @@ func (p *Posix) CreateMultipartUpload(ctx context.Context, mpu *s3.CreateMultipa
...
+		err := p.meta.StoreAttribute(bucket, filepath.Join(objdir, uploadID), contentEncHdr,
+			[]byte(*mpu.ContentEncoding))
+		if err != nil {
+			// cleanup object if returning error
+			os.RemoveAll(filepath.Join(tmppath, uploadID))
+			os.Remove(tmppath)
+			return s3response.InitiateMultipartUploadResult{}, fmt.Errorf("set content-encoding: %w", err)
+		}
+	}
+
```
mlt180, line 504: I think it would be better to have an integration test case, that covers multipart uploads with Content-Encoding?

backend/posix/posix.go
```diff
@@ -1408,11 +1416,21 @@ func (p *Posix) PutObject(ctx context.Context, po *s3.PutObjectInput) (string, e
...
 
+		ctype := getString(po.ContentType)
+		if ctype != "" {
+			err := p.meta.StoreAttribute(*po.Bucket, *po.Key, contentTypeHdr,
+				[]byte(*po.ContentType))
+			if err != nil {
+				return "", fmt.Errorf("set content-type attr: %w", err)
+			}
+		}
+
```
mlt180: Why would we need to add Content-Type for directory objects?
If it's for better compatibility with S3, then we would probably need to add Content-Encoding either?

tests/integration/tests.go
```diff
@@ -3028,6 +3028,49 @@ func HeadObject_non_existing_dir_object(s *S3Conf) error {
...
+	testName := "HeadObject_with_contenttype"
+	return actionHandler(s, testName, func(s3client *s3.Client, bucket string) error {
+		obj, dataLen := "my-obj", int64(1234567)
+		ctype := "text/plain"
+
+		_, _, err := putObjectWithData(dataLen, &s3.PutObjectInput{
+			Bucket:      &bucket,
+			Key:         &obj,
+			ContentType: &ctype,
+		}, s3client)
```
mlt180: Maybe you can include Content-Encoding check here?

backend/posix/posix.go
```diff
@@ -1505,6 +1525,24 @@ func (p *Posix) PutObject(ctx context.Context, po *s3.PutObjectInput) (string, e
...
+	cenc := getString(po.ContentEncoding)
+	if cenc != "" {
+		err := p.meta.StoreAttribute(*po.Bucket, *po.Key, contentEncHdr,
+			[]byte(*po.ContentEncoding))
+		if err != nil {
+			return "", fmt.Errorf("set content-encoding attr: %w", err)
+		}
+	}
+
```
mlt180, line 1548: An integration test case to check Content-Encoding for PutObject?

backend/posix/posix.go
mlt180: An integration test to check the Content-Type for directory objects?

benmcclelland: actually thats a good point, I dont think we need to do this for directory objects

PR #806, opened 2024-09-13
fix: list objects trim common prefixes that match marker prefix

### LINKED ISSUE #778: s3 ls --page-size can return duplicate output
**Describe the bug**
See repro below - adding a combination of directory and file and then listing with a small page size results in duplicate lines in the output

**To Reproduce**
printf "File 100.\n" >file100.txt
d1aws s3api put-object --bucket ngenea-test --key dir14/subdir/file.txt --body file100.txt
d1aws s3api put-object --bucket ngenea-test --key dir14/subdir.ext --body file100.txt
d1aws s3 ls  --page-size=2 s3://ngenea-test/dir14/

OUTPUT:
{
    "ETag": "86c27a184101d828becca6ec4b65f3b8"
}
{
    "ETag": "86c27a184101d828becca6ec4b65f3b8"
}
                           PRE subdir/
2024-08-30 17:30:31         10 subdir.ext
                           PRE subdir/                       #      <<<<<<<<<<<<<<<<< ???

**Expected behavior**
No duplicate output

**Server Version**
built from clone on 30 aug 2024

**Additional context**

benmcclelland:
This checks to see if the common prefix is before the marker and thus would have been returned in earlier list objects request.

The error case was aws cli listing multiple entries for the same common prefix when the listing required multiple pagination requests.

Fixes #778

### DIFF
--- a/tests/integration/tests.go
+++ b/tests/integration/tests.go
@@ -3590,6 +3590,33 @@ func ListObjects_with_prefix(s *S3Conf) error {
 	})
 }
 
+func ListObjects_paginated(s *S3Conf) error {
+	testName := "ListObjects_paginated"
+	return actionHandler(s, testName, func(s3client *s3.Client, bucket string) error {
+		_, err := putObjects(s3client, []string{"dir1/subdir/file.txt", "dir1/subdir.ext", "dir1/subdir1.ext", "dir1/subdir2.ext"}, bucket)
+		if err != nil {
+			return err
+		}
+
+		objs, prefixes, err := listObjects(s3client, bucket, "dir1/", "/", 2)
+		if err != nil {
+			return err
+		}
+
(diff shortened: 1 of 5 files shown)


jonaustin09 (commented):
Could you add an integration test case for paginated common prefixes ?

Could you make sure azure also handles this case correctly and passes the tests?

tests/integration/tests.go
```diff
@@ -3601,6 +3601,49 @@ func ListObjects_with_prefix(s *S3Conf) error {
...
+}
+
+func pfxStrings(pfxs []types.CommonPrefix) []string {
+	pfxStrs := make([]string, len(pfxs))
+	for i, pfx := range pfxs {
+		pfxStrs[i] = *pfx.Prefix
+	}
+	return pfxStrs
+}
+
```
mlt180: Maybe we should move this to utils.go?

jonaustin09 (approved):

PR #849, opened 2024-09-27
fix: Changed the GetBucketVersioning action return type, to return em…

### LINKED ISSUE #840: [Bug] - GetBucketVersioning should return empty response for unset versioning config
**Describe the bug**
GetBucketVersioning should returns non-empty response, when attempting to get unset bucket versioning configuration
Sample response from AWS CLI:
```
{
    "Status": ""
}
```

**To Reproduce**
```
// create a bucket
// get bucket versioning
aws s3api get-bucket-versioning --bucket test 
```

**Expected behavior**
The action should response empty response.

jonaustin09:
Fixes #840

Changed `GetBucketVersioning` response type to return empty response when versioning is not set(Enabled, Suspended).

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -39,7 +39,7 @@ type Backend interface {
 	PutBucketAcl(_ context.Context, bucket string, data []byte) error
 	DeleteBucket(context.Context, *s3.DeleteBucketInput) error
 	PutBucketVersioning(_ context.Context, bucket string, status types.BucketVersioningStatus) error
-	GetBucketVersioning(_ context.Context, bucket string) (*s3.GetBucketVersioningOutput, error)
+	GetBucketVersioning(_ context.Context, bucket string) (s3response.GetBucketVersioningOutput, error)
 	PutBucketPolicy(_ context.Context, bucket string, policy []byte) error
 	GetBucketPolicy(_ context.Context, bucket string) ([]byte, error)
 	DeleteBucketPolicy(_ context.Context, bucket string) error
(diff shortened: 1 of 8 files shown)


backend/backend.go
```diff
@@ -39,7 +39,7 @@ type Backend interface {
 	PutBucketAcl(_ context.Context, bucket string, data []byte) error
 	DeleteBucket(context.Context, *s3.DeleteBucketInput) error
 	PutBucketVersioning(_ context.Context, bucket string, status types.BucketVersioningStatus) error
-	GetBucketVersioning(_ context.Context, bucket string) (*s3.GetBucketVersioningOutput, error)
+	GetBucketVersioning(_ context.Context, bucket string) (*s3response.GetBucketVersioningOutput, error)
```
benmcclelland: make this s3response.GetBucketVersioningOutput instead of a pointer to match other s3response types

s3response/s3response.go
```diff
@@ -384,3 +384,9 @@ type ListVersionsResult struct {
 	VersionIdMarker     *string
 	Versions            []types.ObjectVersion `xml:"Version"`
 }
+
+type GetBucketVersioningOutput struct {
+	XMLName   xml.Name `xml:"http://s3.amazonaws.com/doc/2006-03-01/ VersioningConfiguration" json:"-"`
+	MFADelete *types.MFADeleteStatus
+	Status    *types.BucketVersioningStatus
```
benmcclelland, line 391: I think these need an omitempty xml tag

benmcclelland (changes requested):

s3response/s3response.go
benmcclelland: oh, does nil achieve this already? I didnt realize these are pointers before. that might make it so oomitempty is not needed.

benmcclelland (approved):

PR #853, opened 2024-09-28
field alignment fix

T-M-A:
field alignment fix

### DIFF
--- a/auth/acl.go
+++ b/auth/acl.go
@@ -45,18 +45,18 @@ type GetBucketAclOutput struct {
 
 type PutBucketAclInput struct {
 	Bucket              *string
-	ACL                 types.BucketCannedACL
 	AccessControlPolicy *AccessControlPolicy
 	GrantFullControl    *string
 	GrantRead           *string
 	GrantReadACP        *string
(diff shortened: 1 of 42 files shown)


versity-github:

benmcclelland:
@T-M-A I appreciate the attempt at optimizing the structure fields. I think this is just the `fieldalignment` tool run to update all files? The problem with this tool is the following:

- removes comments in structs when reordering fields
- modifies files that are auto generated and not to be edited (see moq files)
- updates some structs that we would prefer readability over minimal memory use (such as options)
- changes some code styling such as embedded structs first, and mutex directly before members it protects

For a change like this, it is recommended to open a PR for a single directory (or less) at a time. This allows accepting smaller PRs while we address issues in other PRs.

The commit message should explain the purpose of this change and what was changed. For example:

```
chore: run fieldalignment tool to optimize struct memory layout

- Executed Go's fieldalignment tool to detect and optimize struct field ordering
- Improved memory alignment by rearranging fields where necessary
- This change reduces memory padding and may result in performance improvements
```

PR #904, opened 2024-10-21
ListObjects: don't recurse beyond prefix

### LINKED ISSUE #903: [Bug] - os.PathSeparator incorrect for io/fs
**Describe the bug**
[backend.Walk](https://github.com/versity/versitygw/blob/main/backend/walk.go#L88) constructs paths using `os.PathSeparator`

This has two problems:

1. io/fs uses forward slash on all platforms, [including windows](https://pkg.go.dev/io/fs#ValidPath) "Note that paths are slash-separated on all systems, even Windows", so `os.PathSeparator` is incorrect for the io/fs abstraction.
2. The prefix passed in by the caller probably wants to work with either `/` or delimiter, not `os.PathSeparator`

I'm looking at speeding up `backend.Walk`. I'll see about improving this while I'm in there.

lunixbochs:
This fixes the biggest performance issue with ListObjects for me - it would recurse into all subdirectories, even when using a delimiter.

There are still a few issues:
- Large directories take forever to list.
- Parent directories are still walked for no reason - perhaps the posix backend should scope the os.DirFS() to the last directory in the prefix rather than starting at the root, and stitch the paths back together in `p.fileToObj`. I looked at doing that but wasn't sure if I'd hit a `prefix = ""` edge case.

Also fixes #903

### DIFF
--- a/backend/walk.go
+++ b/backend/walk.go
@@ -75,6 +75,9 @@ func Walk(ctx context.Context, fileSystem fs.FS, prefix, delimiter, marker strin
 			return fs.SkipAll
 		}
 
+		// After this point, return skipflag instead of nil
+		// so we can skip a directory without an early return
+		var skipflag error
 		if d.IsDir() {
 			// If prefix is defined and the directory does not match prefix,
 			// do not descend into the directory because nothing will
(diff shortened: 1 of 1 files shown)


versity-github:

backend/walk.go
```diff
@@ -85,8 +84,16 @@ func Walk(ctx context.Context, fileSystem fs.FS, prefix, delimiter, marker strin
 			// building to match. So only skip if path isn't a prefix of prefix
 			// and prefix isn't a prefix of path.
 			if prefix != "" &&
-				!strings.HasPrefix(path+string(os.PathSeparator), prefix) &&
-				!strings.HasPrefix(prefix, path+string(os.PathSeparator)) {
+				!strings.HasPrefix(path+"/", prefix) &&
```
benmcclelland, line 90: set a const in this file to "/" and use that instead of using "/" everywhere

backend/walk.go
```diff
@@ -315,8 +322,16 @@ func WalkVersions(ctx context.Context, fileSystem fs.FS, prefix, delimiter, keyM
...
+				!strings.HasPrefix(prefix, path+"/") {
+				return fs.SkipDir
+			}
+
+			// Don't recurse into subdirectories when listing with delimiter.
+			if delimiter == "/" &&
+				prefix != path+"/" &&
+				strings.HasPrefix(path+"/", prefix) {
+				cpmap[path+"/"] = struct{}{}
```
benmcclelland, line 335: I think this is adding an optimization for the special case of delimiter="/", maybe the comment can be updated to clarify this? Also I think this might be causing the functional test failure:
```
RUN  ListObjects_paginated
16:39:31 | 200 |     583.638µs | 127.0.0.1 | PUT | /test-bucket-97 | - | 
16:39:31 | 200 |      858.73µs | 127.0.0.1 | PUT | /test-bucket-97/dir1/subdir/file.txt | - | x-id=PutObject
16:39:31 | 200 |      692.35µs | 127.0.0.1 | PUT | /test-bucket-97/dir1/subdir.ext | - | x-id=PutObject
16:39:31 | 200 |     759.425µs | 127.0.0.1 | PUT | /test-bucket-97/dir1/subdir1.ext | - | x-id=PutObject
16:39:31 | 200 |     730.151µs | 127.0.0.1 | PUT | /test-bucket-97/dir1/subdir2.ext | - | x-id=PutObject
16:39:31 | 200 |     1.34157ms | 127.0.0.1 | GET | /test-bucket-97 | - | delimiter=%2F&list-type=2&max-keys=2&prefix=dir1%2F
16:39:31 | 200 |     673.755µs | 127.0.0.1 | GET | /test-bucket-97 | - | continuation-token=dir1%2Fsubdir.ext&delimiter=%2F&list-type=2&max-keys=2&prefix=dir1%2F
16:39:31 | 200 |     673.735µs | 127.0.0.1 | GET | /test-bucket-97 | - | continuation-token=dir1%2Fsubdir1.ext&delimiter=%2F&list-type=2&max-keys=2&prefix=dir1%2F
FAIL ListObjects_paginated: expected prefixes [dir1/subdir/], instead got [dir1/subdir/ dir1/subdir/ dir1/subdir/]
```
you can manually run the single test with:
```
./versitygw test -a <accesskey> -s <secretkey> -e http://127.0.0.1:7070 ListObjects_paginated
```

lunixbochs: I think I can technically make this work with other delimiters too, this was just slightly easier to implement because it happens to line up perfectly with directory separation

backend/walk.go
lunixbochs: I have mixed feelings about this, it reduces readability, fs.FS is never going to use something other than a slash, and several other files in this repo also hardcode "/"

lunixbochs:
The new logic should work with any delimiter:
```go
if delimiter != "" &&
    strings.HasPrefix(path+"/", prefix) &&
    strings.Contains(strings.TrimPrefix(path+"/", prefix), delimiter) {
    skipflag = fs.SkipDir
```

If the delimiter is non-empty, and our path is beyond the prefix, and the delimiter is still in the directory path after trimming the prefix from it, we shouldn't recurse.

I fixed the pagination issue by not doing an early return, instead opting to set a `skipflag` variable which is used instead of `return nil` for the rest of the walk closure.

benmcclelland (approved):

PR #912, opened 2024-10-24
test: REST multipart upload listing, completion

lrm25:


tests/util_bucket.sh
```diff
@@ -0,0 +1,304 @@
...
+  fi
+  if [ $exists -eq 0 ]; then
+    return 0
+  fi
+  return 1
+}
+
+# params:  client, bucket name
+# return 0 for success, 1 for error
+delete_bucket_or_contents() {
```
anodelman: I'm a little confused as to this method - why isn't it two methods?  One for deleting only bucket contents and one for deleting the bucket itself?  Considering that it is mainly just a wrapper with an variable check it just seems weird to me.

tests/util_bucket.sh
```diff
@@ -0,0 +1,304 @@
...
+    log 2 "error with recursive bucket delete"
+    return 1
+  fi
+  log 5 "bucket deletion success"
+  return 0
+}
+
+# params: client, bucket name
+# return 0 for success, 1 for error
+delete_bucket_or_contents_if_exists() {
```
anodelman: Same question here?  Seems like the 'or' clause belongs a level higher in the code and not in a method name.

tests/util_bucket.sh
lrm25: It was just to combine operations for static and non-static bucket cleanup.  For both, occasionally you need to delete individual objects due to WORM protection.  And just to avoid putting the static check in each test.  If you think there's a better way to do it though, let me know.

anodelman: Could maybe just use different naming to be clearer what is going on - bucket_cleanup?  bucket_cleanup_worm_compliant?  And then add a comment to each section explaining why there's different operations for static/non-static.

anodelman (approved):
Easier for me to read, thanks.

Issue #919, opened 2024-10-26
[Feature] - Improve GetObject allocation overhead


lunixbochs:
`GetObject` causes quite a lot of allocations due to `io.Copy` falling back to a generic implementation which allocates an internal buffer. The way to avoid this is to make sure the OS can use the sendfile `syscall` for zero copy transfers where possible.

To transfer using `sendfile`, you basically need to `io.Copy` between a `net.Conn` and an `os.File`, with no wrapper types around either object. The first change needed is to skip using a `SectionReader` when transferring a whole file (i.e. when there's no Range request).

```diff
@@ -2896,7 +2901,11 @@ func (p *Posix) GetObject(_ context.Context, input *s3.GetObjectInput) (*s3.GetO
                return nil, fmt.Errorf("open object: %w", err)
        }
 
-       rdr := io.NewSectionReader(f, startOffset, length)
+       var body io.ReadCloser = f
+       if startOffset != 0 || length != objSize {
+           rdr := io.NewSectionReader(f, startOffset, length)
+           body = &backend.FileSectionReadCloser{R: rdr, F: f}
+       }
 
        return &s3.GetObjectOutput{
                AcceptRanges:    &acceptRange,
```

The second issue is that fasthttp wraps the `net.Conn` in a `bufio.Writer`, in which case Go will never use `sendfile`, because the heuristic uses a very simple type assertion. See [here](https://github.com/valyala/fasthttp/issues/1889). I did work around this and confirmed if you bypass the bufio, `sendfile` starts working with `versitygw` and the high allocator load when serving files completely disappears.

(NOTE: for ranges, despite the underlying `sendfile` syscall supporting offset/size, I don't think Go has a way to signal that you want to use a partial `sendfile` for an `io.Copy` - the fast path only happens when sending an unwrapped `os.File` object with no ability afaict to limit the sending size)

benmcclelland:
one option here might be to write a modified io.Copy that allows the wrapped types for sendfile since we know what the types will be in these cases

lunixbochs:
As is this needs to be fixed with a fasthttp refactor, because they're writing into a bufio which can't even be unwrapped correctly.

My workaround was as follows:
- I patched my fasthttp to take an `interface { io.Writer, Flush() error }` for the whole relevant API instead of bufio.Writer (which is a subset with bufio.Writer anyway)
- at the point we'd write a response body stream I flush the bufio and swap in the actual socket

[...]

PR #921, opened 2024-10-26
make zero-copy GetObject possible via sendfile

lunixbochs:
part of https://github.com/versity/versitygw/issues/919 (but we still need to solve https://github.com/valyala/fasthttp/issues/1889)


versity-github:

benmcclelland (approved):

PR #922, opened 2024-10-26
ListObjects: don't walk parent directories

lunixbochs:
similar to https://github.com/versity/versitygw/pull/904, avoid walking parent directories that are above the prefix


versity-github:

benmcclelland (approved):

PR #926, opened 2024-10-28
Listbuckets pagination

### LINKED ISSUE #911: [Feature] - Implement pagination for ListBuckets
AWS S3 provides built-in pagination support for `ListBuckets` actions. Pagination is managed using the following parameters:
1. max-uploads
2. continuation-token

Additionally, S3 supports bucket filtering based on a prefix.

The goal is to implement this functionality in the gateway.

**Additional context**
[S3 docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListBuckets.html)
[AWS CLI docs](https://docs.aws.amazon.com/cli/latest/reference/s3api/list-buckets.html)

jonaustin09:
Closes #911

Implements pagination for the ListBuckets action both in posix and azure backends.
Removes double sorting by bucket name in posix, as os.ReadDir already gives sorted result.

### DIFF
--- a/tests/integration/utils.go
+++ b/tests/integration/utils.go
@@ -39,7 +39,6 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/aws/smithy-go"
 	"github.com/versity/versitygw/s3err"
-	"github.com/versity/versitygw/s3response"
 )
 
 var (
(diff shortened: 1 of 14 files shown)


tests/integration/utils.go
```diff
@@ -593,19 +592,13 @@ func areMapsSame(mp1, mp2 map[string]string) bool {
...
-	elementMap := make(map[string]bool)
-
-	for _, elem := range list1 {
-		elementMap[*elem.Name] = true
-	}
-
-	for _, elem := range list2 {
-		if _, found := elementMap[elem.Name]; !found {
+	for i, elem := range list1 {
+		if *elem.Name != *list2[i].Name {
```
benmcclelland, line 601: This requires lists to both be sorted the same. Is that an ok assumption here?

mlt180: Yes, exactly.

The comparison requires both lists to be sorted, as ListBuckets returns a sorted result. Based on this assumption, we also verify the order of the bucket names.

benmcclelland (approved):

PR #944, opened 2024-11-08
feat: Fixes all struct fields alignment issues, Adds betteralign chec…

### LINKED ISSUE #940: [Feature] - Integrate fieldalignment in the pipeline
1. Resolve all struct field alignment issues across the codebase.
2. Once resolved, add the `fieldalignment` check to the pipeline.

[fieldalignment link](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/fieldalignment)

jonaustin09:
Closes #940 

Improves struct field alignment to optimize memory usage.
Integrates `betteralign` check into the pipeline to ensure proper struct field alignment.

[betteralign](https://github.com/dkorunic/betteralign) is the improved version of `fieldalignment`, which doesn't modify any auto generated files. Test files modification is optional, configurable with the cli flags.


auth/iam_cache.go
```diff
@@ -36,14 +36,14 @@ type IAMCache struct {
...
 	exp   time.Time
+	value Account
 }
 
 type icache struct {
-	sync.RWMutex
-	expire time.Duration
 	items  map[string]item
+	expire time.Duration
+	sync.RWMutex
```
benmcclelland, line 46: I think we should still follow some good style guidelines here.  Embedded structs should still go first (https://github.com/uber-go/guide/blob/master/style.md#embedding-in-structs), and mutex should precede what they protect

auth/iam_internal.go
```diff
@@ -33,15 +33,15 @@ const (
...
 type IAMServiceInternal struct {
+	dir     string
+	rootAcc Account
 	// This mutex will help with racing updates to the IAM data
 	// from multiple requests to this gateway instance, but
 	// will not help with racing updates to multiple load balanced
 	// gateway instances. This is a limitation of the internal
 	// IAM service. All account updates should be sent to a single
 	// gateway instance if possible.
 	sync.RWMutex
```
benmcclelland, line 44: same here, mutex/embedded members first

backend/posix/posix.go
```diff
@@ -102,11 +102,11 @@ const (
 )
 
 type PosixOpts struct {
```
benmcclelland, line 104: Some of these structs like this one happen exactly once. Not sure if this is very important to optimize.

s3response/s3response.go
```diff
@@ -283,30 +283,30 @@ type Bucket struct {
...
+	IsAdmin           bool
 }
 
 type ListAllMyBucketsResult struct {
 	XMLName           xml.Name `xml:"http://s3.amazonaws.com/doc/2006-03-01/ ListAllMyBucketsResult" json:"-"`
 	Owner             CanonicalUser
-	Buckets           ListAllMyBucketsList
 	ContinuationToken string `xml:"ContinuationToken,omitempty"`
 	Prefix            string `xml:"Prefix,omitempty"`
+	Buckets           ListAllMyBucketsList
```
benmcclelland, line 297: order probably doesnt matter here, but do we want to try to keep with similar field order to aws?

s3response/s3response.go
```diff
@@ -163,9 +162,10 @@ type ListObjectsV2Result struct {
 	MaxKeys               *int32
 	Delimiter             *string
 	IsTruncated           *bool
+	XMLName               xml.Name `xml:"http://s3.amazonaws.com/doc/2006-03-01/ ListBucketResult" json:"-"`
```
benmcclelland, line 165: I think the xmlname member should be first for these always

s3response/s3response.go
```diff
@@ -128,32 +128,31 @@ type ListMultipartUploadsResult struct {
...
 type ListObjectsResult struct {
-	XMLName        xml.Name `xml:"http://s3.amazonaws.com/doc/2006-03-01/ ListBucketResult" json:"-"`
 	Name           *string
 	Prefix         *string
 	Marker         *string
 	NextMarker     *string
 	MaxKeys        *int32
 	Delimiter      *string
 	IsTruncated    *bool
+	XMLName        xml.Name `xml:"http://s3.amazonaws.com/doc/2006-03-01/ ListBucketResult" json:"-"`
```
benmcclelland, line 149: I think the xmlname member should be first for these always

s3response/s3response.go
```diff
@@ -97,23 +97,23 @@ func (o ObjectAttributes) IsValid() bool {
...
-	ObjectParts  *ObjectParts
+	ETag        *string
+	ObjectSize  *int64
+	ObjectParts *ObjectParts
 
 	// Not included in the response body
 	VersionId    *string
 	LastModified *time.Time
 	DeleteMarker *bool
+	StorageClass types.StorageClass `xml:",omitempty"`
```
benmcclelland, line 108: this got moved under "// Not included in the response body" which seems wrong?

s3log/webhook.go
```diff
@@ -33,8 +33,8 @@ import (
 
 // WebhookLogger is a webhook URL audit log
 type WebhookLogger struct {
-	mu  sync.Mutex
 	url string
+	mu  sync.Mutex
```
benmcclelland, line 37: mutex should come before member it protects

s3api/server_test.go
```diff
@@ -39,9 +39,9 @@ func TestNew(t *testing.T) {
 	port := ":7070"
 
 	tests := []struct {
-		name            string
-		args            args
 		wantS3ApiServer *S3ApiServer
+		args            args
+		name            string
```
benmcclelland, line 44: i usually see test name as first member fro readability in tests, seems thats better than the size optimization here?

benmcclelland (changes requested):
I think the better approach here is to identify structs that might be allocated often and target those for optimization.  Structs like the backend options are allocated once, so I would much prefer readability over save a couple bytes for these. Is there a way to have a list of ones we care about to have tests validate?

PR #958, opened 2024-11-19
Test/rest get bucket ownership controls

lrm25:


tests/test_s3api_policy_bucket.sh
```diff
@@ -0,0 +1,232 @@
...
+
+  setup_user "$username" "$password" "user" || fail "error creating user"
+
+  run setup_bucket "s3api" "$BUCKET_ONE_NAME"
+  assert_success
+
+  if delete_bucket_policy_with_user "$BUCKET_ONE_NAME" "$username" "$password"; then
+    fail "able to delete bucket policy with user $username without right permissions"
+  fi
+  setup_policy_with_single_statement "$TEST_FILE_FOLDER/$policy_file" "dummy" "$effect" "$principal" "$action" "$resource" || fail "failed to set up policy"
```
anodelman: Any reason this, and the commands below it, aren't wrapped in a `run`?

tests/test_s3api_policy_bucket.sh
```diff
@@ -0,0 +1,232 @@
...
+
+  setup_user "$username" "$password" "user" || fail "error creating user"
+
+  run setup_bucket "s3api" "$BUCKET_ONE_NAME"
+  assert_success
+
+  if get_bucket_acl_with_user "$BUCKET_ONE_NAME" "$username" "$password"; then
+    fail "user able to get bucket ACLs despite permissions"
+  fi
+  setup_policy_with_single_statement "$TEST_FILE_FOLDER/$policy_file" "dummy" "$effect" "$principal" "$action" "$resource" || fail "failed to set up policy"
```
anodelman: Same question here about wrapping in `run`.

anodelman (changes requested):
A lot of ` | fail` showing up in tests instead of run/assert - any reason?  Just an oversite?

tests/test_s3api_policy_bucket.sh
lrm25: These are old, unconverted functions.  Just broke up a large file, and didn't change some of the fails to run/assserts in it yet.  I've gradually been doing this commit by commit.  Can change some more of these ones now if you want though.

lrm25: Changed them for this file.  Let me know if there are any other issues.

PR #967, opened 2024-11-26
Test/rest put bucket policy

lrm25:


tests/test_s3api_multipart.sh
```diff
@@ -15,27 +15,112 @@
...
+  local expected_tag_key="TestTag"
+  local expected_tag_val="TestTagVal"
+
+  os_name="$(uname)"
+  if [[ "$os_name" == "Darwin" ]]; then
+    now=$(date -u +"%Y-%m-%dT%H:%M:%S")
+    later=$(date -j -v +15S -f "%Y-%m-%dT%H:%M:%S" "$now" +"%Y-%m-%dT%H:%M:%S")
+  else
+    now=$(date +"%Y-%m-%dT%H:%M:%S")
+    later=$(date -d "$now 15 seconds" +"%Y-%m-%dT%H:%M:%S")
```
anodelman: Does this need to be $now + 15 seconds?  Looks like it is missing the addition symbol.

tests/test_s3api_policy_object.sh
```diff
@@ -286,67 +310,76 @@ test_s3api_policy_put_wildcard() {
...
   run create_test_folder "$test_folder"
   assert_success
 
   run create_test_files "$test_folder/$test_file" "$policy_file"
   assert_success
 
+  run setup_user_versitygw_or_direct "$USERNAME_ONE" "$PASSWORD_ONE" "user" "$BUCKET_ONE_NAME"
+  assert_success
+  # shellcheck disable=SC2154
+  username=${lines[0]}
```
anodelman, line 323: Is this assumed the setup_user_versitygw_or_direct returns a variable called "lines"?  This could be a direct return from the function and would make the code more readable.

anodelman (changes requested):

tests/test_s3api_policy_object.sh
lrm25: "lines" is a variable returned by bats (output lines in an array).  Probably could do a direct (echo) return though, if you think it's necessary and/or better.

tests/test_s3api_multipart.sh
lrm25: No, that's the linux format for adding time.  It works on my Linux machine.  The addition symbol actually subtracts 15 hours for some reason.

anodelman (dismissed):
Thanks for the clarifications!

benmcclelland (approved):

PR #981, opened 2024-12-10
fix: Sets uploadable files limit to 5 GiB for PutObject and UploadPar…

### LINKED ISSUE #961: [Bug] - UploadPart max size limit
**Describe the bug**
If uploading a part with a size greater than `5 GiB`, the `UploadPart` operation should fail.

**To Reproduce**
```bash

aws s3api create-multipart-upload --bucket test --key obj

// file_1 should be greater than 5Gib
aws s3api upload-part --bucket test --key obj --upload-id <mp_upload_id> --part-number 1 --body file_1
```

**Expected behavior**
`UploadPart` should return the following error:
```
An error occurred (EntityTooLarge) when calling the UploadPart operation: Your proposed upload exceeds the maximum allowed size
```

niksis02:
Fixes #961

Sets an upper limit of 5 GiB (as specified in the AWS documentation) on incoming data for large data actions such as PutObject and UploadPart.
Since wrapBodyReader is used to combine the request body reader with other io.Reader instances, a new io.Reader is layered on top to enforce the data size limit.

### DIFF
--- a/tests/integration/tests.go
+++ b/tests/integration/tests.go
@@ -40,6 +40,7 @@ import (
 
 var (
 	shortTimeout  = 10 * time.Second
+	longTimeout   = 40 * time.Second // for large files upload
 	iso8601Format = "20060102T150405Z"
 	nullVersionId = "null"
 )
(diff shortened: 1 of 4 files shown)


versity-github:

tests/integration/tests.go
```diff
@@ -3088,6 +3089,23 @@ func PutObject_with_object_lock(s *S3Conf) error {
 	return nil
 }
 
+func PutObject_5gib_exceeding_file(s *S3Conf) error {
+	testName := "PutObject_5gib_exceeding_file"
+	return actionHandler(s, testName, func(s3client *s3.Client, bucket string) error {
+		obj := "my-obj"
+
+		_, err := putObjectWithData(5*1024*1024*1024+1, &s3.PutObjectInput{ // more than 5 Gib
```
benmcclelland, line 3097: I think this is too large for github actions to handle. Maybe we need a separate test group for larger tests that we cant run automated in actions?

PR #983, opened 2024-12-11
fix: Return XAdminMethodNotSupported error for single IAM methods.

niksis02:
In single `IAM` user mode, all IAM methods return a 'not supported' error. 
This update introduces an admin-specific error for this scenario.


versity-github:

PR #988, opened 2024-12-14
fix: Adds the minimum allowed size(5 Mib) check for mp parts sizes in…

### LINKED ISSUE #960: [Bug] - CompleteMultipartUpload lower limit of uploading parts size
**Describe the bug**
When initiating a multipart upload, all uploaded parts should have the same size (except for the last part) and must be between `5 MiB` and `5 GiB`. If more than two parts are uploaded and any part is smaller than 5 MiB, the `CompleteMultipartUpload` operation should fail.

**To Reproduce**
```bash

aws s3api create-multipart-upload --bucket test --key obj

// file_1 should be smaller than 5Mib
aws s3api upload-part --bucket test --key obj --upload-id <mp_upload_id> --part-number 1 --body file_1

// file_2 should be smaller than 5Mib
aws s3api upload-part --bucket test --key obj --upload-id <mp_upload_id> --part-number 2 --body file_2

aws s3api complete-multipart-upload --bucket test --key obj --upload-id <mp_upload_id> --multipart-upload '{"Parts": [{"ETag": "<file_1_ETag>", "PartNumber": 1}, {"ETag": "file_2_ETag", "PartNumber": 2}]}'
```

**Expected behavior**
`CompleteMultipartUpload` should return the following error:
```
An error occurred (EntityTooSmall) when calling the CompleteMultipartUpload operation: Your proposed upload is smaller than the minimum allowed size
```

niksis02:
Fixes #960 

In multipart uploads, the size of uploaded parts must be lower than 5 MiB, except for the final part.

This check has been added for the `CompleteMultipartUpload` operation in the `Azure` and `POSIX` backends.


versity-github:

niksis02:
@lrm25 Could you check to fix the bats tests due to the new changes here ?

PR #1005, opened 2024-12-25
Implement IPA IAM backend

yhal-nesi:
[FreeIPA](https://www.freeipa.org) is an identity management system for Linux, similar to AD. It supports an API over HTTP allowing user retrieval. FreeIPA doesn't store passwords but it has an ability to store secrets in a [password vault](https://www.freeipa.org/page/V4/Password_Vault_1.0).

Example usage:
```
 versitygw --debug --ipa-host https://ipa.example.test --ipa-vault-name versity --ipa-user admin --ipa-password hello_world
```

For each user it expects a vault with a name set by --ipa-vault-name, and retrieves UID and GID. Only GetUserAccount is implemented, it is expected all accounts are managed with IPA. 

User retrieval is simple, just a call to user_show/1. Secret retrieval is more complex because IPA has extra level of encryption:
* Versity generates a session key and encrypts it with IPA key recovery authority public key
* that encrypted key is passed to IPA
* IPA returns vault data encrypted with a session key and negotiated algorithm. Versity decrypts the value using the session key. I've only implemented aes-128-cbc which is a default in current version, but will probably need additional algorithms in the future.


versity-github:

auth/iam_ipa.go
```diff
@@ -0,0 +1,394 @@
...
+	"net/http"
+	"net/http/cookiejar"
+	"net/url"
+	"strconv"
+	"strings"
+)
+
+const IpaVersion = "2.254"
+
+type ipaIAMService struct {
```
benmcclelland: make this type exported so that the New() function can return the struct instead of the interface type

auth/iam_ipa.go
```diff
@@ -0,0 +1,394 @@
...
+	host            string
+	vaultName       string
+	username        string
+	password        string
+	kraTransportKey *rsa.PublicKey
+	debug           bool
+	rootAcc         Account
+}
+
+//var _ IAMService = &ipaIAMService{}
```
benmcclelland: its fine to leave this uncommented for a sanity check on the the interface

auth/iam_ipa.go
```diff
@@ -0,0 +1,394 @@
...
+
+func NewIpaIAMService(rootAcc Account, host, vaultName, username, password string, isInsecure, debug bool) (IAMService, error) {
```
benmcclelland: return *IpaIAMService here instead of IAMService

auth/iam_ipa.go
```diff
@@ -0,0 +1,394 @@
...
+		vaultName: vaultName,
+		username:  username,
+		password:  password,
+		debug:     debug,
+		rootAcc:   rootAcc,
+	}
+	jar, err := cookiejar.New(nil)
+	if err != nil {
+		// this should never happen
+		panic(err)
```
benmcclelland: return the error instead of panic

auth/iam_ipa.go
```diff
@@ -0,0 +1,394 @@
...
+
+	mTLSConfig := &tls.Config{InsecureSkipVerify: isInsecure}
+	tr := &http.Transport{
+		TLSClientConfig: mTLSConfig,
+	}
+	ipa.client = http.Client{Jar: jar, Transport: tr}
+
+	err = ipa.login()
+	if err != nil {
+		return nil, err
```
benmcclelland: when returning the errors, wrap the error with more context, for example:
```go
return nil, fmt.Errorf("login: %w", err)
```

auth/iam_ipa.go
```diff
@@ -0,0 +1,394 @@
...
+		Gidnumber []string
+		Uidnumber []string
+	}{}
+	_, err := ipa.rpc(req, &userResult)
+	if err != nil {
+		return Account{}, err
+	}
+
+	uid, _ := strconv.Atoi(userResult.Uidnumber[0])
+	gid, _ := strconv.Atoi(userResult.Gidnumber[0])
```
benmcclelland: since these are returned as strings, it would be best to not ignore the errors here

auth/iam_ipa.go
```diff
@@ -0,0 +1,394 @@
...
+		Role:    RoleUser,
+		UserID:  uid,
+		GroupID: gid,
+	}
+
+	session_key := make([]byte, 16)
+	rand.Read(session_key)
+	encrypted_key, err := rsa.EncryptPKCS1v15(rand.Reader, ipa.kraTransportKey, session_key)
+	if err != nil {
+		return account, err
```
benmcclelland: wrap returned error with more context

auth/iam_ipa.go
```diff
@@ -0,0 +1,394 @@
...
+	data := struct {
+		Vault_data Base64EncodedWrapped
+		Nonce      Base64EncodedWrapped
+	}{}
+	_, err = ipa.rpc(req, &data)
+	if err != nil {
+		return account, err
+	}
+
+	aes, _ := aes.NewCipher(session_key)
```
benmcclelland: dont skip error check here

auth/iam_ipa.go
```diff
@@ -0,0 +1,394 @@
...
+	cbc := cipher.NewCBCDecrypter(aes, data.Nonce)
+	cbc.CryptBlocks(data.Vault_data, data.Vault_data)
+	secret_unpadded_json, _ := pkcs7Unpad(data.Vault_data, 16)
```
benmcclelland: dont skip error check here

auth/iam_ipa.go
```diff
@@ -0,0 +1,394 @@
...
+	return []Account{}, fmt.Errorf("not implemented")
+}
+
+func (ipa *ipaIAMService) Shutdown() error {
+	return nil
+}
+
+// Implementation
+
+func (ipa *ipaIAMService) login() error {
```
benmcclelland: shouldn't the login be returning some sort of token or something for future requests?

auth/iam_ipa.go
```diff
@@ -0,0 +1,394 @@
...
+func (ipa *ipaIAMService) newRequest(method string, args []string, dict map[string]any) rpcRequest {
+
+	id := ipa.id
+	ipa.id++
+
+	dict["version"] = ipa.version
+
+	jmethod, _ := json.Marshal(method)
+	jargs, _ := json.Marshal(args)
+	jdict, _ := json.Marshal(dict)
```
benmcclelland: best not to skip errors, and return them instead

auth/iam_ipa.go
```diff
@@ -0,0 +1,394 @@
...
+	err := ipa.login()
+	if err != nil {
+		return rpcResponse{}, err
+	}
+
+	res, err := ipa.rpcInternal(req)
+	if err != nil {
+		return res, err
+	}
+	err = json.Unmarshal(res.Result, value)
```
benmcclelland, line 262: is this correct that the res.Result should be whatever is provided to the rpc method? I would have expected this should come from the response?

auth/iam_ipa.go
benmcclelland: is data ever populated with real values? it looks like rpc() doesnt actually modify data, and above is just creating these with 0 values?

auth/iam_ipa.go
yhal-nesi: Full rpc response isn't really used anywhere (it includes things like ipa version etc.). I left it in rpcInternal signature in case it is needed in the future and removed it from rpc(...) signature for now.

auth/iam_ipa.go
yhal-nesi: I added error propagation in all other places but in this instance marshal can only cause error in very limited circumstances. From documentation:
`// Floating point, integer, and [Number] values encode as JSON numbers.
// NaN and +/-Inf values will return an [UnsupportedValueError].
...
// Channel, complex, and function values cannot be encoded in JSON.
// Attempting to encode such a value causes Marshal to return
// an [UnsupportedTypeError].`

this can never happen with strings (lines 317 and 318). It can happen with dict but maybe there is more elegant way to prevent this without passing errors? 

This is my first time writing anything substantial in Go, so maybe my aversion to "if err != nill return err" is not justified... :).

auth/iam_ipa.go
yhal-nesi: It does, it un-marshals value in rpc call. I updated that method signature to avoid confusion.

yhal-nesi:
@benmcclelland thank you for review! I think all issues except error handling in newRequest are addressed. 
There another comment I cannot find now, about login needing to return a token. Login sets a cookie in http.Client. That's how IPA HTTP authentication works.

auth/iam_ipa.go
yhal-nesi: Login sets a cookie in http.Client.

auth/iam_ipa.go
benmcclelland: "if err != nill return err" feels tedious to write over and over, but its really awesome when debugging an issue. This will report a very clear error of exactly what happened, and I can often debug issues by just looking at logs or error returns. Even if it seems trivial/unnecessary, I usually go ahead and handle all errors just in case some malformed string or something happens to get into the code paths.

benmcclelland: And even better is to wrap it with some context, so I use the pattern:
```go
if err != nil {
  return fmt.Errorf("<description of context>: %w", err)
}
```
everywhere. The wrapping allows the callers to unwrap the error and check for the originating error type if needed.

yhal-nesi:
All review comments should now be addressed.

benmcclelland (approved):

PR #1010, opened 2024-12-31
Test/rest upload part copy

lrm25:


tests/rest_scripts/rest.sh
```diff
@@ -65,3 +65,18 @@ $canonical_request_hash"
...
+add_parameter() {
+  if [ "$#" -ne 2 ]; then
+    return
+  fi
+  if [ "$first_param_added" != "true" ]; then
+    if [ "$1" == "url" ]; then
+      eval "$1"+='?'
+    fi
+    eval "$1"+="$2"
+    first_param_added="true"
```
anodelman: Having this global var controlled outside the function seems really risky to me - it assumes a 'correct' way to use this and could generate weird bugs when you don't reset it.  Maybe you could parse the url to see what state it is in to correctly add first/additional param?

anodelman (changes requested):

tests/rest_scripts/rest.sh
lrm25: Yeah, I'll try to fix this

anodelman (dismissed):

benmcclelland (approved):

PR #1023, opened 2025-01-13
Implements object integrity checksums

### LINKED ISSUE #928: [Bug] - GetObjectAttributes doesn't return checksum when Checksum is requested
**Describe the bug**
A direct-to-s3 GetObjectAttributes call requesting the `Checksum` field returns an enclosing `Checksum` tag containing the specific checksum type used and then the checksum itself, e.g.:
```
<GetObjectAttributesResponse xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Checksum><ChecksumSHA256>42SQpSFdSKg4uk5vh1UEqrGuG7vEddCYqxblLEeyW9k=</ChecksumSHA256></Checksum></GetObjectAttributesResponse>
```

This checksum info is not returned by versitygw.

**To Reproduce**
1. Run versitygw, e.g.:  `./versitygw --access=******** --secret=******** --region=us-east-1 --cert /Users/lukemccrone/devel/versitygw/cert.pem --key /Users/lukemccrone/devel/versitygw/versitygw.pem posix --versioning-dir /tmp/versioning /tmp/gw`
2. Create bucket, e.g.:  `aws --no-verify-ssl s3api create-bucket --bucket versity-gwtest-bucket-one-1`
3. Use script at https://github.com/versity/versitygw/blob/test/rest_get_obj_attributes/tests/rest_scripts/put_object.sh to send command with sha256 checksum.  It should look like this:  `curl -ks -w %{http_code} -X PUT https://localhost:7070/versity-gwtest-bucket-one-1/test_file -H Authorization: AWS4-HMAC-SHA256 Credential=********/20241029/us-east-1/s3/aws4_request,SignedHeaders=host;x-amz-content-sha256;x-amz-date;x-amz-checksum-sha256,Signature=122f41b26e6b6253bae6509f419e72dfd232d807df563f02c47a79b4f1baf552 -H x-amz-content-sha256: 92f6c55b20aaa27a3185f798e8454265318c8732e91c598f1b2cb66932c4f735 -H x-amz-date: 20241029T140541Z -H x-amz-checksum-sha256: kvbFWyCqonoxhfeY6EVCZTGMhzLpHFmPGyy2aTLE9zU= -T /Users/lukemccrone/devel/versitygw/versity-gwtest-files/test_file -o /Users/lukemccrone/devel/versitygw/versity-gwtest-files/result.txt`
4. Get attributes with **Checksum**, e.g.:  `curl -ks -w %{http_code} https://localhost:7070/versity-gwtest-bucket-one-1/test_file?attributes= -H Authorization: AWS4-HMAC-SHA256 Credential=********/20241029/us-east-1/s3/aws4_request,SignedHeaders=host;x-amz-content-sha256;x-amz-date;x-amz-object-attributes,Signature=faadfdb5c9670cd03fb39c8993ed90f7d5345d315598d8d310b5e45258cd3137 -H x-amz-content-sha256: UNSIGNED-PAYLOAD -H x-amz-date: 20241029T140541Z -H x-amz-object-attributes: Checksum -o /Users/lukemccrone/devel/versitygw/versity-gwtest-files/attributes.txt`

**Expected behavior**
sha256 checksum is properly returned

**Server Version**
Version  : v1.0.8
Build    : d788010
BuildTime: 2024-10-29_02:05:18PM
Darwin Lukes-MacBook-Pro.local 23.4.0 Darwin Kernel Version 23.4.0: Fri Mar 15 00:12:25 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T6030 arm64

niksis02:
Fixes #928 

Implements object integrity checksums in `POSIX`.  

The following checksum algorithms are supported (similar to AWS):  
1. **CRC32**  
2. **CRC32C**  
3. **SHA1**  
4. **SHA256**  
5. **CRC64NVME**

The checksums are Base64-encoded.  
- Checksums are calculated during object uploads (`PutObject`, `UploadPart`).  
- A checksum reader is layered on top of the existing readers.  
- When an object is copied (`CopyObject`), the checksum is recalculated if a different checksum type is specified.  

**Constraints:**  
- Multiple checksum types are not allowed for a single object.  
- It is not permitted to upload an object (`PutObject`) or its parts with differing checksum types.  

Checksum data is stored as extended attributes (`xattr`) in JSON format, including the checksum algorithm and its corresponding values. 

The new `checksum-type` support is added in the gateway as stated in [aws docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html)

There are 2 checksum types:
1. **COMPOSITE**
2. **FULL_OBJECT**

`FULL_OBJECT` is calculated based on the object data from the first till the last byte.
`COMPOSITE` checksum is used in multipart uploads, which is calculated based on all the uploaded parts checksums.

All the new changes mentioned in the doc will become available with this PR in the `posix` and `s3proxy` backends

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"net/http"
 	"os"
 	"path/filepath"
 	"sort"
@@ -39,6 +40,7 @@ import (
 	"github.com/versity/versitygw/auth"
 	"github.com/versity/versitygw/backend"
 	"github.com/versity/versitygw/backend/meta"
+	"github.com/versity/versitygw/s3api/utils"
 	"github.com/versity/versitygw/s3err"
 	"github.com/versity/versitygw/s3response"
 )
@@ -87,6 +89,7 @@ const (
 	aclkey              = "acl"
 	ownershipkey        = "ownership"
 	etagkey             = "etag"
+	checksumsKey        = "checksums"
 	policykey           = "policy"
 	bucketLockKey       = "bucket-lock"
 	objectRetentionKey  = "object-retention"
@@ -1308,6 +1311,21 @@ func (p *Posix) CreateMultipartUpload(ctx context.Context, mpu *s3.CreateMultipa
 		}
 	}
 
+	// Set object checksum algorithm
+	if mpu.ChecksumAlgorithm != "" {
+		err := p.storeChecksums(nil, bucket, filepath.Join(objdir, uploadID), s3response.Checksum{
+			Algorithms: []types.ChecksumAlgorithm{
+				mpu.ChecksumAlgorithm,
+			},
+		})
+		if err != nil {
+			// cleanup object if returning error
+			os.RemoveAll(filepath.Join(tmppath, uploadID))
+			os.Remove(tmppath)
+			return s3response.InitiateMultipartUploadResult{}, fmt.Errorf("store mp checksum algorithm: %w", err)
+		}
+	}
+
 	return s3response.InitiateMultipartUploadResult{
 		Bucket:   bucket,
 		Key:      object,
@@ -1373,6 +1391,28 @@ func (p *Posix) CompleteMultipartUpload(ctx context.Context, input *s3.CompleteM
 
 	objdir := filepath.Join(metaTmpMultipartDir, fmt.Sprintf("%x", sum))
 
+	checksums, err := p.retreiveChecksums(nil, bucket, filepath.Join(objdir, uploadID))
+	if err != nil && !errors.Is(err, meta.ErrNoSuchKey) {
+		return nil, fmt.Errorf("get mp checksums: %w", err)
+	}
+
+	// if len(checksums.Algorithms) != 0 {
+	// 	algorithm := checksums.Algorithms[0]
+
+	// 	if input.ChecksumCRC32 != nil && algorithm != types.ChecksumAlgorithmCrc32 {
+	// 		return nil, s3err.GetInvalidChecksumHeaderErr("x-amz-checksum-crc32")
+	// 	}
+	// 	if input.ChecksumCRC32C != nil && algorithm != types.ChecksumAlgorithmCrc32c {
+	// 		return nil, s3err.GetInvalidChecksumHeaderErr("x-amz-checksum-crc32c")
+	// 	}
+	// 	if input.ChecksumSHA1 != nil && algorithm != types.ChecksumAlgorithmSha1 {
+	// 		return nil, s3err.GetInvalidChecksumHeaderErr("x-amz-checksum-sha1")
+	// 	}
+	// 	if input.ChecksumSHA256 != nil && algorithm != types.ChecksumAlgorithmSha256 {
+	// 		return nil, s3err.GetInvalidChecksumHeaderErr("x-amz-checksum-sha256")
+	// 	}
+	// }
+
 	// check all parts ok
 	last := len(parts) - 1
 	var totalsize int64
@@ -1403,6 +1443,25 @@ func (p *Posix) CompleteMultipartUpload(ctx context.Context, input *s3.CompleteM
 		if parts[i].ETag == nil || etag != *parts[i].ETag {
 			return nil, s3err.GetAPIError(s3err.ErrInvalidPart)
 		}
+
+		partChecksum, err := p.retreiveChecksums(nil, bucket, partObjPath)
+		if err != nil && !errors.Is(err, meta.ErrNoSuchKey) {
+			return nil, fmt.Errorf("get part checksum: %w", err)
+		}
+
+		// If checksum has been provided on mp initalization
+		err = validatePartChecksum(partChecksum, part)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	var hashRdr *utils.HashReader
+	if len(checksums.Algorithms) != 0 {
+		hashRdr, err = utils.NewHashReader(nil, "", utils.HashType(strings.ToLower(string(checksums.Algorithms[0]))))
+		if err != nil {
+			return nil, fmt.Errorf("initialize hash reader: %w", err)
+		}
 	}
 
 	f, err := p.openTmpFile(filepath.Join(bucket, metaTmpDir), bucket, object,
@@ -1426,7 +1485,14 @@ func (p *Posix) CompleteMultipartUpload(ctx context.Context, input *s3.CompleteM
 		if err != nil {
 			return nil, fmt.Errorf("open part %v: %v", *part.PartNumber, err)
 		}
-		_, err = io.Copy(f.File(), pf)
+
+		var rdr io.Reader = pf
+		if hashRdr != nil {
+			hashRdr.SetReader(rdr)
+			rdr = hashRdr
+		}
+
+		_, err = io.Copy(f.File(), rdr)
 		pf.Close()
 		if err != nil {
 			if errors.Is(err, syscall.EDQUOT) {
@@ -1524,6 +1590,53 @@ func (p *Posix) CompleteMultipartUpload(ctx context.Context, input *s3.CompleteM
 		}
 	}
 
+	var crc32 *string
+	var crc32c *string
+	var sha1 *string
+	var sha256 *string
+
+	// set checksum
+	if hashRdr != nil {
+		algo := checksums.Algorithms[0]
+		checksum := s3response.Checksum{
+			Algorithms: []types.ChecksumAlgorithm{
+				algo,
+			},
+		}
+
+		sum := hashRdr.Sum()
+		switch hashRdr.Type() {
+		case utils.HashTypeCRC32:
+			if input.ChecksumCRC32 != nil && *input.ChecksumCRC32 != sum {
+				return nil, s3err.GetChecksumBadDigestErr(algo)
+			}
+			checksum.CRC32 = &sum
+			crc32 = &sum
+		case utils.HashTypeCRC32C:
+			if input.ChecksumCRC32C != nil && *input.ChecksumCRC32C != sum {
+				return nil, s3err.GetChecksumBadDigestErr(algo)
+			}
+			checksum.CRC32C = &sum
+			crc32c = &sum
+		case utils.HashTypeSha1:
+			if input.ChecksumSHA1 != nil && *input.ChecksumSHA1 != sum {
+				return nil, s3err.GetChecksumBadDigestErr(algo)
+			}
+			checksum.SHA1 = &sum
+			sha1 = &sum
+		case utils.HashTypeSha256:
+			if input.ChecksumSHA256 != nil && *input.ChecksumSHA256 != sum {
+				return nil, s3err.GetChecksumBadDigestErr(algo)
+			}
+			checksum.SHA256 = &sum
+			sha256 = &sum
+		}
+		err := p.storeChecksums(f.File(), bucket, object, checksum)
+		if err != nil {
+			return nil, fmt.Errorf("store object checksum: %w", err)
+		}
+	}
+
 	// load and set retention
 	ret, err := p.meta.RetrieveAttribute(nil, bucket, upiddir, objectRetentionKey)
 	if err != nil && !errors.Is(err, meta.ErrNoSuchKey) {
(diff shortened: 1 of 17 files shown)


niksis02:
Azure tests failure is intentional. All the failures will be gone after the next commit, implementing checksums in the azure backend.

backend/posix/posix.go
```diff
@@ -1308,6 +1311,22 @@ func (p *Posix) CreateMultipartUpload(ctx context.Context, mpu *s3.CreateMultipa
...
+		err := p.storeChecksums(nil, bucket, filepath.Join(objdir, uploadID), s3response.Checksum{
+			Algorithms: []types.ChecksumAlgorithm{
+				mpu.ChecksumAlgorithm,
+			},
+			Type: &mpu.ChecksumType,
+		})
+		if err != nil {
+			// cleanup object if returning error
+			os.RemoveAll(filepath.Join(tmppath, uploadID))
+			os.Remove(tmppath)
```
chaowang-versity: check the returned error or explicitly ignore the error ` _ = os.Remove()`

backend/posix/posix.go
```diff
@@ -1334,6 +1353,29 @@ func (p *Posix) getChownIDs(acct auth.Account) (int, int, bool) {
...
+		return ""
+	}
+
+	return *t
+}
+
+func getPartChecksum(algo types.ChecksumAlgorithm, part types.CompletedPart) string {
+	switch algo {
+	case types.ChecksumAlgorithmCrc32:
+		return *part.ChecksumCRC32
```
chaowang-versity: Could we check if all the pointers are nil before defering them?  I also wonder if we can change the type of all the `part.ChecksumCRC32` , `part.ChecksumCRC32c`, etc from `*string` to `*types.ChecksumType`, so that we can call `getChecksumType()` here to avoid the nil pointer check.

backend/posix/posix.go
```diff
@@ -1334,6 +1353,29 @@ func (p *Posix) getChownIDs(acct auth.Account) (int, int, bool) {
 	return uid, gid, needsChown
 }
 
+func getChecksumType(t *types.ChecksumType) types.ChecksumType {
```
chaowang-versity: I would recommend making this function a method of `types.ChecksumType`

```suggestion 
func (t *types.ChecksumType) Get() {
```

chaowang-versity: I am not sure this is correct, but based on the line 1431 , if `t==nil` , should it return `null` instead of ""?

backend/posix/posix.go
```diff
@@ -1373,6 +1415,25 @@ func (p *Posix) CompleteMultipartUpload(ctx context.Context, input *s3.CompleteM
 
 	objdir := filepath.Join(metaTmpMultipartDir, fmt.Sprintf("%x", sum))
 
+	checksums, err := p.retreiveChecksums(nil, bucket, filepath.Join(objdir, uploadID))
+	if err != nil && !errors.Is(err, meta.ErrNoSuchKey) {
+		return nil, fmt.Errorf("get mp checksums: %w", err)
+	}
+	var checksumAlgorithm types.ChecksumAlgorithm
```
chaowang-versity, line 1412: Should the default value of `checksumAlgorithm` be empty string or `HashTypeNone`?

backend/posix/posix.go
```diff
@@ -1524,6 +1623,67 @@ func (p *Posix) CompleteMultipartUpload(ctx context.Context, input *s3.CompleteM
...
+		switch checksumType {
+		case types.ChecksumTypeComposite:
+			sum = compositeChecksumRdr.Sum()
+		case types.ChecksumTypeFullObject:
+			sum = hashRdr.Sum()
+		}
+
+		switch checksumAlgorithm {
+		case types.ChecksumAlgorithmCrc32:
+			if input.ChecksumCRC32 != nil && *input.ChecksumCRC32 != sum {
```
chaowang-versity, line 1646: Seems like the `checksumAlgorithm` is retrived from ` p.retreiveChecksums` and `input` is passed in as parameter.  What happen if `input` has a different checksum type from `checksumAlgorithm`?
Should it check if `input.ChecksumType` matches `checksumAlgorithm` and retrive the the checksum based on `input.ChecksumType` (I feel like there should be a method on the `input` to return the checksum as a string based on `input.ChecksumType` )?

backend/posix/posix.go
```diff
@@ -1556,13 +1716,152 @@ func (p *Posix) CompleteMultipartUpload(ctx context.Context, input *s3.CompleteM
...
+		Bucket:            &bucket,
+		ETag:              &s3MD5,
+		Key:               &object,
+		VersionId:         &versionID,
+		ChecksumCRC32:     crc32,
+		ChecksumCRC32C:    crc32c,
+		ChecksumSHA1:      sha1,
+		ChecksumSHA256:    sha256,
+		ChecksumCRC64NVME: crc64nvme,
+		ChecksumType:      checksumType,
```
chaowang-versity: I think it is mentioned that only one type of check is supported for one upload.  Instead of saving all the possible checksum types, can you save the check as a string since the type information is already saved in `ChecksumType`?
```suggestion
		Bucket:            &bucket,
		ETag:              &s3MD5,
		Key:               &object,
		VersionId:         &versionID,
		Checksum: checksum,
		ChecksumType:      checksumType,
```

backend/posix/posix.go
```diff
@@ -1556,13 +1716,152 @@ func (p *Posix) CompleteMultipartUpload(ctx context.Context, input *s3.CompleteM
...
 	}, nil
 }
 
+func validatePartChecksum(checksum s3response.Checksum, part types.CompletedPart) error {
+	n := numberOfChecksums(part)
```
chaowang-versity, line 1728: move 
```
if n > 1 {
	return s3err.GetAPIError(s3err.ErrInvalidChecksumPart)
}
```
		
right after this line

backend/posix/posix.go
```diff
@@ -1556,13 +1716,152 @@ func (p *Posix) CompleteMultipartUpload(ctx context.Context, input *s3.CompleteM
...
+
+	if len(checksum.Algorithms) != 0 {
```
chaowang-versity: nit: this a minor style thing but it makes indentation better if we do early return by 
```
 if len(checksum.Algorithms) = 0 {
   if n != 0 {
		return s3err.GetAPIError(s3err.ErrInvalidPart)
	}
   return nil
 }
 ```
 so that you don't have to indent the rest of the code

File: backend/posix/posix.go [hunk 9]
```diff
@@ -1556,13 +1716,152 @@ func (p *Posix) CompleteMultipartUpload(ctx context.Context, input *s3.CompleteM
...
+				Code:           "InvalidRequest",
+				Description:    fmt.Sprintf("The upload was created using a %v checksum. The complete request must include the checksum for each part. It was missing for part %v in the request.", strings.ToLower(string(algo)), *part.PartNumber),
+				HTTPStatusCode: http.StatusBadRequest,
+			}
+		}
+		if n > 1 {
+			return s3err.GetAPIError(s3err.ErrInvalidChecksumPart)
+		}
+
+		if part.ChecksumCRC32 != nil {
```
chaowang-versity: Since the `checksum` has the checksum algorithm in `checksum.Algorithms[0]` , can we put the algorithm in `part` as well.  Update the `checksum` and `part` to have a single string for the checksum value instead of multiple string pointers. So that instead of checking each pointer for all five algorithms , we can simply do two check , 1. check `checksum. Algorithms[0] == part.Algorithms`, 2. check `checksum.ChecksumStr == part.ChecksumStr`. 
I think by using a specific type field and a single checksum string, most of the code can be simplified.

chaowang-versity: Even if the suggestions above is not feasible, you can use a loop to refactor this code to make it concise:
```
for _, cs := range []struct {
    partChecksum   *string
    algoType       types.ChecksumAlgorithm
    expectedChecksum string
    algoName       string
}{
    {part.ChecksumCRC32, types.ChecksumAlgorithmCrc32, getString(checksum.CRC32), "crc32"},
    {part.ChecksumCRC32C, types.ChecksumAlgorithmCrc32c, getString(checksum.CRC32C), "crc32c"},
    // add the other three ...

} {
    if cs.partChecksum == nil {
        continue
    }
    
    if ok := utils.IsValidChecksum(*cs.partChecksum, cs.algoType); !ok {
        return s3err.GetAPIError(s3err.ErrInvalidChecksumPart)
    }

    if *cs.partChecksum != cs.expectedChecksum {
        if algo == cs.algoType {
            return s3err.GetAPIError(s3err.ErrInvalidPart)
        }
        return s3err.APIError{
            Code:           "BadDigest",
            Description:    fmt.Sprintf("The %s you specified for part %v did not match what we received.", cs.algoName, *part.PartNumber),
            HTTPStatusCode: http.StatusBadRequest,
        }
    }
}
```

backend/posix/posix.go
niksis02: types.ChecksumType comes from aws sdk, we can't attach any method to it

backend/posix/posix.go
niksis02: ChecksumCRC32, ,ChecksumCRC32C ... properties can't be a type of *types.ChecksumType as they can be any string and it makes sense to have them as string pointers

backend/posix/posix.go
niksis02: input.ChecksumType and input.ChecksumAlgorithm props validation is in the front-end. So here in the backend we are sure that they are valid and match. 
I guess I don't need to reapeat myself for double validation ?

backend/posix/posix.go
niksis02: I don't want to touch the `part` structre, as I will need to do so many changes in the codebase. It's Unmarshalled in the front-end and passed to back-end. But I will add the for loop part, looks good to me.

backend/posix/posix.go
niksis02: `s3.CompleteMultipartUploadOutput` is then marshalled to xml and sent as response body. I would like to simplify the structure, but I will need to do a bunch of changes in the front-end or add another layer which will reconstructe the structere, which is not a good idead I think ?

backend/posix/posix.go
niksis02: They are not the same type. `checksumAlgorithm` is a type of `types.ChecksumAlgorithm`, on the other hand `HashTypeNone ` is a type of `utils.Hashtype`. It makes sense to have the default value for checksumAlgorithm as empty string, because the hashreader is initialized when checksumType and checksumAlgorithm are both provided and their validation is put on front-end, otherwise we won't initialize the hashreader at all.

backend/posix/posix.go
```diff
@@ -4107,6 +4778,25 @@ func (p *Posix) ListBucketsAndOwners(ctx context.Context) (buckets []s3response.
...
+func (p *Posix) storeChecksums(f *os.File, bucket, object string, chs s3response.Checksum) error {
+	checksums, err := json.Marshal(chs)
+	if err != nil {
+		return fmt.Errorf("parse checksum: %w", err)
+	}
+
+	return p.meta.StoreAttribute(f, bucket, object, checksumsKey, checksums)
+}
+
+func (p *Posix) retreiveChecksums(f *os.File, bucket, object string) (checksums s3response.Checksum, err error) {
```
chaowang-versity: typo: retrieveChecksums

backend/posix/posix.go
```diff
@@ -1893,16 +2145,26 @@ func (p *Posix) ListParts(_ context.Context, input *s3.ListPartsInput) (s3respon
...
-			Size:         fi.Size(),
+			PartNumber:        pn,
+			ETag:              etag,
+			LastModified:      fi.ModTime(),
+			Size:              fi.Size(),
+			ChecksumCRC32:     checksum.CRC32,
+			ChecksumCRC32C:    checksum.CRC32C,
+			ChecksumSHA1:      checksum.SHA1,
+			ChecksumSHA256:    checksum.SHA256,
+			ChecksumCRC64NVME: checksum.CRC64NVME,
```
chaowang-versity, line 2167: Not related to this change, but in this loop, should we check  (may be every 64 or 128 entries)
```
select {
case <-ctx.Done():
    return s3response.ListPartsResult{}, ctx.Err()
default:
}
```
as this loop could take a while?

chaowang-versity: Also not related to this change but the `parts` could be pre-allocated as `
parts := make([]s3response.Part, 0, len(ents))`

chaowang-versity: Another issue I noticed is that the `maxParts` should default to 1000 instead of 0 to match the S3 spec


> The ListParts request returns a maximum of 1,000 uploaded parts. The limit of 1,000 parts is also the default value. You can restrict the number of parts in a response by specifying the max-parts request parameter. If your multipart upload consists of more than 1,000 parts, the response returns an IsTruncated field with the value of true, and a NextPartNumberMarker element. To list remaining uploaded parts, in subsequent ListParts requests, include the part-number-marker query string parameter and set its value to the NextPartNumberMarker field value from the previous response.

backend/posix/posix.go
```diff
@@ -1962,126 +2226,240 @@ func (p *Posix) UploadPart(ctx context.Context, input *s3.UploadPartInput) (stri
...
 		}
-		return "", fmt.Errorf("open temp file: %w", err)
+		return nil, fmt.Errorf("open temp file: %w", err)
 	}
 	defer f.cleanup()
 
 	hash := md5.New()
 	tr := io.TeeReader(r, hash)
+
+	var hashRdr *utils.HashReader
```
chaowang-versity: refactor with a loop:

```
type hashConfig struct {
    value    *string
    hashType utils.HashType
}

hashConfigs := []hashConfig{
    {input.ChecksumCRC32, utils.HashTypeCRC32},
    {input.ChecksumCRC32C, utils.HashTypeCRC32C},
    {input.ChecksumSHA1, utils.HashTypeSha1},
    {input.ChecksumSHA256, utils.HashTypeSha256},
    {input.ChecksumCRC64NVME, utils.HashTypeCRC64NVME},
}

var hashRdr *utils.HashReader
for _, config := range hashConfigs {
    if config.value != nil {
        hashRdr, err = utils.NewHashReader(tr, *config.value, config.hashType)
    // ...
}
```

backend/posix/posix.go
niksis02: I will make these changes in another PR. Thanks for noticing !

s3api/controllers/base.go
```diff
@@ -2016,16 +2078,62 @@ func (c S3ApiController) PutActions(ctx *fiber.Ctx) error {
...
+					Value: *res.ETag,
+				})
+			}
+			if res.ChecksumCRC32 != nil {
+				headers = append(headers, utils.CustomHeader{
+					Key:   "x-amz-checksum-crc32",
+					Value: *res.ChecksumCRC32,
+				})
+			}
+			if res.ChecksumCRC32C != nil {
```
chaowang-versity: Should it be `switch case` instead of `if` if only one checksum should be valid?

```
switch {

case res.ChecksumCRC32 != nil:
//...
case res.ChecksumCRC32C != nil:
//...

File: s3api/controllers/base.go [hunk 14]
```diff
@@ -2417,6 +2558,42 @@ func (c S3ApiController) PutActions(ctx *fiber.Ctx) error {
 			Value: res.VersionID,
 		})
 	}
+	if getstring(res.ChecksumCRC32) != "" {
+		hdrs = append(hdrs, utils.CustomHeader{
+			Key:   "x-amz-checksum-crc32",
+			Value: getstring(res.ChecksumCRC32),
+		})
+	}
+	if getstring(res.ChecksumCRC32C) != "" {
```
chaowang-versity: Should it be switch case instead of if if only one checksum should be valid?

s3api/controllers/base.go
```diff
@@ -3022,6 +3214,42 @@ func (c S3ApiController) HeadObject(ctx *fiber.Ctx) error {
 			Value: string(res.StorageClass),
 		})
 	}
+	if res.ChecksumCRC32 != nil {
```
chaowang-versity: Should it be `switch case` instead of `if` if only one checksum should be valid?

backend/posix/posix.go
niksis02: A separate [PR](https://github.com/versity/versitygw/pull/1078) for the changes.

s3api/controllers/base.go
niksis02: done

s3api/controllers/base.go
niksis02: done

s3api/controllers/base.go
niksis02: done

backend/posix/posix.go
niksis02: done

benmcclelland (approved):

Issue #979, opened 2024-12-10
Repetition of client certificate in error string


ndjones:
In the following error string client certificate is repeated. The conditionals leading to the error check for the certificate and the certificate key, so the error likely needs to call out both. Suggest updating the second reference to client certificate key.

https://github.com/versity/versitygw/blob/070406926251e901504172ad14358b81bb4d6895/auth/iam_vault.go#L50

```
	if clientCert != "" {
			if clientCertKey == "" {
				return nil, fmt.Errorf("client certificate and client certificate should both be specified")
			}

```

benmcclelland:
@ndjones This looks like a good fix to me. Do you want to open a PR for this to get credit for the fix?

PR #1054, opened 2025-02-02
Update iam_vault.go

ndjones:
clarify error message as noted in https://github.com/versity/versitygw/issues/979


versity-github:

benmcclelland (approved):

PR #1078, opened 2025-02-19
ListParts refactoring

niksis02:
Removes the `maxParts` and `maxUploads` nil checks, as its validation and value defaulting is put on front-end. If it's not provided it defaults to `1000`.

Adds request context cancelation/timeout check in parts loop, as it could take long to read and parse the directories.
Prealocates the `parts` slice length, based on the read directores length.

Adds integration tests for `ListObjects(V2)`, `ListParts`, `ListMultipartUploads` to check the defulat value to be 1000 for max*, if the provided value is greater than that.

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -2008,10 +2008,7 @@ func (p *Posix) ListMultipartUploads(_ context.Context, mpu *s3.ListMultipartUpl
 		}
 	}
 
-	maxUploads := 0
-	if mpu.MaxUploads != nil {
-		maxUploads = int(*mpu.MaxUploads)
-	}
+	maxUploads := int(*mpu.MaxUploads)
 	if (uploadIDMarker != "" && !uploadIdMarkerFound) || (keyMarker != "" && keyMarkerInd == -1) {
 		return s3response.ListMultipartUploadsResult{
 			Bucket:         bucket,
@@ -2064,7 +2061,7 @@ func (p *Posix) ListMultipartUploads(_ context.Context, mpu *s3.ListMultipartUpl
 	}, nil
 }
 
-func (p *Posix) ListParts(_ context.Context, input *s3.ListPartsInput) (s3response.ListPartsResult, error) {
+func (p *Posix) ListParts(ctx context.Context, input *s3.ListPartsInput) (s3response.ListPartsResult, error) {
 	var lpr s3response.ListPartsResult
 
 	if input.Bucket == nil {
@@ -2084,10 +2081,8 @@ func (p *Posix) ListParts(_ context.Context, input *s3.ListPartsInput) (s3respon
 	if input.PartNumberMarker != nil {
 		stringMarker = *input.PartNumberMarker
 	}
-	maxParts := 0
-	if input.MaxParts != nil {
-		maxParts = int(*input.MaxParts)
-	}
+
+	maxParts := int(*input.MaxParts)
 
 	var partNumberMarker int
 	if stringMarker != "" {
@@ -2127,8 +2122,15 @@ func (p *Posix) ListParts(_ context.Context, input *s3.ListPartsInput) (s3respon
 		return lpr, fmt.Errorf("get mp checksum: %w", err)
 	}
 
-	var parts []s3response.Part
-	for _, e := range ents {
+	parts := make([]s3response.Part, 0, len(ents))
+	for i, e := range ents {
+		if i%128 == 0 {
+			select {
+			case <-ctx.Done():
+				return s3response.ListPartsResult{}, ctx.Err()
+			default:
+			}
+		}
 		pn, err := strconv.Atoi(e.Name())
 		if err != nil {
 			// file is not a valid part file
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -1405,7 +1405,7 @@ func (p *Posix) CompleteMultipartUpload(ctx context.Context, input *s3.CompleteM
 
 	objdir := filepath.Join(metaTmpMultipartDir, fmt.Sprintf("%x", sum))
 
-	checksums, err := p.retreiveChecksums(nil, bucket, filepath.Join(objdir, uploadID))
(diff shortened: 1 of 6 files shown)


backend/posix/posix.go
```diff
@@ -2084,10 +2084,8 @@ func (p *Posix) ListParts(_ context.Context, input *s3.ListPartsInput) (s3respon
 	if input.PartNumberMarker != nil {
 		stringMarker = *input.PartNumberMarker
 	}
-	maxParts := 0
-	if input.MaxParts != nil {
-		maxParts = int(*input.MaxParts)
-	}
+
+	maxParts := int(*input.MaxParts)
```
chaowang-versity, line 2085: To match the S3 default value of 1000, should it check 
```
if maxParts == 0 {
maxParts = 1000
}

Comment (niksis02) [reply backend/posix/posix.go:2085]: We don't need this check, as it could be zero and it's a valid value

Comment (chaowang-versity) [reply backend/posix/posix.go:2085]: Actually just based on the S3 document ,  it should be 
```
If maxParts > 1000 || maxParts <= 0 {
    maxParts = 1000
}
```
> The ListParts request returns a maximum of 1,000 uploaded parts. The limit of 1,000 parts is also the default value. You can restrict the number of parts in a response by specifying the max-parts request parameter. If your multipart upload consists of more than 1,000 parts, the response returns an IsTruncated field with the value of true, and a NextPartNumberMarker element. To list remaining uploaded parts, in subsequent ListParts requests, include the part-number-marker query string parameter and set its value to the NextPartNumberMarker field value from the previous response.

Comment (niksis02) [reply backend/posix/posix.go:2085]: I can't find any line referring to 0 defaulting to 1000. If 0 is explicitely specified as `maxParts` it returns an empty parts list. Not sure if this is intentional or not, but I've tried it directoly on s3.

Comment (niksis02) [reply backend/posix/posix.go:2085]: For maxParts > 1000 it defaults to 1000.
For maxParts < 0 it returns `Argument max-parts must be an integer between 0 and 2147483647.`
But for 0 it returns an empty list and it's a valid value.

Comment (niksis02) [reply backend/posix/posix.go:2085]: Response example from aws cli
```json
{
    "Bucket": "niksis02-bucket-1",
    "Key": "hello",
    "UploadId": "AwtKhM1JgF8FsRBjtpu1VM_lNzznfuTfV1TC4lFgxFkfzx_2_9BvqyTqklIdw330IJQButfU8gHiVcrFfkk1eQ--",
    "PartNumberMarker": 0,
    "NextPartNumberMarker": 0,
    "MaxParts": 0,
    "IsTruncated": false,
    "Initiator": {
        "ID": "39bd01d5d2a3ddcf0b885c0f85081ce1107cf48e5124fed5ca96f5ad95023bb3",
        "DisplayName": "sisnikoyan"
    },
    "Owner": {
        "DisplayName": "sisnikoyan",
        "ID": "39bd01d5d2a3ddcf0b885c0f85081ce1107cf48e5124fed5ca96f5ad95023bb3"
    },
    "StorageClass": "STANDARD"
}
```

niksis02: I will add the check for `> 1000` to default to 1000. And it's the same for `ListMultipartUploads`.
I will also need to add 2 integration tests for this. So I will mark the PR as draft for now.

chaowang-versity: I see, sounds like 0 is a valid value. 

Since the doc mentioned :
The limit of 1,000 parts is also the default value.  Could there be a case that the `MaxPart` is missing, i.e. MaxPart == nil, so that 1000 should be returned?

https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListParts.html

niksis02: This is an example response when `?max-parts=MaxParts` query param is not provided(e.g. MaxPart == nil)
URL: https://niksis02-bucket-1.s3.us-east-1.amazonaws.com/hello?uploadId=AwtKhM1JgF8FsRBjtpu1VM_lNzznfuTfV1TC4lFgxFkfzx_2_9BvqyTqklIdw330IJQButfU8gHiVcrFfkk1eQ--

Response body:
```xml
<ListPartsResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
  <Bucket>niksis02-bucket-1</Bucket>
  <Key>hello</Key>
  <UploadId>AwtKhM1JgF8FsRBjtpu1VM_lNzznfuTfV1TC4lFgxFkfzx_2_9BvqyTqklIdw330IJQButfU8gHiVcrFfkk1eQ--</UploadId>
  <Initiator>
    <ID>39bd01d5d2a3ddcf0b885c0f85081ce1107cf48e5124fed5ca96f5ad95023bb3</ID>
    <DisplayName>sisnikoyan</DisplayName>
  </Initiator>
  <Owner>
    <ID>39bd01d5d2a3ddcf0b885c0f85081ce1107cf48e5124fed5ca96f5ad95023bb3</ID>
    <DisplayName>sisnikoyan</DisplayName>
  </Owner>
  <StorageClass>STANDARD</StorageClass>
  <PartNumberMarker>0</PartNumberMarker>
  <NextPartNumberMarker>2</NextPartNumberMarker>
  <MaxParts>1000</MaxParts>
  <IsTruncated>false</IsTruncated>
  <Part>
    <PartNumber>1</PartNumber>
    <LastModified>2025-02-19T19:34:15.000Z</LastModified>
    <ETag>"e1c30273583fd7628abce330084d8b21"</ETag>
    <Size>28210</Size>
  </Part>
  <Part>
    <PartNumber>2</PartNumber>
    <LastModified>2025-02-19T19:34:24.000Z</LastModified>
    <ETag>"fa510c2943bc9fea945ab90450ebea8a"</ETag>
    <Size>4189</Size>
  </Part>
</ListPartsResult>
```


The nil case is handled in `ParseUint` function: https://github.com/versity/versitygw/blob/main/s3api/utils/utils.go#L177-L186
https://github.com/versity/versitygw/blob/main/s3api/controllers/base.go#L265-L280


So if `MaxParts` is not provided, s3 returns 1000 by default

benmcclelland (approved):

PR #1094, opened 2025-02-27
Support BucketLinks on admin command (Fix #1081)

### LINKED ISSUE #1081: [Bug] - versitygw admin does not respect --bucketlinks
**Describe the bug**
When running versitygw with `posix --bucketlinks /some/dir` `versitygw admin list-buckets` does not list the buckets that are represented by symlinks. The buckets that are not listed with the above admin command are listed though with `aws s3api list-buckets`

**To Reproduce**
Just start versitygw with posix and create some normal directories and some symlinks under the bucket root directory (the directory provided as argument to posix). Then call commands mentioned in the "Describe the bug" section and see the difference.

**Expected behavior**
Same buckets are listed via the admin API and the standard S3 API.

**Server Version**
1.0.10

**Additional context**
None.

jbert:
- Share code between backend/posix/ListBuckets and backend/posix/ListBucketsAndOwners, so both honour bucketlinks
- add some basic tests to backend/posix to check functionality working
- pulls in testify for string-array-equality (there were already references in go.mod, but easy to remove dependency if desired)

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -249,35 +249,15 @@ func (p *Posix) doesBucketAndObjectExist(bucket, object string) error {
 }
 
 func (p *Posix) ListBuckets(_ context.Context, input s3response.ListBucketsInput) (s3response.ListAllMyBucketsResult, error) {
-	entries, err := os.ReadDir(".")
+	fis, err := listBucketFileInfos(p.bucketlinks)
 	if err != nil {
-		return s3response.ListAllMyBucketsResult{},
-			fmt.Errorf("readdir buckets: %w", err)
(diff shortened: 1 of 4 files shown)


versity-github:

benmcclelland:
Thank you for the PR for this! Let me think about these added dependencies, we may be able to move this test into the systems tests instead.

benmcclelland:
closing this PR, I cherry-picked and merged the fix without the tests in #1097. Will work with QA team to get this tested in system tests instead so that we dont have the extra dependencies and we can keep unit tests from needing a local fs.

jbert:
Awesome. Thanks for that, and apologies that I didn't understand the testing regime/separation.

PR #1096, opened 2025-02-27
Test/rest checksum crc64nvme

lrm25:


tests/rest_scripts/calculate_crc64nvme.sh
```diff
@@ -0,0 +1,50 @@
...
+fi
+if ! python3 -m pip install awscrt 1>/dev/null 2>&1; then
+  log_rest 2 "error installing awscrt"
+  exit 1
+fi
+if ! checksum_decimal=$(python3 -c "
+import sys
+from awscrt import checksums
+
+with open(sys.argv[1], 'rb') as f:
```
anodelman, line 27: This seems like something that could be built into the test vm itself, so that we don't have to keep installing it.  Perhaps as a pre-built script that is installed and can then be executed as needed?

tests/test_rest_checksum.sh
```diff
@@ -22,16 +22,70 @@ source ./tests/util/util_head_object.sh
...
 
-  run check_checksum_rest "$BUCKET_ONE_NAME" "$test_file" "$TEST_FILE_FOLDER/$test_file"
+  run check_checksum_rest_sha256 "$BUCKET_ONE_NAME" "$test_file" "$TEST_FILE_FOLDER/$test_file"
   assert_success
-}
\ No newline at end of file
+}
+
+@test "REST - PutObject rejects invalid sha256 checksum" {
+  test_file="test_file"
```
anodelman: This is set the same for all tests, should just be a default at the top of the test file.

tests/test_rest_versioning.sh
```diff
@@ -0,0 +1,106 @@
...
+
+  run put_bucket_versioning_rest "$BUCKET_ONE_NAME" "Suspended"
+  assert_success
+
+  run check_versioning_status_rest "$BUCKET_ONE_NAME" "Suspended"
+  assert_success
+}
+
+@test "test_rest_versioning" {
+  test_file="test_file"
```
anodelman: More shared settings.

tests/util/util_head_object.sh
```diff
@@ -134,10 +134,45 @@ check_checksum_rest() {
...
+    log 2 "error checking checksum"
+    return 1
+  fi
+  return 0
+}
+
+crc64_file() {
+  source ./env/bin/activate
+  python3 -m pip install awscrt
+  if ! checksum=$(python3 -c "import sys;from awscrt import checksums;crc64nvme_value = checksums.crc64nvme(sys.argv[1]);print(crc64nvme_value)" "$1"); then
```
anodelman: More of this little built in script - if we are doing repeat code then it really should be stored separately.

anodelman (changes requested):

anodelman (approved):

PR #1099, opened 2025-02-28
fix: Prioritize explicit deny in bucket policy statements

### LINKED ISSUE #1092: [Bug] - Deny statements in bucket policy not applied
**Describe the bug**
According to https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_evaluation-logic_policy-eval-denyallow.html

> Deny evaluation – By default, all requests are denied. This is called an [implicit deny](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_evaluation-logic_AccessPolicyLanguage_Interplay.html). The AWS enforcement code evaluates all policies within the account that apply to the request. These include AWS Organizations SCPs and RCPs, resource-based policies, identity-based policies, IAM permissions boundaries, and session policies. In all those policies, the enforcement code looks for a Deny statement that applies to the request. This is called an [explicit deny](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_evaluation-logic_AccessPolicyLanguage_Interplay.html). If the enforcement code finds even one explicit deny that applies, the enforcement code returns a final decision of Deny. If there is no explicit deny, the enforcement code evaluation continues.

With the following policy, using the rw-except-someprefix user I'm able to PUT objects under `/someprefix/` in the bucket, my understanding is that access should be denied:
```
{
    "Statement": [
        {
            "Action": [
                "s3:*"
            ],

            "Effect": "Allow",
            "Principal": [
                "rw-someprefix"
            ],
            "Resource": [
                "arn:aws:s3:::bucket-name/someprefix/*",
                "arn:aws:s3:::bucket-name"
            ],
            "Sid": "AllowsReadWriteSomePrefix"
        },
        {
            "Action": [
                "s3:*"
            ],
            "Effect": "Allow",
            "Principal": [
                "rw-except-someprefix"
            ],
            "Resource": [
                "arn:aws:s3:::bucket-name/*",
                "arn:aws:s3:::bucket-name"
            ],
            "Sid": "AllowsReadWriteExceptSomePrefix"
        },
        {
            "Action": [
                "s3:*"
            ],
            "Effect": "Deny",
            "Principal": [
                "rw-except-someprefix"
            ],
            "Resource": "arn:aws:s3:::bucket-name/someprefix/*",
            "Sid": "DeniesWriteSomePrefix"
        }
    ],
    "Version": "2012-10-17"
}
```

**To Reproduce**
Just create a bucket, create the two users and apply the bucket policy. After that try to PUT objects to the prefix where access is explicitly denied to the user via the Deny Effect.

**Expected behavior**
Should work the same way as AWS S3, access should be denied to the specific prefix.

**Server Version**
v1.0.10

**Additional context**
Please note that this is also a security issue as based on the AWS policy documentation one would expect that this works, but with this actually users can manipulate data on prefixes where they are not supposed to.

niksis02:
Fixes #1092 

In multi-statement bucket policies, explicit `Deny` rules should take precedence over `Allow` rules.

This change updates the implementation to iterate through all policy statements. If a `Deny` statement is found, it immediately returns `AccessDenied`. If no `Deny` statement is found but an `Allow` statement is present, access is granted.

### DIFF
--- a/auth/bucket_policy.go
+++ b/auth/bucket_policy.go
@@ -48,18 +48,19 @@ func (bp *BucketPolicy) Validate(bucket string, iam IAMService) error {
 }
 
 func (bp *BucketPolicy) isAllowed(principal string, action Action, resource string) bool {
+	var isAllowed bool
 	for _, statement := range bp.Statement {
 		if statement.findMatch(principal, action, resource) {
 			switch statement.Effect {
 			case BucketPolicyAccessTypeAllow:
-				return true
+				isAllowed = true
 			case BucketPolicyAccessTypeDeny:
 				return false
 			}
 		}
 	}
 
-	return false
+	return isAllowed
 }
 
 type BucketPolicyItem struct {
--- a/tests/integration/group-tests.go
+++ b/tests/integration/group-tests.go
@@ -471,6 +471,7 @@ func TestPutBucketPolicy(s *S3Conf) {
 	PutBucketPolicy_incorrect_bucket_name(s)
 	PutBucketPolicy_object_action_on_bucket_resource(s)
 	PutBucketPolicy_bucket_action_on_object_resource(s)
+	PutBucketPolicy_explicit_deny(s)
 	PutBucketPolicy_success(s)
 }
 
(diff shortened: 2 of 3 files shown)


auth/bucket_policy.go
```diff
@@ -47,19 +47,19 @@ func (bp *BucketPolicy) Validate(bucket string, iam IAMService) error {
 	return nil
 }
 
-func (bp *BucketPolicy) isAllowed(principal string, action Action, resource string) bool {
+func (bp *BucketPolicy) isAllowed(principal string, action Action, resource string) (isAllowed bool) {
```
benmcclelland: can we do with without the named/naked return? style-wise we have been avoiding naked returns.

benmcclelland (approved):

PR #1172, opened 2025-03-28
Test/chunked trailer invalid checksums

lrm25:


tests/rest_scripts/calculate_checksum.sh
```diff
@@ -16,29 +16,64 @@
...
+calculate_checksum_python() {
+  if [ "$#" -ne 2 ]; then
+    log 2 "'calculate_checksum_python' requires checksum type, data file"
+    return 1
+  fi
+  if ! DEACTIVATE=false source ./tests/rest_scripts/init_python_env.sh; then
+    log_rest 2 "error initializing python environment"
+    return 1
+  fi
+  if ! checksum_decimal=$(python3 -c "
```
anodelman: Having this script have its own inline subscript seems a little janky.  Maybe have the python script be its own thing that is called out to?

tests/test_user_common.sh
```diff
@@ -23,9 +23,7 @@ source ./tests/util/util_users.sh
 source ./tests/commands/list_buckets.sh
 
 test_admin_user() {
-  if [[ $# -ne 1 ]]; then
-    fail "test admin user command requires command type"
-  fi
+  assert [ $# -eq 1 ]
```
anodelman: Looks like you are removing the information associated with this failure?  Is that stored somewhere else for debugging?

anodelman (changes requested):

tests/test_user_common.sh
lrm25: Looks like I initially set this up for different clients, but am not using them yet.  So I can just remove that check.

PR #1192, opened 2025-04-06
Test/chunked trailer incorrect checksums

lrm25:


tests/test_rest_bucket.sh
```diff
@@ -0,0 +1,104 @@
...
+@test "REST - bucket tagging - no tags" {
+  run setup_bucket "s3api" "$BUCKET_ONE_NAME"
+  assert_success
+
+  run verify_no_bucket_tags_rest "$BUCKET_ONE_NAME"
+  assert_success
+}
+
+@test "REST - bucket tagging - tags" {
+  test_key="testKey"
```
anodelman, line 54: Would there be some value in testing a range of key/value pairs (with numbers, symbols, length)?

anodelman (approved):
Mostly just a comment about potential additional tests.

PR #1212, opened 2025-04-12
feat: add option to disable use of O_TMPFILE

### LINKED ISSUE #1035: [Bug] - posix "link tmpfile" multipart invalid cross-device link with zfs
**Describe the bug**
versitygw creates multipart temp files at `bucket/.sgwtmp/multipart` and tries to hardlink them into place
on zfs it's common to have nested datasets e.g. `pool/dataset` and `pool/dataset/inner`, which are separate mounts
This means versitygw will attempt a cross-device link during multipart uploads

**To Reproduce**
Create a posix bucket where the root is on a different filesystem mount than a subdir, then do a multipart upload to the subdir

**Expected behavior**
The multipart upload succeeds.

**Server Version**
$ versitygw -version && uname -a
Version  : v1.0.7
Build    : 162944d
BuildTime: 2024-10-27_07:28:49PM
Linux nas1 5.15.0-97-generic #107-Ubuntu SMP Wed Feb 7 13:26:48 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

**Additional context**
s3 client is boto3

### LINKED ISSUE #1194: [Bug] - ExistingObjectIsDirectory when PUT file to mount inside bucket (POSIX)
**Describe the bug**
Some (sub)directories within some of my buckets are mountpoints of other filesystems (actually bind-mounts to different directories on the same fs in this case, but I think that's not relevant).
When I try to PUT an object (using s3cmd or boto3) a file to a location inside such a mount, the client fails with HTTP 409/ExistingObjectIsDirectory.

The error gets returned here:
https://github.com/versity/versitygw/blob/7a3de637e4415d65e023cacb937845ca5957304c/backend/posix/posix.go#L2932

However, the actual error is (if you print `err` before that line)
`rename tmpfile: rename BUCKET/.sgwtmp/e7a82d044c9d9607447cf3ffe4c1ea5fbe86c907bb03236bef69cb984274dda4.1504859165 BUCKET/somewhere/MOUNTPOINT/somewhere/newfilename: invalid cross-device link`.

It seems that the file is initially created as a temporary file on the bucket root (bucket fs), which, if everything goes well, gets made permanent by hardlinking it to its final location. This fails because that destination is on a different filesystem.

_Could the temporary file instead be created in the same location as where the file is to be PUT?_

**To Reproduce**
1. Mount something to a subdirectory within a bucket
2. Try to PUT to a place inside that mount

**Expected behavior**
I would like this to work.
If that's not feasible, that would be sad, but at least the error message could be corrected (see below).

```go
err = f.link()
if errors.Is(err, syscall.EEXIST) {return ...}
if err != nil {  // statement too general
	return s3response.PutObjectOutput{}, s3err.GetAPIError(s3err.ErrExistingObjectIsDirectory)
}
```

**Server Version**
v1.0.12
Linux 7dd145af44fa 6.12.21-1-lts #1 SMP PREEMPT_DYNAMIC Sat, 29 Mar 2025 06:39:46 +0000 x86_64 x86_64 x86_64 GNU/Linux

benmcclelland:
O_TMPFILE can fail if the location we need to link the final file is not within the same filesystem. This can happen if there are different filesystem mounts within a bucket or if using zfs nested datasets within a bucket.

Fixes #1194
Fixes #1035

### DIFF
--- a/backend/common.go
+++ b/backend/common.go
@@ -17,11 +17,14 @@ package backend
 import (
 	"crypto/md5"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/service/s3/types"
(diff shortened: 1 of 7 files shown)


niksis02 (approved):

benmcclelland:
testing for this:
```
 losetup /dev/loop10 /tmp/loop1.img
 losetup /dev/loop11 /tmp/loop2.img
 mkfs.xfs /dev/loop10
 mkfs.xfs /dev/loop11
 mkdir -p /mnt/xfs1
 mount /dev/loop10 /mnt/xfs1
 mkdir -p /mnt/xfs1/bucket/data
 mount /dev/loop11 /mnt/xfs1/bucket/data

./versitygw -a test -s test posix /mnt/xfs1
./versitygw -a test -s test posix --disableotmp /mnt/xfs1
```
without `--disableotmp`:
```
# aws s3 cp ./testfile s3://bucket/data/testfile
upload failed: ./testfile to s3://bucket/data/file An error occurred (ExistingObjectIsDirectory) when calling the PutObject operation: Existing Object is a directory.
```

PR #1247, opened 2025-04-29
Add support for supplying s3.Client instance to S3 Proxy

ttschampel:
Provide an alternate S3Proxy initialization using a provided s3.Client instance to support more advanced configuration.

### DIFF
--- a/backend/s3proxy/s3.go
+++ b/backend/s3proxy/s3.go
@@ -63,6 +63,18 @@ type S3Proxy struct {
 
 var _ backend.Backend = &S3Proxy{}
 
+func NewWithClient(ctx context.Context, client *s3.Client, metaBucket string) (*S3Proxy, error) {
+	s := &S3Proxy{
+		metaBucket: metaBucket,
+	}
+	s.client = client
+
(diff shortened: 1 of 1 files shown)


versity-github:

benmcclelland:
@ttschampel This change looks fine, but currently nothing would call NewWithClient(). Do you have a use case for how to make use of this?

ttschampel:
> @ttschampel This change looks fine, but currently nothing would call NewWithClient(). Do you have a use case for how to make use of this?

We are have a custom backend implementation that delegates some operations to the s3proxy, so we would call this function from this custom backend.    

In our case the  AWS S3 Client Configuration options / customization we need are more complex than what is currently supported via the current s3proxy New function

benmcclelland (approved):

benmcclelland:
> > @ttschampel This change looks fine, but currently nothing would call NewWithClient(). Do you have a use case for how to make use of this?
> 
> We are have a custom backend implementation that delegates some operations to the s3proxy, so we would call this function from this custom backend.
> 
> In our case the AWS S3 Client Configuration options / customization we need are more complex than what is currently supported via the current s3proxy New function

Ah! sounds interesting. We could add more options to the s3 backend setup if needed too.

Issue #1134, opened 2025-03-11
[Feature] - path style option for s3 iam


benmcclelland:
**Describe the solution you'd like**
Add option for path style bucket access in the s3 iam options.

PR #1268, opened 2025-05-06
use path style

sebastian-heinz:
I want to proxy to a s3 service that does not support the domain bucket names, therefore exposing the `usePathStyle` options as parameter / env variable


versity-github:

benmcclelland:
Looks good, thanks for the contribution!
Closes #1134.

benmcclelland (approved):

PR #1281, opened 2025-05-09
Test/remove setup bucket param

lrm25:


tests/util/util_bucket.sh
```diff
@@ -247,12 +247,12 @@ bucket_cleanup_if_bucket_exists() {
 # params:  client, bucket name(s)
 # return 0 for success, 1 for failure
 setup_buckets() {
-  if [ $# -lt 2 ]; then
-    log 2 "'setup_buckets' command requires client, bucket names"
+  if [ $# -lt 1 ]; then
```
anodelman: Could use check_param_count here?

tests/util/util_bucket.sh
```diff
@@ -264,28 +264,28 @@ setup_buckets() {
 # return 0 on successful setup, 1 on error
 setup_bucket() {
   log 6 "setup_bucket"
-  if [ $# -ne 2 ]; then
-    log 2 "'setup_bucket' requires client, bucket name"
+  if [ $# -ne 1 ]; then
+    log 2 "'setup_bucket' requires bucket name"
```
anodelman: Use check_param_count?

tests/util/util_bucket.sh
lrm25: check_param_count was intended by me for "run/assert" combinations on common test procedures but with different clients (e.g. in test_common.sh), so that I can return a proper error messages for param count errors there.  I added more of these checks there though.

anodelman: Yeah, the checks look good.  Just figured that it could also be used within the test functions, since it doesn't have run/assert in it it wouldn't disrupt error returns but it would provide better error messaging.

lrm25: Looked at it again, yeah, it could eliminate a line

anodelman (approved):

PR #1286, opened 2025-05-12
add omit on empty for ContinuationToken

ndjones:
mount-s3 has strict XML parsing which doesn't like receiving this tag empty. An empty ContinuationToken causes mount-s3 to error out and stop processing requests.


versity-github:

PR #1297, opened 2025-05-15
Test/log change rest setup converions

lrm25:


tests/drivers/drivers.sh
```diff
@@ -20,7 +20,19 @@ check_param_count() {
...
-    log 2 "function $1 requires $2"
+    log_with_stack_ref 2 "function $1 requires $2" 2
+    return 1
+  fi
+  return 0
+}
+
+assert_param_count() {
+  if [ $# -ne 4 ]; then
+    log 2 "'check_param_count' requires function name, params list, expected, actual"
```
anodelman: Should be 'assert_param_count'

anodelman (changes requested):

PR #1298, opened 2025-05-16
Test/post file delete setup conversions

lrm25:


tests/commands/command.sh
```diff
@@ -27,8 +27,11 @@ send_command() {
...
     # shellcheck disable=SC2154
     echo "${masked_args[*]}" >> "$COMMAND_LOG"
-    "$@"
-    return $?
   fi
-  "$@"
+  local command_result=0
+  "$@" || command_result=$?
+  if [ "$command_result" -ne 0 ] && [ "$1" == "curl" ]; then
+    echo ", curl response code: $command_result"
```
anodelman: Might be worth doing this for all commands, not just curl, for better error tracking?

anodelman (changes requested):

tests/commands/command.sh
lrm25: Some of the other tools can possibly return codes besides 0 or 1, so that could be a good idea

PR #1314, opened 2025-05-23
Test/large direct get object bug

lrm25:


tests/util/util_list_parts.sh
```diff
@@ -118,11 +118,18 @@ upload_check_parts() {
     log 2 "error checking part list before part upload"
     return 1
   fi
+  sleep 5
```
anodelman: Can you please put a comment in for this so we know what you are waiting for?  Makes it easier to diagnose weird timing failures later.

lrm25: Removed it, probably an old delay I was using for debugging

PR #1319, opened 2025-05-28
feat: implements public bucket access.

### LINKED ISSUE #1325: [Bug] - s3:GetBucketObjectLockConfiguration is considered as an invalid bucket policy action
**Describe the bug**
The gateway returns an error when putting a bucket policy which contains the `s3:GetBucketObjectLockConfiguration` action.

Error retruned
```
An error occurred (MalformedPolicy) when calling the PutBucketPolicy operation: Policy has invalid action
```

**To Reproduce**
```bash
// create a test user
./versitygw admin create-user --access grt1 --secret grt1secret --role user

// create a bucket

// put the bucket policy
```

where policy.json is:
```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "Statement",
      "Effect": "Allow",
      "Principal": {
        "AWS": "grt1"
      },
      "Action": "s3:GetBucketObjectLockConfiguration",
      "Resource": "arn:aws:s3:::test"
    }
  ]
}
```

**Expected behavior**
The policy should successfully be updated without any errors

niksis02:
Fixes #1325 

This implementation introduces **public buckets**, which are accessible without signature-based authentication.

Public buckets support a set of actions on buckets and objects, returning various errors based on the S3 action type and permissions (ACL or policy). The implementation aligns with the table provided in [this gist](https://gist.github.com/niksis02/5919d52d6112537a31c14d9abfa89ac0).

There are two ways to grant public access to a bucket:

* **Bucket ACLs**
* **Bucket Policies**

The implementation includes an `AuthorizePublicBucketAccess` middleware, which checks if public access has been granted to the bucket. If so, authentication middlewares are skipped. For unauthenticated requests, appropriate errors are returned based on the specific S3 action.

---

**1. Bucket-Level Operations:**

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": "*",
      "Action": "s3:*",
      "Resource": "arn:aws:s3:::test"
    }
  ]
}
```

**2. Object-Level Operations:**

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": "*",
      "Action": "s3:*",
      "Resource": "arn:aws:s3:::test/*"
    }
  ]
}
```

**3. Both Bucket and Object-Level Operations:**

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": "*",
      "Action": "s3:*",
      "Resource": "arn:aws:s3:::test"
    },
    {
      "Effect": "Allow",
      "Principal": "*",
      "Action": "s3:*",
      "Resource": "arn:aws:s3:::test/*"
    }
  ]
}
```

---

```sh
aws s3api create-bucket --bucket test --object-ownership BucketOwnerPreferred
aws s3api put-bucket-acl --bucket test --acl public-read
```

### DIFF
--- a/s3api/controllers/base.go
+++ b/s3api/controllers/base.go
@@ -77,7 +77,7 @@ func (c S3ApiController) ListBuckets(ctx *fiber.Ctx) error {
 	cToken := ctx.Query("continuation-token")
 	prefix := ctx.Query("prefix")
 	maxBucketsStr := ctx.Query("max-buckets")
-	acct := ctx.Locals("account").(auth.Account)
+	acct := utils.ContextKeyAccount.Get(ctx).(auth.Account)
 
 	var maxBuckets int32 = 10000
 	if maxBucketsStr != "" {
@@ -119,9 +119,10 @@ func (c S3ApiController) GetActions(ctx *fiber.Ctx) error {
 	uploadId := ctx.Query("uploadId")
 	partNumberMarker := ctx.Query("part-number-marker")
 	acceptRange := ctx.Get("Range")
-	acct := ctx.Locals("account").(auth.Account)
-	isRoot := ctx.Locals("isRoot").(bool)
-	parsedAcl := ctx.Locals("parsedAcl").(auth.ACL)
+	acct := utils.ContextKeyAccount.Get(ctx).(auth.Account)
+	isRoot := utils.ContextKeyIsRoot.Get(ctx).(bool)
+	parsedAcl := utils.ContextKeyParsedAcl.Get(ctx).(auth.ACL)
+	isPublicBucket := utils.ContextKeyPublicBucket.IsSet(ctx)
 	versionId := ctx.Query("versionId")
 	if keyEnd != "" {
 		key = strings.Join([]string{key, keyEnd}, "/")
@@ -133,14 +134,15 @@ func (c S3ApiController) GetActions(ctx *fiber.Ctx) error {
 
 	if ctx.Request().URI().QueryArgs().Has("tagging") {
 		err := auth.VerifyAccess(ctx.Context(), c.be, auth.AccessOptions{
-			Readonly:      c.readonly,
-			Acl:           parsedAcl,
-			AclPermission: auth.PermissionRead,
-			IsRoot:        isRoot,
-			Acc:           acct,
-			Bucket:        bucket,
-			Object:        key,
-			Action:        auth.GetObjectTaggingAction,
+			Readonly:       c.readonly,
+			Acl:            parsedAcl,
+			AclPermission:  auth.PermissionRead,
+			IsRoot:         isRoot,
+			Acc:            acct,
+			Bucket:         bucket,
+			Object:         key,
+			Action:         auth.GetObjectTaggingAction,
+			IsBucketPublic: isPublicBucket,
 		})
 		if err != nil {
 			return SendXMLResponse(ctx, nil, err,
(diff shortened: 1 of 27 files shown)


s3api/middlewares/set-defualt-keys.go
maxlerebourg, line 1: typo in file name

s3log/file.go
```diff
@@ -74,7 +75,7 @@ func (f *FileLogger) Log(ctx *fiber.Ctx, err error, body []byte, meta LogMeta) {
 	}
 	errorCode := ""
 	httpStatus := 200
-	startTime := ctx.Locals("startTime").(time.Time)
+	startTime := utils.ContextKeyStartTime.Get(ctx).(time.Time)
```
benmcclelland: this has a conflict with main, needs a rebase. we need to check the type assertion in these cases where the starttime may not be set. there are apparently cases where this is a nil interface.

s3api/utils/context-keys.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (ck ContextKey) IsSet(ctx *fiber.Ctx) bool {
+	val := ctx.Locals(string(ck))
+	return val != nil
+}
+
+func (ck ContextKey) Get(ctx *fiber.Ctx) any {
+	return ctx.Locals(string(ck))
+}
```
benmcclelland, line 65: This is nice that we have one place to enforce the key names, but I this this would be a good place with validate the type assertion. I think in Get we should check for nil from ctx.Locals(), and either return the zero value for the correct type expected based on the key or an error that the caller of Get can handle.

And does public access imply things like region and account may not be set now? so maybe even more reason to validate these in all cases?

s3log/webhook.go
```diff
@@ -71,7 +72,7 @@ func (wl *WebhookLogger) Log(ctx *fiber.Ctx, err error, body []byte, meta LogMet
 	}
 	errorCode := ""
 	httpStatus := 200
-	startTime := ctx.Locals("startTime").(time.Time)
+	startTime := utils.ContextKeyStartTime.Get(ctx).(time.Time)
```
benmcclelland: type assertion check probably needed here too

s3api/controllers/base.go
```diff
@@ -3099,7 +3140,9 @@ func (c S3ApiController) DeleteActions(ctx *fiber.Ctx) error {
 	// The AWS CLI sends 'True', while Go SDK sends 'true'
 	bypass := strings.EqualFold(bypassHdr, "true")
 
-	err = auth.CheckObjectAccess(ctx.Context(), bucket, acct.Access, []types.ObjectIdentifier{{Key: &key, VersionId: &versionId}}, bypass, c.be)
+	fmt.Println("checking object access: ", bypass, IsBucketPublic)
```
benmcclelland: debugging logging removal missed? fine to leave in if useful but move to debug only.

s3api/controllers/base.go
```diff
@@ -3881,6 +3932,7 @@ func SendResponse(ctx *fiber.Ctx, err error, l *MetaOpts) error {
 			ctx.Status(apierr.HTTPStatusCode)
 			return ctx.Send(s3err.GetAPIErrorResponse(apierr, "", "", ""))
 		}
+		fmt.Println(err, "------------")
```
benmcclelland: debugging logging removal missed? fine to leave in if useful but move to debug only.

benmcclelland (changes requested):

s3api/controllers/base.go
niksis02: Pure developement debug log, will remove it :)

s3api/controllers/base.go
niksis02: Will remvoe this, thanks !

s3log/webhook.go
niksis02: Done !

s3log/file.go
niksis02: The conflict has been resolved !

PR #1348, opened 2025-06-18
feat: calculate full object crc for multi-part uploads for compatible checksums

benmcclelland:
The CRC32, CRC32c, and CRC64NVME data integrity checksums support calculating the composite full object values for multi-part uploads using the checksum and length of the individual parts.

Previously, we were reading all of the part data to recalculate the full object checksum values during the complete multipart upload call. This disabled the optimized copy_file_range() for certain filesystems such as XFS because the part data was being read. If the data is not read, and the file handle is passed directly to io.Copy(), then the filesystem is allowed to optimize the copying of the data from the source to destination files.

This now allows both the optimized copy_file_range() optimizations as well as the data integrity features enabled for support composite checksum types.

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -1352,6 +1352,8 @@ func getPartChecksum(algo types.ChecksumAlgorithm, part types.CompletedPart) str
 		return backend.GetStringFromPtr(part.ChecksumSHA1)
 	case types.ChecksumAlgorithmSha256:
 		return backend.GetStringFromPtr(part.ChecksumSHA256)
+	case types.ChecksumAlgorithmCrc64nvme:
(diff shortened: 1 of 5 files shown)


backend/posix/posix.go
```diff
@@ -1678,6 +1716,7 @@ func (p *Posix) CompleteMultipartUpload(ctx context.Context, input *s3.CompleteM
 		if err != nil {
 			return res, "", fmt.Errorf("store object checksum: %w", err)
 		}
+		fmt.Println("object crc:", sum)
```
niksis02: development debug log ? :)

benmcclelland: Yes thanks for the reminder, will remove all the printlns before final version

niksis02 (approved):

PR #1377, opened 2025-07-08
fix: add retry for iam freeipa http requests

benmcclelland:
The IPA service connections have been seen to not always work correctly on the first network connection attempt. Add retry logic for errors that appear to be transient network issues.


auth/iam_ipa.go
```diff
@@ -237,17 +240,35 @@ func (ipa *IpaIAMService) login() error {
...
+	var resp *http.Response
+	for i := range requestRetries {
+		resp, err = ipa.client.Do(req)
+		if err == nil {
+			break
+		}
+		// Check for transient network errors
+		if err == io.EOF || strings.Contains(err.Error(), "EOF") ||
+			strings.Contains(err.Error(), "connection reset") ||
+			strings.Contains(err.Error(), "timeout") {
```
yhal-nesi: timeout messages are upper case unless I am looking at the wrong place? e.g.

				err = &timeoutError{err.Error() + " (Client.Timeout exceeded while awaiting headers)"}
				
in https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/net/http/client.go

niksis02 (approved):

PR #1382, opened 2025-07-09
fix: admin bucket actions for s3proxy

### LINKED ISSUE #1381: [Bug] - change-bucket-owner while using GW as s3proxy to an S3 backend
**Describe the bug**
Since bucket listing is controlled by bucket ownership only (#1326) I tried the following:
`versitygw admin --access "[REDACTED]" --secret "[REDACTED]" --region eu-central-1 --endpoint-url https://[REDACTED] change-bucket-owner --bucket data-amt41-lesesaal-prod --owner [REDACTED] `
Gateway log:

```
Jun 05 10:22:16 x17-s3proxy1 versitygw-s3proxy-central[66686]: [DEBUG]: Internal Error, <?xml version="1.0" encoding="UTF-8"?>
Jun 05 10:22:16 x17-s3proxy1 versitygw-s3proxy-central[66686]: <Error><Code>MethodNotAllowed</Code><Message>The specified method is not allowed against this resource.</Message><Resource>/change-bucket-owner/?bucket=data-amt41-lesesaal-prod&amp;owner=[REDACTED]</Resource><RequestId>1749110771323218</RequestId></Error>
Jun 05 10:22:16 x17-s3proxy1 versitygw-s3proxy-central[66686]: 10:22:16 | 500 |   51.830602ms | 10.168.88.44 | PATCH | /change-bucket-owner/ | - | bucket=data-amt41-lesesaal-prod&owner=[REDACTED]
```

**To Reproduce**

- configure gw to proxy to a backend - using the VGW_S3... Parameters including metadata
- enable IAM in gateway using VGW_IAM_DIR
- create account in gw
- create a bucket in the S3 backends and grant the VGW_S3_ACCESS_KEY the action s3:ListBucket to this bucket
- Try to change bucket ownership via the gateway to the user stored in the gateways IAM with the command shown above

**Expected behavior**
GW stores ownership in the metadata bucket (or somewhere else within its own jurisdiction) - by doing this  encapsulating user and access management - since the s3backend has no knowledge of the gateways IAM users.

**Server Version**
Version  : 1.0.15
Build    : 3ec6e634c35d7e1404dd4729a1a1e6f00b6badf0
BuildTime: 2025-07-04T16:52:01Z
Linux x17-s3proxy1 6.1.0-37-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.140-1 (2025-05-22) x86_64 GNU/Linux

**Additional context**
If the goal of the GW really is (as I understand it and is our goal in using it as well) to provide access to buckets to its own multiple IAM users through one backend user (encapsulating user and access management) this behaviour would be detrimental to that goal. Storing ownership within the metadata bucket would be the way to go, since the backend does not know of the GWs IAM users.

Feel free to tell me if my understanding is wrong.

Thank you very much for looking into this!

benmcclelland:
We were incorrctly trying to pass through the admin request actions through to the backend s3 service in s3proxy. This was resulting in internal server errors since not all s3 backends would understand these requests. Instead the gateway needs to handle these requests directly.

Fixes #1381

### DIFF
--- a/auth/acl.go
+++ b/auth/acl.go
@@ -466,3 +466,30 @@ func VerifyPublicBucketACL(ctx context.Context, be backend.Backend, bucket strin
 
 	return nil
 }
+
+// UpdateBucketACLOwner sets default ACL with new owner and removes
+// any previous bucket policy that was in place
+func UpdateBucketACLOwner(ctx context.Context, be backend.Backend, bucket, newOwner string) error {
+	acl := ACL{
+		Owner: newOwner,
+		Grantees: []Grantee{
+			{
+				Permission: PermissionFullControl,
+				Access:     newOwner,
+				Type:       types.TypeCanonicalUser,
+			},
+		},
+	}
+
+	result, err := json.Marshal(acl)
+	if err != nil {
+		return fmt.Errorf("marshal ACL: %w", err)
+	}
+
+	err = be.PutBucketAcl(ctx, bucket, result)
+	if err != nil {
+		return err
+	}
+
+	return be.DeleteBucketPolicy(ctx, bucket)
+}
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -4888,8 +4888,8 @@ func (p *Posix) GetObjectRetention(_ context.Context, bucket, object, versionId 
 	return data, nil
 }
 
-func (p *Posix) ChangeBucketOwner(ctx context.Context, bucket string, acl []byte) error {
-	return p.PutBucketAcl(ctx, bucket, acl)
+func (p *Posix) ChangeBucketOwner(ctx context.Context, bucket, owner string) error {
+	return auth.UpdateBucketACLOwner(ctx, p, bucket, owner)
 }
 
 func listBucketFileInfos(bucketlinks bool) ([]fs.FileInfo, error) {
--- a/backend/s3proxy/s3.go
+++ b/backend/s3proxy/s3.go
@@ -17,17 +17,12 @@ package s3proxy
 import (
 	"bytes"
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
-	"net/http"
 	"strconv"
 	"time"
 
-	"github.com/aws/aws-sdk-go-v2/aws"
 	v4 "github.com/aws/aws-sdk-go-v2/aws/signer/v4"
 	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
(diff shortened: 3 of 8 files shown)


backend/s3proxy/s3.go
```diff
@@ -1553,81 +1548,38 @@ func (s *S3Proxy) GetObjectLegalHold(ctx context.Context, bucket, object, versio
...
-		body, err := io.ReadAll(resp.Body)
-		if err != nil {
-			return err
-		}
-		defer resp.Body.Close()
-		return fmt.Errorf("%v", string(body))
-	}
-
-	return nil
+	return s.PutBucketAcl(ctx, bucket, acl)
```
niksis02: I guess this will override the existing bucket ACL. Bucket owner update indicates only the `Owner.ID` change. 
This is what is passed from the admin controller to the backend:
```go
	acl := auth.ACL{
		Owner: owner,
		Grantees: []auth.Grantee{
			{
				Permission: auth.PermissionFullControl,
				Access:     owner,
				Type:       types.TypeCanonicalUser,
			},
		},
	}
```

So I think the bucket owner back-end implementation should be like:
1. Get the bucket ACL
2. Map the changes (not override)
3. Put the bucket ACL.

backend/posix/posix.go
```diff
@@ -4889,7 +4889,7 @@ func (p *Posix) GetObjectRetention(_ context.Context, bucket, object, versionId
 }
 
 func (p *Posix) ChangeBucketOwner(ctx context.Context, bucket string, acl []byte) error {
-	return p.PutBucketAcl(ctx, bucket, acl)
+	return auth.UpdateBucketACLOwner(ctx, p, bucket, acl)
```
niksis02: I would suggest changing the `ChangeBucketOwner` interface to accept only the new owner access key id. Because if a backend doesn't support ACLs it looks a bit confusing.
e.g.
ChangeBucketOwner(ctx context.Context, bucket string, ownwerAccess string) error

auth/acl.go
```diff
@@ -466,3 +466,48 @@ func VerifyPublicBucketACL(ctx context.Context, be backend.Backend, bucket strin
...
+		grantees = parsedAcl.Grantees
+	}
+	grantees = append(grantees, inputACL.Grantees...)
+	parsedAcl.Grantees = grantees
+
+	result, err := json.Marshal(parsedAcl)
+	if err != nil {
+		return fmt.Errorf("marshal updated ACL: %w", err)
+	}
+
```
niksis02: Based on the new `ChangeBucketOwner` interface I would suggest implementing it this way:

```go
func UpdateBucketACLOwner(ctx context.Context, be backend.Backend, bucket string, newOwner string) error {
	// get current acl/ownership
	data, err := be.GetBucketAcl(ctx, &s3.GetBucketAclInput{Bucket: &bucket})
	if err != nil {
		return fmt.Errorf("get bucket ACL: %w", err)
	}
	parsedAcl, err := ParseACL(data)
	if err != nil {
		return fmt.Errorf("parse bucket ACL: %w", err)
	}

	// assign the default ACL
	updatedAcl := ACL{
		Owner: newOwner,
		Grantees: []Grantee{
			{
				Permission: PermissionFullControl,
				Access:     newOwner,
				Type:       types.TypeCanonicalUser,
			},
		},
	}

	originalOwner := parsedAcl.Owner

	// update grantees by removing the old owner if exsiting in the list
	// and adding new owner info from input
	if originalOwner != "" {
		grantees := make([]Grantee, 0, len(parsedAcl.Grantees))
		for _, grant := range parsedAcl.Grantees {
			if grant.Access == originalOwner {
				grant.Access = newOwner
			}
			grantees = append(grantees, grant)
		}
		updatedAcl.Grantees = grantees
	}

	result, err := json.Marshal(updatedAcl)
	if err != nil {
		return fmt.Errorf("marshal updated ACL: %w", err)
	}
	return be.PutBucketAcl(ctx, bucket, result)
}
```


This implementation keeps all the ACL setting by only changing the owner. As far as I know there might be some settings in the grantees for the bucket owner as well. So logically `ChangeBucketOwner` only changes the bucket owner by preserving the ACL settings for the bucket. If the bucket ACL doesn't have owner, it proceeds with the default ACL.

niksis02 (approved):

PR #1391, opened 2025-07-11
fix: always log internal server error messages to stderr

benmcclelland:
The debuglogger logs will only get printed if debug is enabled, but we always want the internal server error logs to be logged by the service since this is usually some actionable error that needs to be addressed with the backend storage system.

This changes internal server error logs to always to sent to stderr.

### DIFF
--- a/s3api/controllers/base.go
+++ b/s3api/controllers/base.go
@@ -23,6 +23,7 @@ import (
 	"math"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


niksis02 (commented):
Should internal errors be logged if the gateway is run in `quiet` mode ?

benmcclelland:
> Should internal errors be logged if the gateway is run in `quiet` mode ?

yes, I think so. The intent of internal server error is that this is an unexpected error that needs the admin to look into it to possibly fix whatever is happening. So I think these always need to be logged to give context of what to fix.

niksis02 (approved):

PR #1393, opened 2025-07-12
feat: split the vault mount path into kv and auth

mfhunruh:
Hello! Thank you for your project!

I started using versitygw + iam vault. And I noticed that ```--iam-vault-mount-path``` is used for both auth and kv. So if my authentication method in Vault is called ```myapprole``` and I specify it in ```--iam-vault-mount-path```, then the secret path will be looked up at ```myapprole/path/to/secret```. Or vice versa, if I specify ```--iam-vault-mount-path mykv```, then the role authentication path will be ```auth/mykv/login```.

So I split ```--iam-vault-mount-path``` into two arguments ```--iam-vault-auth-mount-path``` and ```--iam-vault-kv-mount-path```. So that fixed my problem.


versity-github:

cmd/versitygw/main.go
```diff
@@ -381,10 +381,16 @@ func initFlags() []cli.Flag {
 			Destination: &vaultSecretStoragePath,
 		},
 		&cli.StringFlag{
-			Name:        "iam-vault-mount-path",
```
benmcclelland, line 384: I'm not sure if anyone is currently using iam-vault-mount-path arg or VGW_IAM_VAULT_MOUNT_PATH env var, but if so this change would break that config on update. maybe a better option would be to keep VGW_IAM_VAULT_MOUNT_PATH option in place, and add the optional VGW_IAM_VAULT_AUTH_MOUNT_PATH option. So if VGW_IAM_VAULT_AUTH_MOUNT_PATH is specified then it would split these for the new behavior, but if VGW_IAM_VAULT_AUTH_MOUNT_PATH is not specified then use VGW_IAM_VAULT_MOUNT_PATH for the auth path to maintain current behavior.

mfhunruh: That's true. I’ve returned the argument and variable names back

mfhunruh: @benmcclelland But this is still breaking changes. Because the auth method is changed to the default “approle” if it isn’t defined

mfhunruh: If we are worried about it, maybe we should make three arguments ```iam-vault-mount-path```, ```iam-vault-auth-mount-path``` and ```iam-vault-kv-mount-path```?  So if ```iam-vault-auth-mount-path``` and/or ```iam-vault-kv-mount-path``` is not defined and ```iam-vault-mount-path``` is defined, then  ```vaultAuthMountPath``` and/or ```vaultKvMountPath``` will use a value from ```iam-vault-mount-path```. And mark the old argument as deprecated.

benmcclelland: ah yes, you are right, I think approle should be the default auth path here. I think this is a good change. No need to mark the previous as deprecated.

mfhunruh:
@benmcclelland Let’s revert my last commit or rename iam-vault-auth-mount-path to iam-vault-auth-method? Because these args may seem similar at first glance. This will prevent confusion.

benmcclelland:
> @benmcclelland Let’s revert my last commit or rename iam-vault-auth-mount-path to iam-vault-auth-method? Because these args may seem similar at first glance. This will prevent confusion.

`iam-vault-auth-method` sounds good to me. Can you squash these changes into a single commit and force push to the github branch? This will update the PR with a single commit.

PR #1397, opened 2025-07-16
fix: refresh expired iam vault tokens when needed

### LINKED ISSUE #976: [Bug] - IAM approle id permission denied
**Describe the bug**
When using vault approle , always get permission denied.

: setup iam: approle authentication failure: 403 Forbidden: permission denied

**To Reproduce**
Create a approle in vault, with policy
Test the policy : 
````
vault write auth/versitygw-approle/login role_id="c2cefc18-de3c-c90d-2652-2e154d1ccc24" secret_id="5298b069-972b-2e7e-61b2-c4b7d8d87d28" 
````

I got the response.
````
Key                     Value
---                     -----
token                   s.m8thAk3jHiGSVK9SR9QJYyfZ
token_accessor          jscVS4AqxJT4RUmygxNT8p9P
token_duration          1h
token_renewable         true
token_policies          ["default" "versitygw-policy"]
identity_policies       []
policies                ["default" "versitygw-policy"]
token_meta_role_name    versitygw-role
````

vault login --method=token
Token (will be hidden):
Success! You are now authenticated. The token information displayed below
is already stored in the token helper. You do NOT need to run "vault login"
again. Future Vault requests will automatically use this token.
````
Key                     Value
---                     -----
token                   s.bjKAmT3r9cjbwGcna9MMxtB1
token_accessor          ppukCAjeAK4aXUzmaoM5pcY6
token_duration          59m29s
token_renewable         true
token_policies          ["default" "versitygw-policy"]
identity_policies       []
policies                ["default" "versitygw-policy"]
token_meta_role_name    versitygw-role
````
````
vault read vpe/data/prod/instances/object/test          
Key         Value
---         -----
data        map[test:map[access:test groupID:0 role:userplus secret:Z2lD6bT2FPcyZd01aM7jQKoI6tLwnQfPU3LG4cbJ userID:0]]
metadata    map[created_time:2024-12-06T10:41:20.950730138Z custom_metadata:<nil> deletion_time: destroyed:false version:1]
`````

When using the token as above as root token in the configuration the server starts and I can create and list accounts.
In vault

**Expected behavior**
When using the approle id and secret that this automatically works.

**Server Version**

versitygw -v
Version  : 1.0.8
Build    : 45ad3843ef6ee8a325524852e36ed48e45a48721
BuildTime: 2024-10-28T15:42:50Z

**Additional context**

### PR DESCRIPTION
**Pull request by benmcclelland**
The IAM vault client stores an access token once authenticated, but this token will expire after a certain amount of time set by the server generating the token. Once this token is expired or revoked, it can no longer be use by the vault client. So the client should try to refresh the token with any errors indicating expired or revoked tokens.

Fixes #976

### CODE REVIEW

File: auth/iam_vault.go [hunk 1]
```diff
@@ -19,28 +19,33 @@ import (
...
 
-func NewVaultIAMService(rootAcc Account, endpoint, secretStoragePath, authMethod, mountPath, rootToken, roleID, roleSecret, serverCert, clientCert, clientCertKey string) (IAMService, error) {
+func NewVaultIAMService(rootAcc Account, endpoint, secretStoragePath,
+	authMethod, mountPath, rootToken, roleID, roleSecret, serverCert,
+	clientCert, clientCertKey string) (IAMService, error) {
 	opts := []vault.ClientOption{
 		vault.WithAddress(endpoint),
-		// set request timeout to 10 secs
-		vault.WithRequestTimeout(10 * time.Second),
+		vault.WithRequestTimeout(requestTimeout),
```
niksis02, line 48: If request timeout is configured for the whole client, don't we need to control it per method ?

benmcclelland: The documentation:

> WithRequestTimeout, given a non-negative value, will apply the timeout to each request function unless an earlier deadline is passed to the request function through context.Context. Note that this timeout is not applicable to client.ReadRaw(...) or client.ReadRawWithParameters(...).
> 
> Default: 60s

makes it sound like this will set the default timeout for each request

niksis02: I see.

niksis02 (approved):

PR #1430, opened 2025-08-04
Fix race in GetObject

ondrap:
The `GetObject` call may return wrong metadata (even checksum and content-length) when an upload to an existing object is running concurrently. 

Testing: 
- Run concurrent upload (in infinite loop) of 3 files of different size into the same Key, include some checksum with the upload
- Try to read the Key and check the checksum. Occasionally a checksum doesn't fit the data (probably because a wrong content-length is used for the data)
- This test may also fail with 404, which is a separate problem, so if you are running this on Linux, it's necessary to disable build of `./backend/posix/with_otmpfile.go` and use `./backend/posix/without_tmpfile.go`.

This PR seems to solve the problem. I'm not an expert in Go and don't know the codebase, so take it more as an inspiration, but it seems to me it's a reasonable fix. Along with the `OTMPFILE` problem this was found while running concurrent elasticsearch S3 analyze.

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -1563,7 +1563,7 @@ func (p *Posix) CompleteMultipartUpload(ctx context.Context, input *s3.CompleteM
 	upiddir := filepath.Join(objdir, uploadID)
 
 	userMetaData := make(map[string]string)
-	objMeta := p.loadObjectMetaData(bucket, upiddir, nil, userMetaData)
+	objMeta := p.loadObjectMetaData(nil, bucket, upiddir, nil, userMetaData)
 	err = p.storeObjectMetadata(f.File(), bucket, object, objMeta)
 	if err != nil {
 		return res, "", err
@@ -1859,7 +1859,7 @@ type objectMetadata struct {
 
 // fill out the user metadata map with the metadata for the object
 // and return object meta properties as `ObjectMetadata`
-func (p *Posix) loadObjectMetaData(bucket, object string, fi *os.FileInfo, m map[string]string) objectMetadata {
+func (p *Posix) loadObjectMetaData(f *os.File, bucket, object string, fi *os.FileInfo, m map[string]string) objectMetadata {
 	ents, err := p.meta.ListAttributes(bucket, object)
 	if err != nil || len(ents) == 0 {
 		return objectMetadata{}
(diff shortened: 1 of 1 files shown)


versity-github:

benmcclelland:
@ondrap thank you for the pull request! Yes, I think this looks good and should resolve the racing metadata. Let me run some more tests over here and I'll try to get this fix in soon.

backend/posix/posix.go
```diff
@@ -3538,11 +3521,45 @@ func (p *Posix) GetObject(_ context.Context, input *s3.GetObjectInput) (*s3.GetO
...
+	}
+	if err != nil {
+		return nil, fmt.Errorf("stat object: %w", err)
+	}
+
+	objSize := fi.Size()
+	startOffset, length, isValid, err := backend.ParseObjectRange(objSize, *input.Range)
+	if err != nil {
+		return nil, err
+	}
```
benmcclelland, line 3550: The range parsing needs to come before the directory check. AWS returns InvalidRange even for empty directory objects if the range is invalid.

ondrap: What does 'valid range' mean for a directory? I mean, the stat.Size() on a directory doesn't return any S3-relevant data, does it?

benmcclelland: It looks like any range that is parsable is not satisfiable:
```
% aws s3api get-object --bucket versity-gwtest --key dir/ --range bytes=0-0 /tmp/test

An error occurred (InvalidRange) when calling the GetObject operation: The requested range is not satisfiable
```

But if its completely non-parsable, then its ok:
```
 % aws s3api get-object --bucket versity-gwtest --key dir/ --range bytes=badrange /tmp/test
{
    "AcceptRanges": "bytes",
    "LastModified": "2025-08-07T23:29:53+00:00",
    "ContentLength": 0,
    "ChecksumCRC64NVME": "AAAAAAAAAAA=",
    "ServerSideEncryption": "AES256",
    "Metadata": {}
}
```

I guess we need some more test cases to check for these to make sure we return the correct errors in all cases. I can probably clean this up in a different PR since this is not really relevant to the current issue being addressed here.

benmcclelland (approved):

PR #1435, opened 2025-08-07
Fix O_TMPFILE Linkat race, cleanup of scoutfs integration, fix MoveData non-aligned problem

ondrap:
This PR tries to fix some other problems I have found; the problems appear when running the 's3 storage analyze' function of elasticsearch. It's necessarry to install a cluster of at least 2 elasticsearch servers and after configuring an S3 repository, the following command triggers some of these problems: 
```
curl --silent --fail-with-body -k -u elastic:password -XPOST 'https://localhost:9200/_snapshot/scout/_analyze?blob_count=2000&max_blob_size=2g&max_total_data_size=1tb&register_operation_count=100&timeout=99h&pretty'
```

* The first patch solves the race condition that arises on multiple concurrent `PutObject` and concurrent `GetObject`. The `GetObject` sometimes fails with a `404` error. The `Linkat` system command used to rename the virtual temporary file into the target destination doesn't support overwriting existing file. If the file exists, this patch creates a temporary file in the same directory and then uses the `Renameat` system call instead. 
* The second patch is mostly a cleanup of the `scoutfs.go` backend; most of the functionality was a verbatim copy from `posix.go` with no added benefit. I have removed most of the duplicate functions (I hope I didn't miss some scoutfs-specific stuff).
* The scoutfs.MoveData function requires that the data is aligned to 4096 boundary. I think the S3 specification doesn't require that; in case of multipart upload of non-4096 aligned data, the call failed. I have added a failback to use the standard `io.Copy` if the scoutfs `MoveData` fails.
* There seems to be something weird in the scoutfs syscall implementation; when called with unaligned data, the syscall correctly failed, versitygw failed back to `io.Copy`, however the elasticsearch tests still failed. I had to check for alignment before calling the `scoutfs.MoveData` and fail immediatly; the elasticsearch tests then succeeded. This is weird, there seems to be something happening even though the call fails.

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -83,8 +83,8 @@ type Posix struct {
 var _ backend.Backend = &Posix{}
 
 const (
-	metaTmpDir          = ".sgwtmp"
-	metaTmpMultipartDir = metaTmpDir + "/multipart"
+	MetaTmpDir          = ".sgwtmp"
+	MetaTmpMultipartDir = MetaTmpDir + "/multipart"
 	onameAttr           = "objname"
 	tagHdr              = "X-Amz-Tagging"
 	metaHdr             = "X-Amz-Meta"
@@ -433,7 +433,7 @@ func (p *Posix) isBucketEmpty(bucket string) error {
 			return fmt.Errorf("readdir bucket: %w", err)
 		}
 		if err == nil {
-			if len(ents) == 1 && ents[0].Name() != metaTmpDir {
+			if len(ents) == 1 && ents[0].Name() != MetaTmpDir {
 				return s3err.GetAPIError(s3err.ErrVersionedBucketNotEmpty)
 			} else if len(ents) > 1 {
 				return s3err.GetAPIError(s3err.ErrVersionedBucketNotEmpty)
@@ -448,7 +448,7 @@ func (p *Posix) isBucketEmpty(bucket string) error {
 	if errors.Is(err, fs.ErrNotExist) {
 		return s3err.GetAPIError(s3err.ErrNoSuchBucket)
 	}
-	if len(ents) == 1 && ents[0].Name() != metaTmpDir {
+	if len(ents) == 1 && ents[0].Name() != MetaTmpDir {
 		return s3err.GetAPIError(s3err.ErrBucketNotEmpty)
 	} else if len(ents) > 1 {
 		return s3err.GetAPIError(s3err.ErrBucketNotEmpty)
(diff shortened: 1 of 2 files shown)


versity-github:

backend/scoutfs/scoutfs_compat.go
```diff
@@ -58,147 +49,31 @@ func New(rootdir string, opts ScoutfsOpts) (*ScoutFS, error) {
...
-}
-
-func (tmp *tmpfile) File() *os.File {
-	return tmp.f
-}
-
-func moveData(from *os.File, to *os.File) error {
-	return scoutfs.MoveData(from, to)
+	return err
 }
```
benmcclelland, line 202: its important that we call scoutfs.MoveData() for combining the final mulitpart object in scoutfs because scoutfs doesnt support a metadata only copy_file_range(). So without this, this will double the I/O to the filesystem for multipart uploads. I think we can still use scoutfs.MoveData() with a mktemp file instead of using O_TMPFILE to have a fix that will solve the issue here.

ondrap: I do call scoutfs.MoveData, but only when it is 4096-aligned. The patch is a little confusing:

https://github.com/ondrap/versitygw/blob/pr2/backend/scoutfs/scoutfs_compat.go#L57

benmcclelland: oh! yes, sorry I missed that it was passing that function to the posix complete mp upload.

backend/posix/with_otmpfile.go
```diff
@@ -194,21 +191,33 @@ func (tmp *tmpfile) link() error {
...
-		if err != nil {
-			return fmt.Errorf("link tmpfile (fd %q as %q): %w",
-				filepath.Base(tmp.f.Name()), objPath, err)
+			if err != nil {
+				return fmt.Errorf("cannot find free temporary file: %w", err)
+			}
+
+			err = unix.Renameat(int(dirf.Fd()), tmpName, int(dirf.Fd()), filepath.Base(objPath))
+			if err != nil {
+				return fmt.Errorf("Overwriting renameat failed: %w", err)
```
benmcclelland: Small nit that static checks want error strings non capitalized to pass that check.

benmcclelland (changes requested):

ondrap:
I will fix the lowercase/uppercase letter, but looking at the code I think the `ListObjects` and `ListObjectsV2` and `fileToObj` should also be - somehow - merged together. Looking at the code it seems to me that e.g. the `fileToObj` function was updated in the `posix.go` file and the `scoutfs.go` contains an older version of the 'almost the same' function. I will have a look at this.

PR #1447, opened 2025-08-12
fix: add test cases and fix behavior for head/get range requests

benmcclelland:
This adds a bunch of test cases for non-0 len object, 0 len object, and directory objects to match verified AWS responses for the various range bytes cases.

This fixes the posix head/get range responses for these test cases as well.


tests/integration/tests.go
```diff
@@ -4339,38 +4440,149 @@ func GetObject_with_range(s *S3Conf) error {
...
+	return getObject_zero_len_with_range_helper(testName, "my-obj", s)
+}
+
+func GetObject_dir_with_range(s *S3Conf) error {
+	testName := "GetObject_dir_with_range"
+	return getObject_zero_len_with_range_helper(testName, "my-dir/", s)
+}
+
+func getObject_zero_len_with_range_helper(testName, obj string, s *S3Conf) error {
+	return actionHandler(s, testName, func(s3client *s3.Client, bucket string) error {
```
niksis02: Could we move this to `utils.go` to keep `tests.go` clear: only for the integration test cases ?

PR #1456, opened 2025-08-13
Adding -1 to etag to avoid client sdk verfications

### LINKED ISSUE #1380: azure gateway can send back etag responses that are not compatible with some clients
Apparently without the "-" in the etag, this SDK is assuming the etag is an md5 hash but azure is not sending back a compatible md5 etag:
https://github.com/aws/aws-sdk-net/blob/623dc261499331cb38bfec47789ddc4ef456222c/sdk/src/Services/S3/Custom/Internal/AmazonS3ResponseHandler.cs#L83-L97
```
                // If ETag is present and is an MD5 hash (not a multi-part upload ETag), and no byte range is specified,
                // wrap the response stream in an MD5Stream.
                // If there is a customer encryption algorithm the etag is not an MD5.
                // Exclude directory buckets as they don't support MD5
                if (!string.IsNullOrEmpty(getObjectResponse.ETag)
                    && !getObjectResponse.ETag.Contains("-")
                    && !isSse
                    && getObjectRequest.ByteRange == null
                    && !request.IsDirectoryBucket())
                {
                    string etag = getObjectResponse.ETag.Trim(etagTrimChars);
                    byte[] expectedHash = AWSSDKUtils.HexStringToBytes(etag);
                    HashStream hashStream = new MD5Stream(getObjectResponse.ResponseStream, expectedHash, getObjectResponse.ContentLength);
                    getObjectResponse.ResponseStream = hashStream;
                }
```
We can look into modifying the azure response etag since it appears the official aws sdk really wants this to be md5 if not containing "-".

_Originally posted by @benmcclelland in https://github.com/versity/versitygw/discussions/1378#discussioncomment-13700211_

nitinosiris:
Summary
This change appends -1 to the returned ETag values, effectively disabling client SDK ETag verification. This is necessary because Azure Blob Storage ETags are not S3-compatible and can cause mismatches during conditional requests or multipart operations.

Background
Azure’s ETag format does not match S3’s expectations, leading to failed operations when client SDKs attempt ETag validation.

As discussed in #1378, appending -1 to the ETag ensures the client does not attempt strict matching against Azure’s non-S3-compliant ETag values.

Fix
Append -1 to all ETag values returned by the Azure backend.

This prevents client SDKs from performing strict ETag-based validations that would otherwise fail.

Testing Done
Ran unit tests to confirm no regressions:

```
~/nitin/versitygw/cmd/versitygw # go test ./...
ok   github.com/versity/versitygw/cmd/versitygw   1.115s
```
Verified that ETag values in responses now include -1.

Impact
Resolves client-side ETag verification issues when using Azure Blob Storage as backend.

Improves compatibility for SDKs relying on ETag behavior consistent with S3.

Fixes: #1380
References: #1378

### DIFF
--- a/.github/workflows/system.yml
+++ b/.github/workflows/system.yml
@@ -33,6 +33,12 @@ jobs:
             RECREATE_BUCKETS: "true"
             DELETE_BUCKETS_AFTER_TEST: "true"
             BACKEND: "posix"
+          - set: "REST, posix, non-static, not implemented, folder IAM"
+            IAM_TYPE: folder
+            RUN_SET: "rest-not-implemented"
+            RECREATE_BUCKETS: "true"
+            DELETE_BUCKETS_AFTER_TEST: "true"
+            BACKEND: "posix"
           - set: "s3, posix, non-file count, non-static, folder IAM"
             IAM_TYPE: folder
             RUN_SET: "s3-non-file-count"
--- a/backend/azure/azure.go
+++ b/backend/azure/azure.go
@@ -362,8 +362,10 @@ func (az *Azure) PutObject(ctx context.Context, po s3response.PutObjectInput) (s
 		}
 	}
 
+	// Azure ETag values are not S3 compatible,
+	// so append "-1" to avoid client SDK ETag validation issues.
 	return s3response.PutObjectOutput{
-		ETag: string(*uploadResp.ETag),
+		ETag: string(*uploadResp.ETag) + "-1",
 	}, nil
 }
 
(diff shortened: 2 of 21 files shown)


nitinosiris:
I have not modified the test cases, if any update is needed, let me know!

versity-github:

nitinosiris:
@benmcclelland could you pls review it ?

benmcclelland:
> @benmcclelland could you pls review it ?

I think the test failures are happening because there are some missed cases of etag responses. The tests do things like head object and get object attributes and compare the responses to the list objects. So I think all of these cases have to return the modified etags.

nitinosiris:
@benmcclelland Fix the Integration test, could you review once again ?

benmcclelland:
I'm not able to modify your PR, but I grabbed your branch locally over here and modified it in this PR: https://github.com/versity/versitygw/pull/1460

Can you take a look to see if those changes are acceptable? This still has you as the author of the commit since its based on your original commit. This one adjusts all of the etag response locations so that we dont have to modify any tests.

benmcclelland:
fix merged in modified version here #1460

PR #1460, opened 2025-08-14
fix: add -1 to azure etag to avoid client sdk verfications

benmcclelland:
The C++ (and maybe others?) assume that the S3 ETags without a "-" in the string are MD5 checksums. So the Azure ETag that does not have a "-" but also is not an MD5 checksum will fail some of the sdk internal validation checks.

Fix this by appending "-1" to the ETag to make it look like the multipart format ETag that will skip the sdk verfication check.

Fixes: #1380


nitinosiris (approved):
This should be fine IMOP! Thanks @benmcclelland for refactoring it, I am away from my laptop for 2 3 days, won't be able to modify my own PR's :(

PR #1461, opened 2025-08-14
feat: bucket cors implementation

### LINKED ISSUE #1003: CORS headers - would be nice to be able to specify CORS headers for presigned urls
I would like to be able to set CORS headers to presigned URLs

niksis02:
Closes #1003

**Changes Introduced:**

1. **S3 Bucket CORS Actions**

   * Implemented the following S3 bucket CORS APIs:

     * `PutBucketCors` – Configure CORS rules for a bucket.
     * `GetBucketCors` – Retrieve the current CORS configuration for a bucket. * `DeleteBucketCors` – Remove CORS configuration from a bucket.

2. **CORS Preflight Handling**

   * Added an `OPTIONS` endpoint to handle browser preflight requests.
   * The endpoint evaluates incoming requests against bucket CORS rules and returns the appropriate `Access-Control-*` headers.

3. **CORS Middleware**

   * Implemented middleware that:

     * Checks if a bucket has CORS configured.
     * Detects the `Origin` header in the request. * Adds the necessary `Access-Control-*` headers to the response when the request matches the bucket CORS configuration.


auth/bucket_cors.go
```diff
@@ -0,0 +1,316 @@
...
+
+		// validate cors exposed headers
+		if i < exposedHeadersLen {
+			header := cr.ExposeHeaders[i]
+			if !header.IsValid() {
+				debuglogger.Logf("invalid CORS exposed header: %s", header)
+				return s3err.GetInvalidCORSHeaderErr(header.String())
+			}
+		}
+	}
```
benmcclelland: This loop is a bit awkward with the looplimit, seems like this would be a bit more readable to just break this out into separate loops
```suggestion
    for _, header := range cr.AllowedHeaders {
        if !header.IsValid() {
            debuglogger.Logf("invalid CORS allowed header: %s", header)
            return s3err.GetInvalidCORSHeaderErr(header.String())
        }
    }

    for _, method := range cr.AllowedMethods {
        if !method.IsValid() {
            debuglogger.Logf("invalid CORS allowed method: %s", method)
            return s3err.GetUnsopportedCORSMethodErr(method.String())
        }
    }

    for _, header := range cr.ExposeHeaders {
        if !header.IsValid() {
            debuglogger.Logf("invalid CORS exposed header: %s", header)
            return s3err.GetInvalidCORSHeaderErr(header.String())
        }
    }
```

auth/bucket_cors.go
```diff
@@ -0,0 +1,316 @@
...
+			if wildcardMatch(header.String(), reqHeader.String()) {
+				match = true
+				break
+			}
+		}
+
+		if !match {
+			return false, false
+		}
+	}
```
benmcclelland, line 201: slight optimization that you can cache the allowed headers in a map teh same as the allowed methods:
```
    allowedHeaders := make(map[string]struct{}, len(cr.AllowedHeaders))
    for _, h := range cr.AllowedHeaders {
        allowedHeaders[h.String()] = struct{}{}
    }
```
and then do the map lookup instead of the embedded for loop. not super critical though since i think these slices tend to be pretty small. its just a bit inconsistent that the methods are getting set to a map but the headers are not.

auth/bucket_cors.go
```diff
@@ -0,0 +1,316 @@
...
+// ParseCORSOutput parses raw bytes to 'CORSConfiguration'
+func ParseCORSOutput(data []byte) (*CORSConfiguration, error) {
+	var config CORSConfiguration
+	err := xml.Unmarshal(data, &config)
+	if err != nil {
+		debuglogger.Logf("unmarshal cors output: %v", err)
+		return nil, fmt.Errorf("failed to parse cors config: %w", err)
+	}
+
+	return &config, err
```
benmcclelland:
```suggestion
	return &config, nil
```

auth/bucket_cors_test.go
```diff
@@ -0,0 +1,610 @@
...
+	"github.com/versity/versitygw/s3err"
+)
+
+func TestCORSHeader_IsValid(t *testing.T) {
+	tests := []struct {
+		name   string
+		header CORSHeader
+		want   bool
+	}{
+		{"empty", "", true},
```
benmcclelland, line 31: just double checking that the empty header is considered valid in the spec?

s3api/controllers/bucket-get.go
```diff
@@ -187,12 +187,22 @@ func (c S3ApiController) GetBucketCors(ctx *fiber.Ctx) (*Response, error) {
...
+
+	output, err := auth.ParseCORSOutput(data)
 	return &Response{
-		Data: data,
+		Data: output,
 		MetaOpts: &MetaOptions{
 			BucketOwner: parsedAcl.Owner,
 		},
 	}, err
+
```
benmcclelland: nit: extra space added in here

s3api/middlewares/apply-bucket-cors.go
```diff
@@ -0,0 +1,98 @@
...
+// checks if origin and method meets the cors rules and
+// adds the necessary response headers.
+// CORS check is applied only when 'Origin' request header is present
+func ApplyBucketCORS(be backend.Backend) fiber.Handler {
+	return func(ctx *fiber.Ctx) error {
+		bucket := ctx.Params("bucket")
+		origin := ctx.Get("Origin")
+		// if the origin request header is empty, skip cors validation
+		if origin == "" {
+			return nil
```
benmcclelland, line 40: for the few cases this function returns nil, should it be returingin ctx.Next() instead?

s3api/middlewares/apply-bucket-cors.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		// if bucket cors is not set, skip the check
+		data, err := be.GetBucketCors(ctx.Context(), bucket)
+		if err != nil {
+			return nil
```
benmcclelland, line 52: maybe add a debuglogger line in here if the error is not an expected one:
```
            // If CORS is not configured, S3Error will have code NoSuchCORSConfiguration.
            // In this case, we can safely continue. For any other error, we should log it.
            s3Err, ok := err.(s3err.S3Error)
            if !ok || s3Err.Code != "NoSuchCORSConfiguration" {
                debuglogger.Logf("failed to get bucket cors for bucket %q: %v", bucket, err)
            }
```

auth/bucket_cors.go
niksis02: Compared to allowed methods, the allowed headers support wildcard match. That's why I didn't cache the headers in a map. A slight optimization might be applied to first cache in a map and find an exact match(if exists), then do the lookup. But if no mtach is found anyways the time complexity remains the same.

auth/bucket_cors_test.go
niksis02: If it's provided in `AllowedHeaders` or `ExposeHeaders` it's considered as valid. On the other hand if an empty header is provided in `Access-Control-Request-Headers` in `OPTIONS` call, it's considered as invalid.
e.g. `Access-Control-Request-Headers: Content-Type, ,Content-Length`. 

But this case is explicitly taken care in `ParseCORSHeaders`.

s3api/middlewares/apply-bucket-cors.go
niksis02: `ApplyBucketCORS` is passed as a callback to `controllers.ProcessHandlers` which takes care of `ctx.Next()`. Our middleware pattern has been changed a bit in advanced routing implementation.

benmcclelland (approved):

PR #1471, opened 2025-08-20
fix: update marker/continuation token to be the azure next marker

### LINKED ISSUE #1457: [Bug] VersityGW uses last key as continuation token which breaks Azure Go SDK pagination

benmcclelland:
This changes the marker/continuation token from the object name to the marker from the azure list objects pager. This is needed because passing the object name as the token to the azure next call causes the Azure API to throw 400 Bad Request with InvalidQueryParameterValue. So we have to use the azure marker for compatibility with the azure API pager.

Fixes #1457

### DIFF
--- a/backend/azure/azure.go
+++ b/backend/azure/azure.go
@@ -68,6 +68,8 @@ const (
 	onameAttr              key = "Objname"
 	onameAttrLower         key = "objname"
 	metaTmpMultipartPrefix key = ".sgwtmp" + "/multipart"
+
+	defaultListingMaxKeys = 1000
 )
 
 func (key) Table() map[string]struct{} {
@@ -579,26 +581,6 @@ func (az *Azure) GetObjectAttributes(ctx context.Context, input *s3.GetObjectAtt
 }
 
 func (az *Azure) ListObjects(ctx context.Context, input *s3.ListObjectsInput) (s3response.ListObjectsResult, error) {
-	client, err := az.getContainerClient(*input.Bucket)
-	if err != nil {
-		return s3response.ListObjectsResult{}, nil
-	}
-	pager := client.NewListBlobsHierarchyPager(*input.Delimiter, &container.ListBlobsHierarchyOptions{
-		Marker:     input.Marker,
-		MaxResults: input.MaxKeys,
-		Prefix:     input.Prefix,
-	})
-
-	var objects []s3response.Object
-	var cPrefixes []types.CommonPrefix
-	var nextMarker *string
-	var isTruncated bool
-	var maxKeys int32 = math.MaxInt32
-
-	if input.MaxKeys != nil {
-		maxKeys = *input.MaxKeys
-	}
-
 	// Retrieve the bucket acl to get the bucket owner
 	// All the objects in the bucket are owner by the bucket owner
 	aclBytes, err := az.getContainerMetaData(ctx, *input.Bucket, string(keyAclCapital))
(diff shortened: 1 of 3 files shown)


nitinosiris:
HI @benmcclelland, just pull these changes and tested it, getting the below error 
```
-----------------------------------------------------
┌─────────────────────────────────────────────────[ RESPONSE HEADERS ]─────────────────────────────────────────────────┐
│ Content-Type  : text/plain; charset=utf-8                                                                            │
│ Server        : VERSITYGW                                                                                            │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
───────────────────────────────────────────────────[ RESPONSE BODY ]────────────────────────────────────────────────────
<Error><Code>InvalidQueryParameterValue</Code><Message>400 Value for one of the query parameters specified in the request URI is invalid.</Message><Resource></Resource><RequestId></RequestId><HostId></HostId></Error>
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
03:46:29 | 400 |    7.856919ms | 127.0.0.1 | GET | /demo/ | - | continuation-token=edgev2.ps1&list-type=2&max-keys=2
```

continuation token should not be last object name!

nitinosiris:
I used below code snippet for testing this 
```
 var request = new ListObjectsV2Request
 {
     BucketName = containerName,
     Prefix = subDirectoryPath,
     MaxKeys = 2
 };

 ListObjectsV2Response response;

 do
 {
     response = amazonS3Client.ListObjectsV2Async(request).GetAwaiter().GetResult();

     foreach (var blob in response.S3Objects)
     {
         var originalLength = blob.Size.ToString();

         var blobInfo = string.Format(CultureInfo.InvariantCulture, "{0},{1},{2},{3},{4},{5},{6},{7}", containerName, blob.Key,
             "BlockBlob",
             "application/octet-stream", blob.ETag, blob.LastModified,
             originalLength,
             originalLength);

         yield return blobInfo;
     }
     request.ContinuationToken = response.NextContinuationToken;
 }
 while (response.IsTruncated.GetValueOrDefault());
```

nitinosiris:
This is how it should look like 
```
-----------------------------------------------------
┌─────────────────────────────────────────────────[ RESPONSE HEADERS ]─────────────────────────────────────────────────┐
│ Content-Length : 773                                                                                                 │
│ Content-Type  : application/xml                                                                                      │
│ Server        : VERSITYGW                                                                                            │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
───────────────────────────────────────────────────[ RESPONSE BODY ]────────────────────────────────────────────────────
<ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Name>demo</Name><ContinuationToken>2!72!MDAwMDA5IW5pdGluLnR4dCEwMDAwMjghOTk5OS0xMi0zMVQyMzo1OTo1OS45OTk5OTk5WiE-</ContinuationToken><MaxKeys>2</MaxKeys><IsTruncated>false</IsTruncated><Contents><LastModified>2025-08-20T03:45:07Z</LastModified><ChecksumType></ChecksumType><ETag>0x8DDDF9BF48CBA83-1</ETag><Key>nitin.txt</Key><Size>600</Size><StorageClass>STANDARD</StorageClass></Contents><Contents><LastModified>2025-08-12T05:58:15Z</LastModified><ChecksumType></ChecksumType><ETag>0x8DDD9653AACA5B5-1</ETag><Key>s3proxy-service-deployment.yaml.txt</Key><Size>1939</Size><StorageClass>STANDARD</StorageClass></Contents><EncodingType></EncodingType></ListBucketResult>
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
03:56:34 | 200 |    8.433893ms | 127.0.0.1 | GET | /demo/ | - | continuation-token=2%2172%21MDAwMDA5IW5pdGluLnR4dCEwMDAwMjghOTk5OS0xMi0zMVQyMzo1OTo1OS45OTk5OTk5WiE-&list-type=2&max-keys=2
```

niksis02 (dismissed):

benmcclelland:
@nitinosiris I see what the issue was before, we can't mix the object name and azure markers for partial page results. So I think the tests have some bad assumptions for this case. I need to fix the tests or skip these with the azure backend, but I think this version of the code should pass your tests now. This is now just aligning the S3 ListObjects with the Azure ListBlobsHierarchyPager for a one to one in the paginated results. This should now always keep the marker as the Azure compatible marker. The tricky part was the StartAfter in the ListObjectV2 where you can specify the object to start listing from. For this case, its keep running through the Azure paginated results until it finds the start position. But this can be a partial page blob listing, so can return fewer results in the first call to list objects than the tests assume.

nitinosiris:
HI @benmcclelland , yes the code snippet is working as expected now!

backend/azure/azure.go
```diff
@@ -578,191 +578,133 @@ func (az *Azure) GetObjectAttributes(ctx context.Context, input *s3.GetObjectAtt
...
-	})
-
-	var objects []s3response.Object
-	var cPrefixes []types.CommonPrefix
-	var nextMarker *string
-	var isTruncated bool
-	var maxKeys int32 = math.MaxInt32
-	var fetchOwner bool
 
+	var maxKeys int32 = 10000
```
nitinosiris: Just a suggestion though, aws sdk supports upto 1000 keys/per response and initializing it as 10000 could be misleading :)

benmcclelland: oh! yes, that was a typo, thanks for catching that

benmcclelland:
this still needs a little more work. turns out the list-objects marker is expected to be an object name, and when specified in the original request works like the start after in the v2 call. so while the list-objects-v2 can use the native azure marker as the continuation token, looks like we need to do a bit less optimal listings for the list-objects.

niksis02 (approved):

PR #1502, opened 2025-08-29
correct a bug when using glauth as LDAP IAM

tannevaled:
be able to debug IAM LDAP queries.
be consistent between GetUserAccount() and ListUserAccounts() on how to build the search filters.
objectClasses were missing in GetUserAccount research filter, leading to a bad result for example when a posixgGroup have the same name as a posixUser.


versity-github:

PR #1504, opened 2025-08-29
feat: add get bucket location frontend handlers

### LINKED ISSUE #1499: [Bug] - GetBucketLocation returns response for ListObjects
**Describe the bug**
GetBucketLocation should return the current configured region

**To Reproduce**
```
aws s3api get-bucket-location --bucket mybucket
```
response
```
{
    "LocationConstraint": null
}
```

access log:
```
test mybucket [28/August/2025:21:23:20 -0700] 127.0.0.1 test B5B25C87BE76CFD8 s3_ListObjects - /mybucket?location 200 - 501 0 0 0 - aws-cli/2.23.12 md/awscrt#0.23.8 ua/2.0 os/macos#24.6.0 md/arch#x86_64 lang/python#3.12.6 md/pyimpl#CPython cfg/retry-mode#standard md/installer#exe md/prompt#off md/command#s3api.get-bucket-location - - SigV4 - AuthHeader s3.us-east-1.amazonaws.com - arn:aws:s3:::/mybucket Yes
```

**Expected behavior**
Expected LocationConstraint to return the valid configured region

benmcclelland:
GetBucketLocation is being deprecated by AWS, but is still used by some clients. We don't need any backend handlers for this since the region is managed by the frontend. All we need is to test for bucket existence, so we can use HeadBucket for this.

Fixes #1499

### DIFF
--- a/auth/bucket_policy_actions.go
+++ b/auth/bucket_policy_actions.go
@@ -87,6 +87,7 @@ const (
 	PutBucketWebsiteAction                   Action = "s3:PutBucketWebsite"
 	GetBucketWebsiteAction                   Action = "s3:GetBucketWebsite"
 	GetBucketPolicyStatusAction              Action = "s3:GetBucketPolicyStatus"
+	GetBucketLocationAction                  Action = "s3:GetBucketLocation"
 
 	AllActions Action = "s3:*"
 )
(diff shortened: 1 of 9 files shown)


benmcclelland:
reference: https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html

niksis02 (commented):
Maybe we need 2 integration tests here ?
one for non exitent bucket and one for success case: when it should return the configured region.

niksis02 (approved):

PR #1505, opened 2025-08-29
fix: head object should set X-Amz-Bucket-Region on access denied

### LINKED ISSUE #1500: [Bug] - HeadBucket should set x-amz-bucket-region on error
**Describe the bug**
According to https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadBucket.html, the head bucket api call should set the x-amz-bucket-region in the response headers for error responses.

current behavior, success:
```
2025-08-28 21:34:44,051 - MainThread - botocore.parsers - DEBUG - Response headers: {'Server': 'VERSITYGW', 'Date': 'Fri, 29 Aug 2025 04:34:43 GMT', 'X-Amz-Bucket-Region': 'us-east-1', 'X-Amz-Access-Point-Alias': 'false', 'Connection': 'close'}
```
error:
```
2025-08-28 21:35:05,646 - MainThread - botocore.parsers - DEBUG - Response headers: {'Server': 'VERSITYGW', 'Date': 'Fri, 29 Aug 2025 04:35:04 GMT', 'Content-Type': 'text/plain; charset=utf-8', 'Content-Length': '230', 'Connection': 'close'}
```

**To Reproduce**
set incorrect secret key, or use an account that would not have access to the bucket, then run:
```
aws s3api --debug head-bucket --bucket mybucket
```

**Expected behavior**
following header should be set:
```
'X-Amz-Bucket-Region': 'us-east-1'
```

benmcclelland:
The HeadObject API states that the x-amz-bucket-region header will still get set for an access denied error to correctly indicate region of bucket. This is needed due to the way polices work across regions in aws, and some apps rely on this behavior.

See notes in GetBucketLocation:
In a bucket's home Region, calls to the GetBucketLocation operation are governed by the bucket's policy. In other Regions, the bucket policy doesn't apply, which means that cross-account access won't be authorized. However, calls to the HeadBucket operation always return the bucket’s location through an HTTP response header, whether access to the bucket is authorized or not. Therefore, we recommend using the HeadBucket operation for bucket Region discovery and to avoid using the GetBucketLocation operation.

Fixes #1500

### DIFF
--- a/s3api/controllers/bucket-head.go
+++ b/s3api/controllers/bucket-head.go
@@ -15,10 +15,13 @@
 package controllers
 
 import (
+	"errors"
+
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/gofiber/fiber/v2"
 	"github.com/versity/versitygw/auth"
 	"github.com/versity/versitygw/s3api/utils"
+	"github.com/versity/versitygw/s3err"
 )
 
 func (c S3ApiController) HeadBucket(ctx *fiber.Ctx) (*Response, error) {
@@ -42,6 +45,9 @@ func (c S3ApiController) HeadBucket(ctx *fiber.Ctx) (*Response, error) {
 		})
 	if err != nil {
 		return &Response{
+			Headers: map[string]*string{
+				"X-Amz-Bucket-Region": utils.GetStringPtr(region),
+			},
 			MetaOpts: &MetaOptions{
 				BucketOwner: parsedAcl.Owner,
 			},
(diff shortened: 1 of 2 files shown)


niksis02 (commented):
What about the public buckets ? 
[AuthorizePublicBucketAccess](https://github.com/versity/versitygw/blob/main/s3api/middlewares/public-bucket.go#L31) returns error if the bucket is public but doesn't grant the user access to `HeadObject`. Maybe we need to set the bucket region header there either ?

benmcclelland:
> What about the public buckets ? [AuthorizePublicBucketAccess](https://github.com/versity/versitygw/blob/main/s3api/middlewares/public-bucket.go#L31) returns error if the bucket is public but doesn't grant the user access to `HeadObject`. Maybe we need to set the bucket region header there either ?

I need to check how aws handles this, the docs are a bit limited on what the errors cases are here

PR #1506, opened 2025-08-29
cleanup: minor fixes to ldap exported functions and test

benmcclelland:
The buildSearchFilter function doesn't need to be exported, and can use strings.Builder. Add a unit test to make sure this didn't change any logic.


auth/iam_ldap.go
```diff
@@ -168,6 +169,7 @@ func (ld *LdapIAMService) GetUserAccount(access string) (Account, error) {
 	})
 
 	if ld.debug {
+		fmt.Println("LDAP Search Result")
```
niksis02: What about using our debug logger ?

benmcclelland: I was thinking about this too, but the debuglogger is in the s3api module. Maybe we need a more generic top level debuglogger? The current one relies too much on fibre ctx, so wouldnt really be appropriate for the other modules that shouldnt care about this.

niksis02: I was talking about only debuglogger.Logf, which is a simple string formatting and colors

niksis02 (approved):

PR #1507, opened 2025-08-30
feat: add response header overrides for GetObject

### LINKED ISSUE #1501: [Feature] - S3 - Override response headers

When you use the Amazon S3 REST API (or pre-signed URLs) you can control the HTTP headers that S3 sends back when someone downloads an object.

For example, adding the query parameter:

```response-content-disposition=attachment%3B%20filename%3D"example.txt"```

will make S3 include this header in the response:

```Content-Disposition: attachment; filename="example.txt"```

This is especially useful if:

- You want browsers to force a download instead of displaying inline (e.g., for images, PDFs).

- You want to override the file name the browser suggests when saving.

Similarly, you can override other response headers like:

- response-content-type
- response-cache-control
- response-content-language
- response-expires

Here is the manual on GetObject where it's explained:

https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html

I need this for my download site where I store things by their ID in the database but ensure the filename returned is correct.

### PR DESCRIPTION
**Pull request by benmcclelland**
GetObject allows overriding response headers with the following paramters:
response-cache-control
response-content-disposition
response-content-encoding
response-content-language
response-content-type
response-expires

This is only valid for signed (and pre-singed) requests. An error is returned for anonymous requests if these are set.

More info on the GetObject overrides can be found in the GetObject API reference.

Fixes #1501

### DIFF
--- a/s3api/controllers/object-get.go
+++ b/s3api/controllers/object-get.go
@@ -41,15 +41,15 @@ func (c S3ApiController) GetObjectTagging(ctx *fiber.Ctx) (*Response, error) {
 	isPublicBucket := utils.ContextKeyPublicBucket.IsSet(ctx)
 
 	err := auth.VerifyAccess(ctx.Context(), c.be, auth.AccessOptions{
-		Readonly:       c.readonly,
-		Acl:            parsedAcl,
-		AclPermission:  auth.PermissionRead,
-		IsRoot:         isRoot,
-		Acc:            acct,
-		Bucket:         bucket,
-		Object:         key,
-		Action:         auth.GetObjectTaggingAction,
-		IsBucketPublic: isPublicBucket,
+		Readonly:        c.readonly,
+		Acl:             parsedAcl,
+		AclPermission:   auth.PermissionRead,
+		IsRoot:          isRoot,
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: s3api/controllers/object-get.go [hunk 1]
```diff
@@ -380,13 +380,42 @@ func (c S3ApiController) GetObject(ctx *fiber.Ctx) (*Response, error) {
...
 
+	// Validate that response override parameters are not used with anonymous requests
+	if hasResponseOverrides && ctx.Get("X-Amz-Signature") == "" {
+		return &Response{
+			MetaOpts: &MetaOptions{
+				BucketOwner: parsedAcl.Owner,
+			},
+		}, s3err.GetAPIError(s3err.ErrAnonymousResponseHeaders)
+	}
+
```
niksis02: Instead of checking ctx.Get("X-Amz-Signature") == "", we would rather check `isPublicBucket` from `ctx.Locals`. If an anonymous request has reached to this point, it means the bucket is public. With the current check we ignore the presigned urls case, which is considered an authorized access, but in this case it's ignored.

niksis02 (approved):

PR #1511, opened 2025-09-02
Fix scoutfs backend s3 upload with non-aligned size.

ondrap:
The customMove (i.e. `scoutfs.MoveData`) won't affect the file position on the target file (unlike `io.Copy`)... so this fixes the problem of uploading files not aligned to 4096 blocks..


versity-github:

benmcclelland (approved):

PR #1515, opened 2025-09-03
fix: lex sort order of listobjects backend.Walk

### LINKED ISSUE #1283: [Bug] - ListObjects(V2) returns incorrect order with the same name directory and file objects
**Describe the bug**
`ListObjectsV2` retrun incorrectcly ordered objects list, if the bucket contains both directory and file objects with the same name.

**To Reproduce**
```bash

aws s3api put-object --bucket test --key obj+
aws s3api put-object --bucket test --key obj/
aws s3api put-object --bucket test --key obj1

aws s3api list-objects(-v2) --bucket test 
```

**Expected behavior**
The objects order in result should be: `obj+`, `obj/`, `obj1`, instead getting `obj/`, `obj+`, `obj1`

benmcclelland:
The original Walk function used the native WalkDir and ReadDir which did not
guarantee lexicographic ordering of results for cases where including directory
slash changes the sort order. This caused incorrect paginated responses because
S3 APIs require strict lexicographic ordering where directories with trailing
slashes sort correctly relative to files. For example, dir1/a.b/ must come
before dir1/a/ in the results, but fs.WalkDir was returning them in filesystem
sort order which reversed the order due to not taking in account the trailing
"/".

This also lead to cases of continuous looping of paginated listobjects results
when the marker was set out of order from the expected results.

To address this fundamental ordering issue, the entire directory traversal
mechanism was replaced with a custom lexicographic sorting approach. The new
implementation reads each directory's contents using ReadDir, then sorts the
entries using custom sort keys that append trailing slashes to directory paths.
This ensures that dir1/a.b/ correctly sorts before dir1/a/, as well as other
similar failing cases,  according to ASCII character ordering rules.

Fixes #1283

### DIFF
--- a/backend/walk.go
+++ b/backend/walk.go
@@ -19,8 +19,11 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"slices"
+	"sort"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/versity/versitygw/s3response"
@@ -36,6 +39,69 @@ type WalkResults struct {
 type GetObjFunc func(path string, d fs.DirEntry) (s3response.Object, error)
 
 var ErrSkipObj = errors.New("skip this object")
+
+var (
+	pathSeparator = "/"
+	pathDot       = "."
+)
+
+// fakeDirEntry implements fs.DirEntry for implied directories
+type fakeDirEntry struct {
+	name  string
+	isDir bool
+}
+
+func (f *fakeDirEntry) Name() string {
+	return f.name
+}
+
+func (f *fakeDirEntry) IsDir() bool {
+	return f.isDir
+}
+
+func (f *fakeDirEntry) Type() fs.FileMode {
+	if f.isDir {
+		return fs.ModeDir
+	}
+	return 0
+}
+
+func (f *fakeDirEntry) Info() (fs.FileInfo, error) {
+	return &fakeDirInfo{name: f.name, isDir: f.isDir}, nil
+}
+
+// fakeDirInfo implements fs.FileInfo for implied directories
+type fakeDirInfo struct {
+	name  string
+	isDir bool
+}
+
+func (f *fakeDirInfo) Name() string {
+	return f.name
+}
+
+func (f *fakeDirInfo) Size() int64 {
+	return 0
+}
+
+func (f *fakeDirInfo) Mode() fs.FileMode {
+	if f.isDir {
+		return fs.ModeDir
+	}
+	return 0
+}
+
+func (f *fakeDirInfo) ModTime() time.Time {
+	return time.Time{} // Return zero time for fake directories
+}
+
+func (f *fakeDirInfo) IsDir() bool {
+	return f.isDir
+}
+
+func (f *fakeDirInfo) Sys() any {
+	return nil
+}
 
 // map to store object common prefixes
 type cpMap map[string]int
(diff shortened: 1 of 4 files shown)


benmcclelland:
verified the fixed ListObjectsV2_truncated_common_prefixes does not return truncated prefixes on aws:
```
% aws s3api list-objects-v2 --bucket $bucket | grep Key   
            "Key": "d1/f1",
            "Key": "d2/f2",
            "Key": "d3/f3",
            "Key": "d4/f4",
% aws s3api list-objects-v2 --bucket $bucket --delimiter "/" --max-items 3
{
    "CommonPrefixes": [
        {
            "Prefix": "d1/"
        },
        {
            "Prefix": "d2/"
        },
        {
            "Prefix": "d3/"
        },
        {
            "Prefix": "d4/"
        }
    ],
    "RequestCharged": null,
    "Prefix": ""
}
```

But at least on azurite, azure is returning truncated prefixes from `client.NewListBlobsHierarchyPager` with `MaxResults` set, I wonder if they know about this discrepancy.

tests/integration/tests.go
```diff
@@ -6369,8 +6369,11 @@ func ListObjectsV2_truncated_common_prefixes(s *S3Conf) error {
 			return err
 		}
 
-		if !comparePrefixes([]string{"d1/", "d2/", "d3/"}, out.CommonPrefixes) {
-			return fmt.Errorf("expected the common prefixes to be %v, instead got %v", []string{"d1/", "d2/", "d3/"}, out.CommonPrefixes)
+		expected := []string{"d1/", "d2/", "d3/", "d4/"}
+
+		if !comparePrefixes(expected, out.CommonPrefixes) {
+			return fmt.Errorf("expected the common prefixes to be %v, instead got %v",
+				expected, sprintPrefixes(out.CommonPrefixes))
```
niksis02: The test was correct originally: this shouldn't be the case.
The result is limited to be 3 by `maxKeys`(3) and it should return a truncated result with `d3/` as `NextMarker`

backend/walk.go
niksis02, line 1: We should most probably have the same issue with `ListObjectVersions`: the `Walk` method responsible for parsing filesystem to object versions.

Maybe we would rather do this with another PR ?

benmcclelland: yeah, lets check if its needed, and do that in a different pr. i dont currently have a failing test for that to verify a fix

niksis02: Technically `ListObjectVersions` is the same as `ListObjects(V2)` by appending the object versions in the result. So if we don't create any object versions, it should return the exact same entries with some additional metadata(versionId ...). So it should be reproducable for `ListObjectVersions` as well.

benmcclelland:
> verified the fixed ListObjectsV2_truncated_common_prefixes does not return truncated prefixes on aws:
> 
> ```
> % aws s3api list-objects-v2 --bucket $bucket | grep Key   
>             "Key": "d1/f1",
>             "Key": "d2/f2",
>             "Key": "d3/f3",
>             "Key": "d4/f4",
> % aws s3api list-objects-v2 --bucket $bucket --delimiter "/" --max-items 3
> {
>     "CommonPrefixes": [
>         {
>             "Prefix": "d1/"
>         },
>         {
>             "Prefix": "d2/"
>         },
>         {
>             "Prefix": "d3/"
>         },
>         {
>             "Prefix": "d4/"
>         }
>     ],
>     "RequestCharged": null,
>     "Prefix": ""
> }
> ```
> 
> But at least on azurite, azure is returning truncated prefixes from `client.NewListBlobsHierarchyPager` with `MaxResults` set, I wonder if they know about this discrepancy.

aws command bug, they document the option as `--max-items`, but the real option (not documented correctly) is `--max-keys`. So the original test was correct.

PR #1516, opened 2025-09-04
Test/more list buckets, general coverage

lrm25:


tests/drivers/rest.sh
```diff
@@ -204,27 +211,48 @@ send_rest_go_command_expect_error_callback() {
...
 
+bypass_continues() {
+  status_line_idx=1
+  status_code=""
+  while true; do
+    status_line=$(sed -n "${status_line_idx}p" <<< "$result")
+    status_code=$(echo "$status_line" | awk '{print $2}')
+    if [ "$status_code" != "100" ]; then
+      break
+    fi
```
anodelman, line 228: This is a risky loop if this status is never reached.  Can you put in a limit of some sort so that you can ensure that it doesn't get stuck?

anodelman (changes requested):

Issue #1527, opened 2025-09-11
[Bug] - X-Amz-Checksum-Mode


nick-stephen:
**Describe the bug**

When we configure the AWS SDK with automatic checksums enabled, i.e.:
```
    export AWS_REQUEST_CHECKSUM_CALCULATION=WHEN_SUPPORTED;
    export AWS_RESPONSE_CHECKSUM_VALIDATION=WHEN_SUPPORTED;
```
And we execute a file write request to versity (Files.write(...)), we can see in the logs that the AWS SDK sends the header `X-Amz-Checksum-Mode: enabled`, but VeristyGW rejects this as invalid. 
```
versity_gw_ensemble_rest_provider  | [DEBUG]: invalid x-amz-checksum-mode header value: enabled

...

<Error><Code>InvalidRequest</Code><Message>Value for x-amz-checksum-mode header is invalid.</Message><Resource></Resource><RequestId></RequestId><HostId></HostId></Error>
```

Looking at versity's code it looks like the comparison is case-sensitive and it expects "ENABLED", whereas the AWS SDK is setting lower-case "enabled".

The string comparison for this header in the versity code should be case-insensitive.

**To Reproduce**

Use AWS SDK client to write to an S3 bucket with environment variable `AWS_REQUEST_CHECKSUM_CALCULATION=WHEN_SUPPORTED` set.

**Expected behavior**

The API should work with the AWS S3 client call.

**Server Version**

Versity 1.0.17, docker image

**Additional context**

AWS SDK Client version: `aws-sdk-java/2.31.40`

nick-stephen:
This looks like a suitable fix - I am not authorized to submit a MR:

```
diff --git a/s3api/controllers/object-get.go b/s3api/controllers/object-get.go
index bb31768..0c3ead8 100644
--- a/s3api/controllers/object-get.go
+++ b/s3api/controllers/object-get.go
@@ -379,7 +379,7 @@ func (c S3ApiController) GetObject(ctx *fiber.Ctx) (*Response, error) {
        key := strings.TrimPrefix(ctx.Path(), fmt.Sprintf("/%s/", bucket))
        versionId := ctx.Query("versionId")
        acceptRange := ctx.Get("Range")
-       checksumMode := types.ChecksumMode(ctx.Get("x-amz-checksum-mode"))
+       checksumMode := types.ChecksumMode(strings.ToUpper(ctx.Get("x-amz-checksum-mode", "")))
[...]

PR #1528, opened 2025-09-11
fix: #1527 - case-insensitive x-amz-checksum-mode header value

nick-stephen:
Use case-insensitive lookup of x-amz-checksum-mode header value to
support compatibility with AWS S3 client library.


versity-github:

PR #1529, opened 2025-09-11
fix: #1527 - case-insensitive x-amz-checksum-mode header value

nick-stephen:
Fix bad header value comparison to be case insensitive

### DIFF
--- a/s3api/controllers/object-get.go
+++ b/s3api/controllers/object-get.go
@@ -379,7 +379,7 @@ func (c S3ApiController) GetObject(ctx *fiber.Ctx) (*Response, error) {
 	key := strings.TrimPrefix(ctx.Path(), fmt.Sprintf("/%s/", bucket))
 	versionId := ctx.Query("versionId")
 	acceptRange := ctx.Get("Range")
-	checksumMode := types.ChecksumMode(ctx.Get("x-amz-checksum-mode"))
+	checksumMode := types.ChecksumMode(strings.ToUpper(ctx.Get("x-amz-checksum-mode")))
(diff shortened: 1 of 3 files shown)


versity-github:

s3api/controllers/object-get.go
```diff
@@ -379,7 +379,7 @@ func (c S3ApiController) GetObject(ctx *fiber.Ctx) (*Response, error) {
 	key := strings.TrimPrefix(ctx.Path(), fmt.Sprintf("/%s/", bucket))
 	versionId := ctx.Query("versionId")
 	acceptRange := ctx.Get("Range")
-	checksumMode := types.ChecksumMode(ctx.Get("x-amz-checksum-mode"))
```
niksis02, line 382: The second argument in ctx.Get in is unnecessary: anyways it will return an empty string

s3api/controllers/object-head.go
```diff
@@ -80,7 +80,7 @@ func (c S3ApiController) HeadObject(ctx *fiber.Ctx) (*Response, error) {
 		partNumber = &partNumberQuery
 	}
 
-	checksumMode := types.ChecksumMode(ctx.Get("x-amz-checksum-mode"))
```
niksis02, line 83: same here

s3api/controllers/object-get.go
nick-stephen: fixed, thx, wasn't sure so had been defensive.

s3api/controllers/object-head.go
nick-stephen: fixed

PR #1552, opened 2025-09-19
Test/direct tests two

lrm25:


tests/util/util_bucket.sh
```diff
@@ -108,6 +109,7 @@ direct_wait_for_bucket() {
   if ! check_param_count "direct_wait_for_bucket" "bucket" 1 $#; then
     return 1
   fi
+  sleep 5
```
anodelman: Is there a state that you could be waiting for?

tests/util/util_bucket.sh
```diff
@@ -120,6 +122,23 @@ direct_wait_for_bucket() {
   return 0
 }
 
+direct_wait_for_bucket_deletion() {
+  if ! check_param_count "direct_wait_for_bucket" "bucket" 1 $#; then
+    return 1
+  fi
+  sleep 5
```
anodelman: Wait for state instead of sleep?

anodelman (changes requested):

tests/util/util_bucket.sh
lrm25: Direct-to-s3, I've been having issues since sometimes it reports existence (or not), and then the opposite later.  But I can try by number of verifications of expected state and see how that works.

tests/test_rest_versioning.sh
```diff
@@ -20,138 +20,134 @@ load ./bats-assert/load
...
-  run bucket_cleanup_if_bucket_exists "$BUCKET_ONE_NAME"
+  run setup_bucket_object_lock_enabled_v2 "$BUCKET_ONE_NAME"
   assert_success
-
-  # in static bucket config, bucket will still exist
-  if ! bucket_exists "$BUCKET_ONE_NAME"; then
-    run create_bucket_object_lock_enabled "$BUCKET_ONE_NAME"
-    assert_success
-  fi
+  bucket_name="${lines[${#lines[@]} - 1]}"
```
anodelman: I think that you should consider this method to pass a scalar variable name into a function by reference for the purpose of "returning" a value - https://mywiki.wooledge.org/BashFAQ/006#eval

This would remove this weird check to get the second to last line.

lrm25: The "run" commands run in a subshell, so I don't see a way for this to work unless the bucket name is written to a tempfile, etc., and then read back.  Another possibility is to place a function before `setup_bucket_v2` that retrieves the suffix, and then passing this suffix to the function.  What do you think?

anodelman: Yeah, I think that pushing the bucket name to the func instead of attempting to pull it out of the results would be cleaner.

PR #1553, opened 2025-09-22
test: update tests to use new bucket creation function

lrm25:


tests/util/util_users.sh
```diff
@@ -419,10 +421,10 @@ change_bucket_owner() {
     return 1
   fi
   if [[ $DIRECT == "true" ]]; then
-    if ! change_bucket_owner_direct "$1" "$2" "$3" "$4"; then
```
anodelman, line 422: Since this isn't implemented, should it error instead of just skip?

PR #1558, opened 2025-09-23
fix: fixes the invalid Content-Length error

### LINKED ISSUE #1426: [Bug] - Invalid Content-Length
**Describe the bug**
Versitygw returns some custom error in the response, if `Content-Length` is invalid.
```
error when reading request headers: cannot parse Content-Length: non-numeric chars found. Buffer size=4096, contents: "PUT /test/hello HTTP/1.1\r\nHost: \r\nX-Amz-Date: 20250731T201705Z\r\nX-Amz-Storage-Class: REDUCED_REDUNDANCY\r\nAuthorization: AWS4-HMAC-SHA256 Credential=a/20250731/us-east-1/s3/aws4_request,SignedHeaders=c"..."SIX file backend storage, Versity’s open source ScoutFS filesystem, Azure Blob Storage, and other S3 servers.  \n\nThe gateway is completely stateless. Multiple Versity Gateway instances may be deploy
```

**To Reproduce**
1. Run the gateway with https.
2. generate a command with `tests/rest_scripts/put_object_openssl_chunked_trailer_example.sh`

```bash
INVALID_CHECKSUM_TYPE=false TRAILER=x-amz-checksum-crc32c TEST_MODE=false AWS_ENDPOINT_URL=127.0.0.1:7070 AWS_ACCESS_KEY_ID=access AWS_SECRET_ACCESS_KEY=secret AWS_REGION=us-east-1 BUCKET_NAME=test OBJECT_KEY=hello DATA_FILE=README.md TEST_FILE_FOLDER=. COMMAND_FILE=command.txt ./tests/rest_scripts/put_object_openssl_chunked_trailer_example.sh
```

Modify the `Content-Length` header to contain some letters:
e.g.
`Content-Length: 56asdflk24`

Make a request with openssl: `cat command.txt | openssl s_client -quiet -connect 127.0.0.1:7070`

**Expected behavior**
S3 returns an empty 400 Bad Request error status with no response body. Maybe this is some default behaviour for unexpected cases.

niksis02:
Fixes #1426

Fiber returns a custom error, if it fails to parse the `Content-Length` header. This implementation adds a check in the fiber global error handler to return an empty `400` Bad Request error, if fiber fails to parse the `Content-Length` header.

### DIFF
--- a/s3api/server.go
+++ b/s3api/server.go
@@ -16,7 +16,9 @@ package s3api
 
 import (
 	"crypto/tls"
+	"errors"
 	"net/http"
+	"strings"
 
 	"github.com/gofiber/fiber/v2"
 	"github.com/gofiber/fiber/v2/middleware/logger"
@@ -180,7 +182,16 @@ func globalErrorHandler(ctx *fiber.Ctx, er error) error {
 		// log it as a panic log
 		debuglogger.Panic(er)
 	} else {
-		// otherwise log it as an internal error
+		// handle the fiber specific errors
+		var fiberErr *fiber.Error
+		if errors.As(er, &fiberErr) {
+			if strings.Contains(fiberErr.Message, "cannot parse Content-Length") {
+				ctx.Status(http.StatusBadRequest)
+				return nil
+			}
+		}
+
+		// additionally log the internal error
 		debuglogger.InernalError(er)
 	}
 


s3api/server.go
```diff
@@ -180,7 +182,16 @@ func globalErrorHandler(ctx *fiber.Ctx, er error) error {
 		// log it as a panic log
 		debuglogger.Panic(er)
 	} else {
-		// otherwise log it as an internal error
+		// handle the fiber specific errors
+		var fiberErr *fiber.Error
+		if errors.As(er, &fiberErr) {
+			if strings.Contains(fiberErr.Message, "cannot parse Content-Length") {
```
benmcclelland, line 188: This is a bummer that we have to string match for the error here. I dont see any other way to do this though. Is it possible to somehow stick in our own content length check before fasthttp does its header checks? Or maybe we could try to submit some changes to fasthttp to return typed errors for some of these cases that we need to specifically handle.

niksis02: I need to investigate this part. Not so sure to be honest !

PR #1573, opened 2025-10-08
feat: add option to disable strict bucket name checks

### LINKED ISSUE #1564: [Bug] - posix backend support is broken (since v1.0.15)
**Describe the bug**

The posix backend does not work, I cannot even list a bucket.

* v1.0.11 - works ok
* v1.0.12 - works ok
* v1.0.13 - gives an error `[DEBUG] Internal Error, walk mybucket: file to object "myfile.txt": get bucket acl: no such key` (myfile.txt is present and correct)
* v1.0.14 - works ok
* v1.0.15 - returns error `ClientError InvalidBucketName when calling ListObjects`
* v1.0.16 - returns error `ClientError InvalidBucketName when calling ListObjects`
* v1.0.17 - returns error `ClientError InvalidBucketName when calling ListObjects`
* v1.0.18 - returns error `ClientError InvalidBucketName when calling ListObjects`

**To Reproduce**

```python
import boto3
s3=boto.resource('s3', endpoint_url='http://fs:7070/', aws_access_key_id='me', aws_secret_access_key='password')
print([x for x in s3.Bucket('mybucket').objects.all()])
```

**Expected behavior**

Expecting a list of the objects in the bucket (i.e. files in the posix filesystem).

**Server Version**

Various, see above, using your docker containers downloaded from docker hub.

Run with options:
```
podman run -d
-e ROOT_ACCESS_KEY=xx
-e ROOT_SECRET_KEY=yy
-p 7070:7070
-v /myroot/:/myroot/:ro
-v /myroot/s3:/myroot/s3:ro
-v /vgw_meta/:/myroot/vgw_meta/
-v /iam/:/etc/iam/
-v /log/:/var/log/versity/
versitygw:v1.0.18
--iam-dir /etc/iam
--admin-port 127.0.0.1:7071
--access-log /var/log/versity/access.log
posix
--sidecar /myroot/vgw_meta
--bucketlinks
/myroot/s3
```

**Additional context**

Client is Python 3.12.3 boto3 1.40.21

benmcclelland:
Some systems may choose to allow non-aws compliant bucket names and/or handle the bucket naem validation in the backend instead. This adds the option to turn off the strict bucket name validation checks in the frontend API handlers.

Fixes #1564

### DIFF
--- a/backend/common.go
+++ b/backend/common.go
@@ -570,3 +570,19 @@ func EvaluateObjectDeletePreconditions(etag string, modTime time.Time, size int6
 
 	return nil
 }
+
+// IsValidDirectoryName returns true if the string is a valid name
+// for a directory
+func IsValidDirectoryName(name string) bool {
+	// directories may not contain a path separator
+	if strings.ContainsRune(name, '/') {
+		return false
+	}
+
+	// directories may not contain null character
+	if strings.ContainsRune(name, 0) {
+		return false
+	}
+
+	return true
+}
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -78,6 +78,10 @@ type Posix struct {
 	// if the filesystem supports it. This is needed for cases where
 	// there are different filesystems mounted below the bucket level.
 	forceNoTmpFile bool
+
+	// enable posix level bucket name validations, not needed if the
+	// frontend handlers are already validating bucket names
+	validateBucketName bool
 }
 
 var _ backend.Backend = &Posix{}
@@ -131,6 +135,10 @@ type PosixOpts struct {
 	// ForceNoTmpFile disables the use of O_TMPFILE even if the filesystem
 	// supports it
 	ForceNoTmpFile bool
+	// ValidateBucketNames enables minimal bucket name validation to prevent
+	// incorrect access to the filesystem. This is only needed if the
+	// frontend is not already validating bucket names.
+	ValidateBucketNames bool
 }
 
 func New(rootdir string, meta meta.MetadataStorer, opts PosixOpts) (*Posix, error) {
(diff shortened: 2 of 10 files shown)


benmcclelland:
> Completely removing the bucket name validation contains a security risk for `posix`. Malicious requests may ruin the filesystem.
> 
> e.g. bucket: ../../bucket (allowing to upload a bucket then objects out of posix dir) bucket: /etc/passwd ...

I was thinking about this too, maybe with this disabled we can pass a flag for some minimal validation in posix still...

niksis02:
> > Completely removing the bucket name validation contains a security risk for `posix`. Malicious requests may ruin the filesystem.
> > e.g. bucket: ../../bucket (allowing to upload a bucket then objects out of posix dir) bucket: /etc/passwd ...
> 
> I was thinking about this too, maybe with this disabled we can pass a flag for some minimal validation in posix still...

I kind of don't like this solution, because we need to validate the bucket name for all the actions. What if we anyway don't allow these malicious bucket names at all for all the backends in non-strict mode ?

benmcclelland:
> > > Completely removing the bucket name validation contains a security risk for `posix`. Malicious requests may ruin the filesystem.
> > > e.g. bucket: ../../bucket (allowing to upload a bucket then objects out of posix dir) bucket: /etc/passwd ...
> > 
> > 
> > I was thinking about this too, maybe with this disabled we can pass a flag for some minimal validation in posix still...
> 
> I kind of don't like this solution, because we need to validate the bucket name for all the actions. What if we anyway don't allow these malicious bucket names at all for all the backends in non-strict mode ?

They are only malicious because of the semantics of the posix backend. Other backend types could be fine with this. Since this is a protection against a posix specific issue, I think its fine that this lives in the posix backend. It just happens that the normal bucket name checks are a superset of the posix name checks, so its fine to skip the posix check if the normal checks are enabled.

niksis02:
> > > > Completely removing the bucket name validation contains a security risk for `posix`. Malicious requests may ruin the filesystem.
> 
> > > > e.g. bucket: ../../bucket (allowing to upload a bucket then objects out of posix dir) bucket: /etc/passwd ...
> 
> > > 
> 
> > > 
> 
> > > I was thinking about this too, maybe with this disabled we can pass a flag for some minimal validation in posix still...
> 
> > 
> 
> > I kind of don't like this solution, because we need to validate the bucket name for all the actions. What if we anyway don't allow these malicious bucket names at all for all the backends in non-strict mode ?
> 
> 
> 
> They are only malicious because of the semantics of the posix backend. Other backend types could be fine with this. Since this is a protection against a posix specific issue, I think its fine that this lives in the posix backend. It just happens that the normal bucket name checks are a superset of the posix name checks, so its fine to skip the posix check if the normal checks are enabled.

Yeah it makes sense, just trying to avoid code repetition :)

PR #1581, opened 2025-10-13
feat: adds bucket policy version support

### LINKED ISSUE #1536: [Bug] - Bucket policy version
**Describe the bug**
There are 2 supported versions of bucket policy documents in S3: `2008-10-17` and `2012-10-17`.
If any other value is provided other than these 2, the gateway should return an error.

Note that if no `Version` is provided in the document, it takes `2008-10-17` by default.

**To Reproduce**
```bash

```

Where `policy.json` is:
```json
{
  "Version": "invalid policy",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": "*"
      },
      "Action": "s3:GetObject",
      "Resource": "arn:aws:s3:::test/*"
    }
  ]
}

```

**Expected behavior**
The gateway should return the following error for the above scenario:
```
An error occurred (MalformedPolicy) when calling the PutBucketPolicy operation: The policy must contain a valid version string
```

And if no `Version` is provided, the gateway should assign `2008-10-17` to it and store. Then the version should be returned in `GetBucketPolicy` call.

niksis02:
Closes #1536

Adds bucket policy version support. Two versions are supported: **2008-10-17** and **2012-10-17**. If the `Version` field is omitted in the bucket policy document, it defaults to **2008-10-17**. However, if an empty string (`""`) is provided, it is considered invalid.

### DIFF
--- a/auth/bucket_policy.go
+++ b/auth/bucket_policy.go
@@ -40,14 +40,17 @@ const (
 	policyErrInvalidFirstChar     = policyErr("Policies must be valid JSON and the first byte must be '{'")
 	policyErrEmptyStatement       = policyErr("Could not parse the policy: Statement is empty!")
 	policyErrMissingStatmentField = policyErr("Missing required field Statement")
+	policyErrInvalidVersion       = policyErr("The policy must contain a valid version string")
 )
 
 type BucketPolicy struct {
+	Version   PolicyVersion      `json:"Version"`
 	Statement []BucketPolicyItem `json:"Statement"`
 }
 
 func (bp *BucketPolicy) UnmarshalJSON(data []byte) error {
 	var tmp struct {
+		Version   *PolicyVersion
 		Statement *[]BucketPolicyItem `json:"Statement"`
 	}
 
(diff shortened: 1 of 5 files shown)


auth/bucket_policy.go
```diff
@@ -60,12 +63,22 @@ func (bp *BucketPolicy) UnmarshalJSON(data []byte) error {
 		return policyErrMissingStatmentField
 	}
 
-	// Assign the parsed value to the actual struct
+	if tmp.Version == nil {
+		// bucket policy version should defualt to '2008-10-17'
+		bp.Version = PolicyVersion2008
```
benmcclelland, line 68: Do we want this to happen for both field is absent and when the JSON explicitly sets "Version": null? Does AWS returned MalformedPolicy when `"Version": null`?

niksis02: The implementation works exactly as in S3. If `Version` is explicitly set to `null`, it should return an invalid version error(as it does). When both are missing, it returns `Missing required field Statement`, as in S3. When `Version` is set to `null` and `Statement` is missing, it should return an invalid version error(as in S3).

PR #1591, opened 2025-10-17
Allow self-signed ldaps certificates

madic-creates:
When using LDAP IAM with self-signed certificates (common in development/testing environments or internal infrastructure), VersityGW would reject the connection due to certificate verification failures. This PR adds an optional configuration to disable TLS certificate verification for LDAP connections.

**New Environment Variable**: `VGW_IAM_LDAP_TLS_SKIP_VERIFY` (default: `false`)
  - When set to `true`, disables TLS certificate verification for LDAPS connections
  - When set to `false` or omitted, enforces strict certificate validation (default behavior)

## Testing

- ✅ Code compiles successfully
- ✅ Existing LDAP tests passed
- ✅ Docker image builds and runs
- ✅ Deployed on Local OnPrem Kubernetes Environment and successfully uploaded files after authentication

### DIFF
--- a/auth/iam.go
+++ b/auth/iam.go
@@ -119,6 +119,7 @@ type Opts struct {
 	LDAPRoleAtr                 string
 	LDAPUserIdAtr               string
 	LDAPGroupIdAtr              string
+	LDAPTLSSkipVerify           bool
 	VaultEndpointURL            string
 	VaultNamespace              string
 	VaultSecretStoragePath      string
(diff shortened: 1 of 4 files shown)


versity-github:

madic-creates:
Is it possible to re-run the failed test? From my point of view, the failed test `9 REST - UploadPart w/o part number` doesn't seem to be related to my adjustments but more like a temporary networking problem.

niksis02:
> Is it possible to re-run the failed test? From my point of view, the failed test `9 REST - UploadPart w/o part number` doesn't seem to be related to my adjustments but more like a temporary networking problem.

No worries about the test failure. It's not related to your changes

benmcclelland (approved):

PR #1658, opened 2025-11-20
test: create-bucket, get-bucket-location

lrm25:


tests/test_rest_get_bucket_location.sh
```diff
@@ -0,0 +1,54 @@
...
+# under the License.
+
+load ./bats-support/load
+load ./bats-assert/load
+
+source ./tests/drivers/create_bucket/create_bucket_rest.sh
+source ./tests/drivers/get_bucket_location/get_bucket_location_rest.sh
+source ./tests/setup.sh
+
+@test "REST - GetBucketLocation - no bucket" {
```
anodelman, line 24: Are all the buckets created for testing here cleaned up?  I don't see a teardown function.

lrm25: The bucket teardown function is `teardown` in `./tests/setup.sh`

PR #1668, opened 2025-12-01
fix: fixes ipa iam GetUserAccount id parsing panic

niksis02:
In the IPA `GetUserAccount` function, the `userID`, `groupID`, and `projectID` values were parsed from a string slice into integers, which caused a panic when the slice was empty. This has been fixed by defaulting empty slices to `0` for each ID.

### DIFF
--- a/auth/iam_ipa.go
+++ b/auth/iam_ipa.go
@@ -140,17 +140,17 @@ func (ipa *IpaIAMService) GetUserAccount(access string) (Account, error) {
 		return Account{}, err
 	}
 
-	uid, err := strconv.Atoi(userResult.Uidnumber[0])
+	uid, err := parseToInt(userResult.Uidnumber, "userID")
 	if err != nil {
-		return Account{}, fmt.Errorf("ipa uid invalid: %w", err)
+		return Account{}, err
 	}
-	gid, err := strconv.Atoi(userResult.Gidnumber[0])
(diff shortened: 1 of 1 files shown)


auth/iam_ipa.go
```diff
@@ -500,3 +500,20 @@ func (b *Base64Encoded) UnmarshalJSON(data []byte) error {
 	*b, err = base64.StdEncoding.DecodeString(intermediate)
 	return err
 }
+
+// parseToInt parses the first argument of input string slice
+// to an integer. If slice is empty, it defaults to 0
+func parseToInt(input []string, argName string) (int, error) {
+	if len(input) == 0 {
+		debuglogger.IAMLogf("empty %s slice: defaulting to 0")
```
benmcclelland: missing argName for %s

auth/iam_ipa.go
```diff
@@ -500,3 +500,20 @@ func (b *Base64Encoded) UnmarshalJSON(data []byte) error {
...
+		return 0, nil
+	}
+
+	id, err := strconv.Atoi(input[0])
+	if err != nil {
+		debuglogger.IAMLogf("failed to parse %s: %w", argName, err)
```
benmcclelland: I think %w is only available for fmt.Errorf wrapping. This might need to be %v instead here.

PR #1670, opened 2025-12-02
SQLite metadata backend and Sidecar directory files as JSON

spinto:
Hello,
I kingly ask to consider integrating in the main versitygw gateway codebase two updates to the versitygw software from our local installation.

One, commit 907704abe7b83f3c0e8accb0bee5b1c0e59ba985 , updates the sidecar directory metadata backend by encoding the metadata in a single JSON file. This helps in having less small files in the file system and avoid collisions when you start versitygw on filesystems who do not support xattr. Note that this breaks compatibility with the old sidecar directory metadata format.

The second, commit b11dfd22aa02cdc72a19b3714ed2b9cdb65025a5 , adds an "sqlite" directory metadata backend, via the --sqlite flag in the command line. The sqlite metadata backend will store the metadata into multiple sqlite databases in the sqlite metadata directory, one per bucket. This helps when you start versitygw on filesystems who do not support xattr when you do not want plenty of small files around to store the metadata. This update should not impact any other software functionality, it requires anyway adding the SQLite database libraries which will increase the size of the versitygw executable.

Feel free to ask further information on the changes or any other thing or change which would be required to integrate them.


versity-github:

benmcclelland:
These would need to be separate PRs to be considered. My main concern with both of them is how you can make this consistent when multiple gateways are running at the same time with load balancing. Neither of these changes appear to be safe in a multi-gateway environment.

spinto:
Right, ok, I will have a look about it and submit two separate PRs. For the sqlite, if the database is into a shared NFS, it works, although it is not a recommended solution... postgres could be used instead.

Regarding the changes on the sidecar, even the old implementation I believe has the problem of not working when there is load balancing in place...

PR #1677, opened 2025-12-05
fix: Checks that x-amz-decoded-content-length matches the actual payload in unsigned streaming upload

### LINKED ISSUE #1676: [Bug] - `STREAMING-AWS4-HMAC-PAYLOAD` - payload bytes and decoded content length mismatch
**Describe the bug**
If the `x-amz-decoded-content-length` request header value doesn’t match the actual number of payload bytes, S3 returns a specific error. In VersityGW, however, this header is ignored, and the object is uploaded successfully as long as the request body encoding is correct.

**To Reproduce**
```go
package main

import (
	"bytes"
	"context"
	"fmt"
	"io"
	"log"
	"net/http"
	"time"

	"github.com/aws/aws-sdk-go-v2/aws"
	v4 "github.com/aws/aws-sdk-go-v2/aws/signer/v4"
)

var (
	accessKeyId = "<access_key_id>"
	secret      = "<secret_access_key>"
	region      = "us-east-1"
	service     = "s3"
)

func main() {
	signer := v4.NewSigner()

	reqBody := []byte("a\r\ndummy data\r\n0\r\n\r\n")

	req, err := http.NewRequest(http.MethodPut, "http://127.0.0.1:7070/test/hello", bytes.NewReader(reqBody))
	if err != nil {
		log.Fatalf("failed to create the request: %v", err)
	}

	req.Header.Set("x-amz-content-sha256", "STREAMING-UNSIGNED-PAYLOAD-TRAILER")
	req.Header.Set("x-amz-decoded-content-length", "5")

	err = signer.SignHTTP(context.Background(), aws.Credentials{
		AccessKeyID:     accessKeyId,
		SecretAccessKey: secret,
	}, req, "STREAMING-UNSIGNED-PAYLOAD-TRAILER", service, region, time.Now())
	if err != nil {
		log.Fatalf("failed to sign the request: %v", err)
	}

	client := &http.Client{}
	resp, err := client.Do(req)
	if err != nil {
		log.Fatalf("failed to send the request: %v", err)
	}

	defer resp.Body.Close()
	body, err := io.ReadAll(resp.Body)
	if err != nil {
		log.Fatalf("failed to read the request body: %v", err)
	}

	fmt.Println(string(body))
}

```

**Expected behavior**
By running this script, an error should be returned similiar to the following:
```xml
<Error><Code>IncompleteBody</Code><Message>You did not provide the number of bytes specified by the Content-Length HTTP header</Message><NumberBytesExpected>5</NumberBytesExpected><NumberBytesProvided>10</NumberBytesProvided><RequestId>WZPYQFT3SPK1P8FV</RequestId><HostId>ZBDDR4i0aUm9RBku5Bx10/ukuHH3rMAv9KdNipM8y0n92o+GfSW8TgrifeDgBmuS4dcUeICQG4k=</HostId></Error>
```

niksis02:
Fixes #1676

`x-amz-decoded-content-length` in streaming uploads specifies the number of actual data-payload bytes, with encoding characters removed. If the value does not match the actual payload after decoding, now an `IncompleteBody` error is returned.

### DIFF
--- a/s3api/utils/unsigned-chunk-reader.go
+++ b/s3api/utils/unsigned-chunk-reader.go
@@ -50,9 +50,13 @@ type UnsignedChunkReader struct {
 	// this data is necessary for 'InvalidChunkSizeError' error
 	// TODO: add 'Chunk' and 'BadChunkSize' in the error
 	chunkSizes []int64
+	cLength    int64
+	// This data is necessary for the decoded content length mismatch error
+	// TODO: add 'NumberBytesExpected' and 'NumberBytesProvided' in the error
+	dataRead int64
 }
 
-func NewUnsignedChunkReader(r io.Reader, ct checksumType) (*UnsignedChunkReader, error) {
+func NewUnsignedChunkReader(r io.Reader, ct checksumType, decContentLength int64) (*UnsignedChunkReader, error) {
 	var hasher hash.Hash
 	var err error
 	if ct != "" {
@@ -70,6 +74,7 @@ func NewUnsignedChunkReader(r io.Reader, ct checksumType) (*UnsignedChunkReader,
 		stash:        make([]byte, 0),
 		hasher:       hasher,
 		chunkSizes:   []int64{},
+		cLength:      decContentLength,
 	}, nil
 }
 
@@ -103,6 +108,8 @@ func (ucr *UnsignedChunkReader) Read(p []byte) (int, error) {
 		if err != nil {
 			return 0, err
 		}
+
+		ucr.dataRead += chunkSize
 
 		if chunkSize == 0 {
 			// Stop reading parsing payloads as 0 sized chunk is reached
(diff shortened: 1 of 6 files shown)


s3api/utils/unsigned-chunk-reader.go
```diff
@@ -50,9 +50,13 @@ type UnsignedChunkReader struct {
 	// this data is necessary for 'InvalidChunkSizeError' error
 	// TODO: add 'Chunk' and 'BadChunkSize' in the error
 	chunkSizes []int64
+	cLength    int64
+	// This data is necessary for the decoded content length mismatch error
+	// TODO: add 'NumberBytesExpected' and 'NumberBytesProvided' in the error
```
benmcclelland, line 55: is this todo still needed? looks like its logged. does it need to be part of the s3err?

PR #1717, opened 2025-12-29
fix: cleanup sidecar metadata empty dirs

loktionovam:
Clean up empty sidecar metadata directories after deletions to prevent inode count growth caused by lingering empty directories.


versity-github:

benmcclelland (approved):
This looks good. One future optimization might be calling os.Open() + Readdirnames(1) instead of os.ReadDir() since the check is just looking for empty directories. This might speed up large directories.

PR #1737, opened 2026-01-06
feat: add option for default global cors allow origin headers

benmcclelland:
There is some desire to have a web dashboard for the gateway. So that we dont have to proxy all requests through the webserver and expose credentials over the wire, the better approach would be to enable CORS headers to allow browser requests directly to the s3/admin service.

The default for these headers is off, so that they are only enabled for instances that specfically want to support this workload.

### DIFF
--- a/s3api/middlewares/apply-bucket-cors-preflight.go
+++ b/s3api/middlewares/apply-bucket-cors-preflight.go
@@ -1,0 +1,73 @@
+// Copyright 2026 Versity Software
+// This file is licensed under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package middlewares
+
+import (
+	"strings"
+
+	"github.com/gofiber/fiber/v2"
+	"github.com/versity/versitygw/backend"
+	"github.com/versity/versitygw/s3err"
+)
+
+// ApplyBucketCORSPreflightFallback handles CORS preflight (OPTIONS) requests for S3 routes
+// when no per-bucket CORS configuration exists.
+//
+// If the bucket has no CORS configuration and fallbackOrigin is set, it responds with 204 and:
+// - Access-Control-Allow-Origin: fallbackOrigin
+// - Vary: Origin, Access-Control-Request-Headers, Access-Control-Request-Method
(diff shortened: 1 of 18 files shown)


s3api/middlewares/apply-bucket-cors.go
```diff
@@ -46,12 +49,32 @@ func ApplyBucketCORS(be backend.Backend) fiber.Handler {
...
+					if len(ctx.Response().Header.Peek("Access-Control-Allow-Origin")) == 0 {
+						ctx.Response().Header.Add("Access-Control-Allow-Origin", fallbackOrigin)
+					}
+					if len(ctx.Response().Header.Peek("Vary")) == 0 {
+						ctx.Response().Header.Add("Vary", VaryHdr)
+					}
+					ensureExposeETag(ctx)
+				}
+				return nil
+			}
```
niksis02, line 65: What about the case when bucket cors is set and `origin` isn't? Shouldn't it fall back to `fallbackOrigin` ?

s3api/middlewares/apply-default-cors-preflight.go
```diff
@@ -0,0 +1,62 @@
...
+func ApplyDefaultCORSPreflight(fallbackOrigin string) fiber.Handler {
+	fallbackOrigin = strings.TrimSpace(fallbackOrigin)
+	if fallbackOrigin == "" {
+		return func(ctx *fiber.Ctx) error { return nil }
+	}
+
+	return func(ctx *fiber.Ctx) error {
+		if ctx.Method() != fiber.MethodOptions {
+			return nil
+		}
```
niksis02: Unnecessary check ? Seems the middleware should appear only in `OPTIONS` routes, and the router will take care the method check.

s3api/middlewares/apply-bucket-cors-preflight.go
```diff
@@ -0,0 +1,77 @@
...
+func ApplyBucketCORSPreflightFallback(be backend.Backend, fallbackOrigin string) fiber.Handler {
+	fallbackOrigin = strings.TrimSpace(fallbackOrigin)
+	if fallbackOrigin == "" {
+		return func(ctx *fiber.Ctx) error { return ctx.Next() }
+	}
+
+	return func(ctx *fiber.Ctx) error {
+		if ctx.Method() != fiber.MethodOptions {
+			return ctx.Next()
+		}
```
niksis02: Same for the method check

s3api/middlewares/apply-bucket-cors.go
benmcclelland: CORS is only meaningful when the client sends an Origin. Browsers will always include Origin on cross-origin requests, so this case is typically non-browser / same-origin / internal traffic.

Also, I wanted to make sure that we were not overriding any specific bucket CORS config that is in place already. If the bucket config exists, then fallbackOrigin would effectively override that. I dont want to evaluate per bucket rules here, and only want to install the default if the rules dont already exist.

Unless there is a case we can think of for why we would want to override bucket config. I think I'm ok with the web explorer being broken for bucket access with an incorrect CORS config though.

niksis02 (dismissed):

PR #1748, opened 2026-01-12
fix: optimize sidecar empty-dir checks

loktionovam:
Optimize the cleanup of empty sidecar metadata directories after deleting large directories, as suggested in the review comment: https://github.com/versity/versitygw/pull/1717#pullrequestreview-3620475928


versity-github:

PR #1763, opened 2026-01-17
feat: add web-based UI for S3 object management and admin operations

benmcclelland:
Implements a web interface for VersityGW with role-based access:
- Object explorer for all users to browse, upload, and download S3 objects
- Admin dashboard showing system overview and gateway status
- Admin-only user management for IAM user administration
- Admin-only bucket management for creating and configuring S3 buckets
- User authentication with automatic role-based page access

The web UI is disabled by default and only enabled with the --webui or VGW_WEBUI_PORT env options that specify the listening address/port for the web UI server. This preserves previous version behavior to not enable any new ports/services unless opted in.

Login to the web UI login page with accesskey/secretkey credentials as either user or admin account. UI functionality will auto detect login role.

Regular users have access to the object explorer for managing files within their accessible buckets. Admins additionally have access to user and bucket management interfaces. The web UI is served on a separate port from the S3 server and integrates with existing S3 and Admin API endpoints.

All requests to the S3 and Admin services are signed by the browser and sent directly to the S3/Admin service handlers. The login credentials are never sent over the network for security purposes. This requires the S3/Admin service to configure CORS Access-Control-Allow-Origin headers for these requests.


benmcclelland:
@niksis02 focus on the cli options and server calling for this PR. The js/html we can continue to iterate on in future PRs. I just wanted to get this into this repo so we preserve the git history from here on out.

webui/webserver.go
```diff
@@ -0,0 +1,138 @@
...
+	addr := strings.TrimSpace(s.config.ListenAddr)
+	if addr == "" {
+		return fmt.Errorf("webui: listen address is required")
+	}
+
+	// Check if TLS is configured
+	if s.config.TLSCert != "" && s.config.TLSKey != "" {
+		return s.app.ListenTLS(addr, s.config.TLSCert, s.config.TLSKey)
+	}
+
```
niksis02, line 133: Maybe it would make sense to add a log here?
e.g.
```go
fmt.Printf("Initializing web dashboard on %s", addr)
```

benmcclelland: added to NewServer() instead so that it prints this before the banner like all other subsystems. and errors from the listenaddr will be printed later on.

PR #1777, opened 2026-01-22
Reload TLS certificates on SIGHUP

### LINKED ISSUE #1299: [Feature] - Gracefull reload for changed TLS certificates
It would be great to be able to systemctl reload for graceful loading of changed TLS key/cert without service interruption. My understanding is, that as of now one has to restart the gateway process entirely

**Additional context**
Certificates with a short duration are becoming very common. This even more so as letsencrypt actively drives things for shorter duration certificates.
With our implementation we task Apache mod_md to automatically get new certificates after 2/3 of the certs valid period and have systemd units in place to copy these privkey.pem and pubkey.pem into /etc/versitygw.d/ and restart the versitygw@...service
Since we use software which accesses S3 Storage through Versity Gateway in background, interruption of the gateway should not happen.

This said: Thank you guys for providing a free wonderfully designed S3 Gateway. It helps us a lot.

eest:
* Add utils.CertStorage for holding cert data that can be updated at runtime.
* Add utils.NewTLSListener() to have a central place to control e.g. TLS MinVersion across different servers.
* Move server option setting to after server.app is set so we can call app.SetTLSHandler() in the option code. This is handled by fiber when calling app.ListenTLSWithCertificate() but not when calling app.Listener() directly.
* Add WithTLS() to webserver code so it looks more like the other servers.

Fixes #1299

I have not used fiber before, so I am unsure if we actually want to call app.SetTLSHandler() as well as TLSHandler.GetClientInfo(info). Including those things was an attempt to not clobber the functionality setup when calling app.ListenTLSWithCertificate(), but it also looks a bit weird that the fiber handler writes to a field with no locking:
https://github.com/gofiber/fiber/blob/main/ctx.go#L85-L91

Also the recipe for autocert usage at https://github.com/gofiber/recipes/blob/master/autocert/main.go does not seem to deal with that either... So let me know if I should just drop that (i guess it depended on if calling fiber.Ctx.ClientHelloInfo() is important for this project or not).

### DIFF
--- a/s3api/admin-server.go
+++ b/s3api/admin-server.go
@@ -15,8 +15,6 @@
 package s3api
 
 import (
-	"crypto/tls"
-
 	"github.com/gofiber/fiber/v2"
 	"github.com/gofiber/fiber/v2/middleware/logger"
 	"github.com/gofiber/fiber/v2/middleware/recover"
@@ -25,6 +23,7 @@ import (
 	"github.com/versity/versitygw/debuglogger"
 	"github.com/versity/versitygw/s3api/controllers"
 	"github.com/versity/versitygw/s3api/middlewares"
+	"github.com/versity/versitygw/s3api/utils"
 	"github.com/versity/versitygw/s3log"
 )
 
@@ -33,7 +32,7 @@ type S3AdminServer struct {
 	backend         backend.Backend
 	router          *S3AdminRouter
 	port            string
-	cert            *tls.Certificate
+	CertStorage     *utils.CertStorage
 	quiet           bool
 	debug           bool
 	corsAllowOrigin string
@@ -88,8 +87,8 @@ func NewAdminServer(be backend.Backend, root middlewares.RootUserConfig, port, r
 
 type AdminOpt func(s *S3AdminServer)
 
-func WithAdminSrvTLS(cert tls.Certificate) AdminOpt {
-	return func(s *S3AdminServer) { s.cert = &cert }
+func WithAdminSrvTLS(cs *utils.CertStorage) AdminOpt {
+	return func(s *S3AdminServer) { s.CertStorage = cs }
 }
 
 // WithQuiet silences default logging output
@@ -109,8 +108,13 @@ func WithAdminCORSAllowOrigin(origin string) AdminOpt {
 }
 
 func (sa *S3AdminServer) Serve() (err error) {
-	if sa.cert != nil {
-		return sa.app.ListenTLSWithCertificate(sa.port, *sa.cert)
+	if sa.CertStorage != nil {
+		ln, err := utils.NewTLSListener(sa.app.Config().Network, sa.port, sa.CertStorage.GetCertificate)
+		if err != nil {
+			return err
+		}
+
+		return sa.app.Listener(ln)
 	}
 	return sa.app.Listen(sa.port)
 }
--- a/s3api/utils/utils.go
+++ b/s3api/utils/utils.go
@@ -16,11 +16,13 @@ package utils
 
 import (
 	"bytes"
+	"crypto/tls"
 	"encoding/base64"
 	"encoding/xml"
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"regexp"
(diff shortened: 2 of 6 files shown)


versity-github:

benmcclelland:
@eest Thank you for taking this one on! This is the general approach that I was originally thinking as well in https://github.com/versity/versitygw/issues/1299.

s3api/utils/utils.go
```diff
@@ -910,3 +913,53 @@ func GenerateObjectLocation(ctx *fiber.Ctx, virtualDomain, bucket, object string
...
+		TLSHandler: &fiber.TLSHandler{},
+	}
+}
+
+func (cs *CertStorage) GetCertificate(info *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	// Mimic GetCertificate usage in app.ListenTLSWithCertificate()
+	_, err := cs.TLSHandler.GetClientInfo(info)
+	if err != nil {
+		return nil, err
+	}
```
benmcclelland: I'm surprised that this part is needed. I was expecting that we could just simply return cs.cert here. @niksis02 do you know if cs.TLSHandler.GetClientInfo(info) is needed here?

s3api/admin-server.go
```diff
@@ -88,8 +87,11 @@ func NewAdminServer(be backend.Backend, root middlewares.RootUserConfig, port, r
 
 type AdminOpt func(s *S3AdminServer)
 
-func WithAdminSrvTLS(cert tls.Certificate) AdminOpt {
-	return func(s *S3AdminServer) { s.cert = &cert }
+func WithAdminSrvTLS(cs *utils.CertStorage) AdminOpt {
+	return func(s *S3AdminServer) {
+		s.app.SetTLSHandler(cs.TLSHandler)
```
benmcclelland: I also suspect we dont need s.app.SetTLSHandler(cs.TLSHandler) here

s3api/utils/utils.go
```diff
@@ -910,3 +913,53 @@ func GenerateObjectLocation(ctx *fiber.Ctx, virtualDomain, bucket, object string
...
+func (cs *CertStorage) SetCertificate(certFile string, keyFile string) error {
+	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
+	if err != nil {
+		return fmt.Errorf("unable to set certificate: %w", err)
+	}
+
+	cs.mu.Lock()
+	cs.cert = &cert
+	cs.mu.Unlock()
+
```
benmcclelland, line 935: optional: might be useful to have a log message on successful cert reload for confirmation that it worked

eest: @benmcclelland: Yes this is probably helpful. We currently don't have the context to know _which_ TLS cert we are updating here (s3, admin or web) so either we should pass in an identifying string or similar, or instead handle logging at the call site in the SIGHUP handler where we know which one we are operating on. What would you prefer? Also which log call in particular is fitting for this?

s3api/utils/utils.go
eest: @benmcclelland: Yes I agree this feels a bit weird. I only noticed that your current usage of fiber called either `app.ListenTLS()` (which calls `app.ListenTLSWithCertificate()`) or `app.ListenTLSWithCertificate()` directly and that one installs a GetCertificate handler for updating a field in the TLSHandler struct: https://github.com/gofiber/fiber/blob/v2.52.10/listen.go#L126 https://github.com/gofiber/fiber/blob/v2.52.10/ctx.go#L107-L113

Its not obvious to me if you wanted this or not, but let me know and I'll clean that up.

s3api/admin-server.go
eest: Same as cs.TLSHandler.GetClientInfo(info) above, I'll drop this as well if that is not needed.

s3api/utils/utils.go
niksis02: @benmcclelland @eest 
The TLS handshake is already in progress when tls.Config.GetCertificate is invoked. At that point, Go’s TLS stack has already parsed the ClientHello.

I think it's simply an unnecessary overhead.

s3api/utils/utils.go
```diff
@@ -910,3 +913,53 @@ func GenerateObjectLocation(ctx *fiber.Ctx, virtualDomain, bucket, object string
 		obj,
 	)
 }
+
+type CertStorage struct {
+	cert       *tls.Certificate
```
niksis02: As this is a ready-heavy operation, I would recommend using `atomic.Value` for cert read/update, instead of `RWMutex`. In this way we would avoid lock overhead.

s3api/utils/utils.go
eest: @niksis02 it looks to me like it is there to make the data available to https://docs.gofiber.io/api/ctx/#clienthelloinfo (though the lack of locking around it makes me wonder about possible race conditions anyway). Drop this for now then?

s3api/utils/utils.go
eest: @niksis02 That is a good point! As we are dealing with a `*tls.Certificate` it might even be a good idea to use `atomic.Pointer[tls.Certificate]`. I'll update the patch using that.

eest:
Updated the patch:
* No more magic around the `app.SetTLSHandler()` or `TLSHandler.GetClientInfo(info)`, so also moved back the opts looping to the original location.
* Replace `sync.RWMutex` with `atomic.Pointer[tls.Certificate]` as pointed out by @niksis02 (thanks!)
* Add basic `fmt.Printf()` for the sigHup logging as a reference, but there is probably better loggers to use, just not sure which one is appropriate.
* It is currently possible to call `GetCertificate()` without having called `SetCertificate()` which will return `nil, nil` instead of `nil, error`. This is valid as per tls.Config.GetCertificate:
```
// If GetCertificate is nil or returns nil, then the certificate is
// retrieved from NameToCertificate. If NameToCertificate is nil, the
// best element of Certificates will be used.
```
... but one might want to error out rather than falling back i guess.

s3api/utils/utils.go
```diff
@@ -910,3 +912,39 @@ func GenerateObjectLocation(ctx *fiber.Ctx, virtualDomain, bucket, object string
...
+	cert atomic.Pointer[tls.Certificate]
+}
+
+func NewCertStorage() *CertStorage {
+	return &CertStorage{}
+}
+
+func (cs *CertStorage) GetCertificate(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	return cs.cert.Load(), nil
+}
```
niksis02, line 926: Unnecessary `error` return value

eest: The signature of `GetCertificate()` needs to match
```
GetCertificate func(*ClientHelloInfo) (*Certificate, error)
```
from `tls.Config.GetCertificate()`

niksis02: Ahh, I see. Thank you

benmcclelland:
looks good, I think the logging and fallback handling can be figured out in future PRs

Issue #1702, opened 2025-12-16
[Bug] - Single PUT upload fails on Windows when parent folder does not yet exist


khambrecht:
**Describe the bug**
Although the gateway is not supported for Windows, I gave it a try and it looked very promising.  Using posix backend with sidecar directory for meta data.

Uploading a small file with a single PUT command fails if the parent folder structure does not yet exist.
Multipart uploads, on the other hand, work even if the folder structure does not yet exist.

**To Reproduce**
compile and run GW on Windows with posix backend and sidecar dir, and create bucket:
`./versitygw.exe  --access xxxx  --secret xxxx  posix --sidecar S:/gw/sidecar/ S:/gw/data/`

`aws  --endpoint-url http://x.x.x.x:7070 --region us-east-1  s3 mb s3://mybucket`

upload small file via aws cli to a not yet existing dir:
`aws  --endpoint-url http://x.x.x.x:7070  --region us-east-1  s3 cp small.txt  s3://mybucket/path/not/exists/`
`upload failed: ./small.txt to s3://mybucket/path/not/exists/small.txt An error occurred (ObjectParentIsFile) when calling the PutObject operation: Object parent already exists as a file.`

however multipart upload works:
`aws  --endpoint-url http://x.x.x.x:7070  --region us-east-1    s3 cp large.bin  s3://mybucket/path/not/exists/`
`upload: ./large.bin to s3://mybucket/path/not/exists/large.bin`

once the parent folder structure exists, also the singe PUT upload works:
`aws   --endpoint-url http://x.x.x.x:7070  --region us-east-1    s3 cp small.txt  s3://mybucket/path/not/exists/`
`upload: ./small.txt to s3://mybucket/path/not/exists/small.txt `

**Expected behavior**
single PUT upload should work, even when parten folder dies not exist

**Server Version**
Version  : v1.0.19
Build    : 8fc5620
BuildTime: 2025-12-16_06:17:09PM
Windows_NT WIN-SFPKCHEHA5J 10.0 20348 x86_64 MS/Windows

**Additional context**
The problem is the different behavior of the `os.Stat` command in Windows compared to *nix in `backend/posix/posix.go` in function `PutObjectWithPostFunc`. The `os.Stat` returns `ENOTDIR`
[...]

PR #1782, opened 2026-01-22
Fix Single PUT on Windows

khambrecht:
This will fix the Single PUT upload on Windows when the parent directory does not exist.
Will fix https://github.com/versity/versitygw/issues/1702

### DIFF
--- a/backend/posix/parentdir_other.go
+++ b/backend/posix/parentdir_other.go
@@ -18,9 +18,8 @@ package posix
 
 import (
 	"github.com/versity/versitygw/s3err"
-	"github.com/versity/versitygw/s3response"
 )
 
-func handleParentDirError(_ string) (s3response.PutObjectOutput, error) {
-	return s3response.PutObjectOutput{}, s3err.GetAPIError(s3err.ErrObjectParentIsFile)
+func handleParentDirError(_ string) error {
+	return s3err.GetAPIError(s3err.ErrObjectParentIsFile)
 }
--- a/backend/posix/parentdir_windows.go
+++ b/backend/posix/parentdir_windows.go
@@ -19,18 +19,28 @@ package posix
 import (
 	"os"
 	"path/filepath"
-	"syscall"
 
 	"github.com/versity/versitygw/s3err"
-	"github.com/versity/versitygw/s3response"
 )
 
-func handleParentDirError(name string) (s3response.PutObjectOutput, error) {
+func handleParentDirError(name string) error {
 	dir := filepath.Dir(name)
-	d, statErr := os.Stat(dir)
-	if statErr == nil && !d.IsDir() {
-		return s3response.PutObjectOutput{}, s3err.GetAPIError(s3err.ErrObjectParentIsFile)
+
+	// Walk up the directory hierarchy
+	for dir != "." && dir != "/" {
+		d, statErr := os.Stat(dir)
+		if statErr == nil {
+			// Path component exists
+			if !d.IsDir() {
+				// Found a file in the ancestor path
+				return s3err.GetAPIError(s3err.ErrObjectParentIsFile)
+			}
+			// Found a valid directory ancestor, parent truly doesn't exist
+			break
+		}
+		// Continue checking parent directories
+		dir = filepath.Dir(dir)
 	}
 	// Parent doesn't exist or is a directory, treat as ENOENT
-	return s3response.PutObjectOutput{}, syscall.ENOENT
+	return nil
 }
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -3052,9 +3052,9 @@ func (p *Posix) PutObjectWithPostFunc(ctx context.Context, po s3response.PutObje
 		return s3response.PutObjectOutput{}, s3err.GetAPIError(s3err.ErrKeyTooLong)
 	}
 	if errors.Is(err, syscall.ENOTDIR) {
-		r, parentErr := handleParentDirError(name)
-		if parentErr == s3err.GetAPIError(s3err.ErrObjectParentIsFile) {
-			return r, parentErr
+		parentErr := handleParentDirError(name)
+		if parentErr != nil {
+			return s3response.PutObjectOutput{}, parentErr
 		}
 	}
 	if err != nil && !errors.Is(err, fs.ErrNotExist) {


versity-github:

backend/posix/posix.go
```diff
@@ -3052,7 +3053,18 @@ func (p *Posix) PutObjectWithPostFunc(ctx context.Context, po s3response.PutObje
...
+			return s3response.PutObjectOutput{}, s3err.GetAPIError(s3err.ErrObjectParentIsFile)
+		} else {
+			dir := filepath.Dir(name)
+			d1, err1 := os.Stat(dir)
+			if err1 == nil && !d1.IsDir() {
+				return s3response.PutObjectOutput{}, s3err.GetAPIError(s3err.ErrObjectParentIsFile)
+			} else {
+				err = syscall.ENOENT
+			}
+		}
```
benmcclelland: This logic looks fine to me, but the more go idiomatic way of handling os specific behavior is to pull this out into a small function, and then implement that function the different ways in os specific files with build tags, for example:

dir_windows.go
```go
//go:build windows

func handleDirError(name string) (s3response.PutObjectOutput, error) {
	dir := filepath.Dir(name)
	d, statErr := os.Stat(dir)
	if statErr == nil && !d.IsDir() {
		return s3response.PutObjectOutput{}, s3err.GetAPIError(s3err.ErrObjectParentIsFile)
	}
	// Parent doesn't exist or is a directory, treat as ENOENT
	return s3response.PutObjectOutput{}, syscall.ENOENT
}
```

dir_unix.go
```go
//go:build !windows

func handleDirError(_ string) (s3response.PutObjectOutput, error) {
	return s3response.PutObjectOutput{}, s3err.GetAPIError(s3err.ErrObjectParentIsFile)
}
```
and then call handleDirError() here

khambrecht: That sounds reasonable. Thanks. Updated so that the error handing is now spilt in os specific files.

backend/posix/parentdir_other.go
```diff
@@ -0,0 +1,26 @@
...
+package posix
+
+import (
+	"github.com/versity/versitygw/s3err"
+	"github.com/versity/versitygw/s3response"
+)
+
+func handleParentDirError(_ string) (s3response.PutObjectOutput, error) {
+	return s3response.PutObjectOutput{}, s3err.GetAPIError(s3err.ErrObjectParentIsFile)
+}
```
benmcclelland: this missing newline at the end is failing the gofmt test

backend/posix/parentdir_windows.go
```diff
@@ -0,0 +1,36 @@
...
+
+func handleParentDirError(name string) (s3response.PutObjectOutput, error) {
+	dir := filepath.Dir(name)
+	d, statErr := os.Stat(dir)
+	if statErr == nil && !d.IsDir() {
+		return s3response.PutObjectOutput{}, s3err.GetAPIError(s3err.ErrObjectParentIsFile)
+	}
+	// Parent doesn't exist or is a directory, treat as ENOENT
+	return s3response.PutObjectOutput{}, syscall.ENOENT
+}
```
benmcclelland: this missing newline at the end is failing the gofmt test

khambrecht: Damn ;-)
Now fixed.

benmcclelland:
Still failing some gofmt tests, I think the easiest thing here is just to run `gofmt -w backend/posix/*.go`. This should get all the formatting fixed up.

Also, can you squash these commits to a single commit? We don't need the commit iterations merged into the repo. (I can do this as well when merging if needed).

khambrecht:
I apologize for the inconvenience. This is my first attempt at contributing to a golang project.
And yes, I would be very grateful if you could squash the commits on merging. Thanks.

Apart from the formatting issues, do you agree with the proposed changes?

benmcclelland:
> I apologize for the inconvenience. This is my first attempt at contributing to a golang project. And yes, I would be very grateful if you could squash the commits on merging. Thanks.
> 
> Apart from the formatting issues, do you agree with the proposed changes?

No trouble at all, I'm very happy to have contributors to this project. Yes, I'll do one more pass at this review, but I think the logic looked fine to me.

Thanks for the fix!

benmcclelland:
I think I might see another case that isnt handled correctly here. I think you might need to recursively test all ancestors for existence.

Consider this scenario:
/bucket/a/b is a file (not a directory)
PUT: /bucket/a/b/c/object.txt

filepath.Dir("/bucket/a/b/c/object.txt") returns /bucket/a/b/c, so the current code will stat /bucket/a/b/c 
os.Stat("/bucket/a/b/c") fails with ERROR_PATH_NOT_FOUND/ENOTDIR causing handleParentDirError to return syscall.ENOENT instead of ErrObjectParentIsFile.

Also, minor nit, you can remove the s3response.PutObjectOutput return in handleParentDirError to clean it up since it doesnt do anything in this function.

I think maybe it needs to be something like:
```go
func handleParentDirError(name string) error {
    dir := filepath.Dir(name)
    
    // Walk up the directory hierarchy
    for dir != "." && dir != "/" {
        d, statErr := os.Stat(dir)
        if statErr == nil {
            // Path component exists
            if !d.IsDir() {
                // Found a file in the ancestor path
                return s3err.GetAPIError(s3err.ErrObjectParentIsFile)
            }
            // Found a valid directory ancestor, parent truly doesn't exist
            break
        }
        // Continue checking parent directories
        dir = filepath.Dir(dir)
    }
    
    // Parent doesn't exist or all ancestors are valid directories
    return nil
}
```
non-windows:
```go
func handleParentDirError(_ string) error {
	return s3err.GetAPIError(s3err.ErrObjectParentIsFile)
}
```

and then from the call site, you can check the error is not nil:
```go
	if errors.Is(err, syscall.ENOTDIR) {
		err := handleParentDirError(name)
		if err != nil {
			return s3response.PutObjectOutput{}, err
		}
	}
```

Double check that this is how it should work on windows if you can with that grandparent or great-grandparent is a file test. I haven't tested any of this suggested code.

khambrecht:
Good catch with possible ancestors that could exist as file already. I had already considered this, but did not explicitly take it into account because the upload would have failed as expected anyway. However, with an incorrect error message to the client, namely ExistingObjectIsDirectory instead of the correct ObjectParentIsFile.

I have adopted your suggestion and have updated the PR. Thanks.

khambrecht:
> I think we might need to handle a couple other cases of ancestor is file, but I'm not certain exactly how windows handles these cases.

I did test with several directory/file ancestors combinations and have not yet found any other special case to consider.

benmcclelland:
Great! thanks for working through the testing. I'll get this merged once the tests pass.

PR #1801, opened 2026-01-29
fix: Address slowness with lexical sorting

### LINKED ISSUE #1625: [Bug] - severe performance scaling loss since upgrade to 1.0.18
**Describe the bug**
I created the following directory tree on the posix filesystem:
 :>find /data/TIVGW/ttt/manyfoldertree/ -type d |wc -l
88602
 :>find /data/TIVGW/ttt/manyfoldertree/ -type f |wc -l
41771

As you can see the tree has 88k directories and 41k files distributed around the directories (max nested depth is 15)

Now using versitygw 1.0.14:
:>time vgaws s3 ls ttt/manyfoldertree/ 
                           PRE toplevel/

real    0m0.351s
user    0m0.296s
sys     0m0.047s

same system but with versitygw 1.0.18:

:>time vgaws s3 ls ttt/manyfoldertree/ 
                           PRE toplevel/

real    0m1.764s
user    0m0.282s
sys     0m0.047s

elapsed time is now 1.8s against 0.35s for 1.0.14. 

If I make a clone of the directory inside itself, ie. doubling the number of files and directories:
:>time vgaws s3 ls ttt/manyfoldertree/ 
                           PRE toplevel/

real    0m14.686s
user    0m0.291s
sys     0m0.045s

It now takes 15 seconds. A doubling of input objects has resulted in an 8-fold elapsed time increase for 1.0.18. 1.0.14 still takes only one third of a second to do the same task. Note that although it's a relatively highly populated tree, we are only returning ONE output value: the top level folder. 

I have researched the change and see it is linked to pre-sorting of the entire directory tree in the revised tree walker function. I understand why this was implemented to fix a previous problem, however it is destroying the ability of versitygw to scale to realistic workloads, and will cause front end systems to timeout.

**To Reproduce**
Create a directory with a few thousand files, and clone this directory a few thousand times, then use s3 ls against it.

**Expected behavior**
At worst, linear scaling of elapsed time as the number of objects increases. Ideally no negative impact on listing only the top level just because there are many objects beneath. 

**Server Version**
1.0.18. As stated 1.0.14 is OK (have not tested intervening releases)

**Additional context**
Just use aws cli to see the issue.

wishdev:
I recently hit #1625 and I had the thought (coming from a Ruby background) that one should be able to modify fs.WalkDir to accomplish the task of properly sorting the entries as opposed to rewriting it by hand.

It isn't quite that simple but it's close.

These commits reverts the work of #1515 and #1533 as it concerns walk.go - brings in a cut down version of io/fs/walk.go and modifies the now internal variant of walkDir's entry list to be sorted in an appropriate fashion by adding the missing "/" to directories.

All tests pass and I added a couple of entries to the order test to ensure a brain fart I had when making the sorting change was covered.

Speed is in line with pre #1515 expectations and a clear gain over the #1515 and #1533 work.

Fixes #1625


versity-github:

PR #1802, opened 2026-01-30
fix: posix ListParts optimization

niksis02:
This PR focuses on optimizing **ListParts** in the POSIX backend. Previously, parts were listed using `os.ReadDir` and then sorted. The sorting step was unnecessary because all parts reside under the same directory (`<bucket>/.sgwtmp/multipart/<obj-hash>/<uploadId>/<partNumber>`), and their `partNumber` entries are already returned in sorted order by `os.ReadDir`.

The redundant sorting has now been removed, and an additional check has been introduced to break out of the directory iteration once the `max-parts` limit is reached.

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -2400,7 +2400,9 @@ func (p *Posix) ListParts(ctx context.Context, input *s3.ListPartsInput) (s3resp
 		checksum.Type = types.ChecksumType("null")
 	}
 
-	parts := make([]s3response.Part, 0, len(ents))
+	var nextPartNumber int
+	var isTruncated bool
+	parts := make([]s3response.Part, 0, min(maxParts, len(ents)))
 	for i, e := range ents {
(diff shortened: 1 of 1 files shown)


benmcclelland:
The readdir sorting is lexicographic instead of numeric, for example:
```
Lexicographic: "1", "10", "11", "2", "20", "3"
Numeric: "1", "2", "3", "10", "11", "20"
```
I dont think it is sufficient in this case to rely only on the readdir sort ordering?

niksis02:
> The readdir sorting is lexicographic instead of numeric, for example:
> 
> ```
> Lexicographic: "1", "10", "11", "2", "20", "3"
> Numeric: "1", "2", "3", "10", "11", "20"
> ```
> 
> I dont think it is sufficient in this case to rely only on the readdir sort ordering?

Ahh yes you are right. Closing this one.

PR #1824, opened 2026-02-09
fix: md5 not a required header for PutObjectLockConfiguration

benmcclelland:
This was noticed from an error in the webgui. The md5 header is not required for PutObjectLockConfiguration according to AWS documentation. Fix this so that we do not return an error when the optional header is not provided.

### DIFF
--- a/s3api/router.go
+++ b/s3api/router.go
@@ -216,7 +216,7 @@ func (sa *S3ApiRouter) Init(app *fiber.App, be backend.Backend, iam auth.IAMServ
 			middlewares.AuthorizePublicBucketAccess(be, metrics.ActionPutObjectLockConfiguration, auth.PutBucketObjectLockConfigurationAction, auth.PermissionWrite, region, false),
 			middlewares.VerifyPresignedV4Signature(root, iam, region, false),
 			middlewares.VerifyV4Signature(root, iam, region, false, true),
-			middlewares.VerifyChecksums(false, true, true),
+			middlewares.VerifyChecksums(false, true, false),
 			middlewares.ApplyBucketCORS(be, corsAllowOrigin),
 			middlewares.ParseAcl(be),
 		))


benmcclelland:
Hmmm, the AWS docs make it look like the MD5 header is optional for this case. But if AWS does enforce this, we can add this header to the webgui request instead of making this not required.

niksis02:
Here is the code I used to send the request:
```go
	payload := `
           <ObjectLockConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
              <ObjectLockEnabled>Enbaled</ObjectLockEnabled>
          </ObjectLockConfiguration>`

	signer := v4.NewSigner()

	req, err := http.NewRequest(http.MethodPut, "https://bucket.s3.amazonaws.com?object-lock", bytes.NewReader([]byte(payload)))
	if err != nil {
		log.Fatalf("failed to create the request: %v", err)
	}

	req.Header.Set("x-amz-content-sha256", "UNSIGNED-PAYLOAD")

	err = signer.SignHTTP(context.Background(), aws.Credentials{
		AccessKeyID:     accessKeyId,
		SecretAccessKey: secret,
	}, req, "UNSIGNED-PAYLOAD", service, region, time.Now())
	if err != nil {
		log.Fatalf("failed to sign the request: %v", err)
	}

	client := &http.Client{}
	resp, err := client.Do(req)
	if err != nil {
		log.Fatalf("failed to send the request: %v", err)
	}

	defer resp.Body.Close()
	body, err := io.ReadAll(resp.Body)
	if err != nil {
		log.Fatalf("failed to read the request body: %v", err)
	}

	fmt.Println(string(body))
	for key, val := range resp.Header {
		fmt.Printf("%s: %s\n", key, val[0])
	}
	fmt.Printf("response status: %v\n", resp.StatusCode)
```

benmcclelland:
ok, yeah, looks like this is just incorrect documentation from aws and the md5 header is actually required here.

PR #1826, opened 2026-02-09
feat: update go version to `1.25.7`(latest stable)

niksis02:
Updates the go version from `1.24.1 -> 1.25.7`. It's good to stay up to date with the latest go versions(`1.25.7`, latest stable). It's necessary for fiber `v3` upgrade. Fiber `v3` requires go version `1.25+`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,8 @@
 module github.com/versity/versitygw
 
-go 1.24.0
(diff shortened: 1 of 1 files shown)


niksis02:
The go version update will come alongside with fiber `v3` migration PR.

Issue #1825, opened 2026-02-09
[Feature] - Support anonymous S3 buckets


andriibratanin:
**Describe the solution you'd like**
It is possible to configure a S3 bucket to allow anonymous read-only access:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-bucket-policies.html#example-bucket-policies-anonymous-user

For some reason S3 client software totally ignore this feature by never implementing ability to connect to such buckets (i.e. "access_key" and "secret_key" are always required, but their values make the connection impossible).

VersityGW can obtain additional use case by implementing support of anonymous remote S3 buckets, thus gaining the ability to act as a "proxy" for the software that cannot connect (i.e. VersityGW will expose a "protected" endpoint while redirecting all calls to anonymous S3 bucket)

**Additional context**
Currently it is impossible to connect to an anonymous S3 bucket because VersityGW requires `VGW_S3_ACCESS_KEY` and `VGW_S3_SECRET_KEY` to be set:
```log
2026/02/09 08:39:03 Required flags "access, secret" not set
```

Example Docker compose:
```yaml
services:
  versitygw:
    image: ghcr.io/versity/versitygw:latest
    container_name: versitygw
    environment: 
      - ROOT_ACCESS_KEY=adminkey
      - ROOT_SECRET_KEY=adminpass
      #- VGW_S3_ACCESS_KEY   <-- it is impossible to access a remote S3 bucket without authentication now
      #- VGW_S3_SECRET_KEY   <-- it is impossible to access a remote S3 bucket without authentication now
      - VGW_BACKEND=s3
      - VGW_S3_REGION=ap-northeast-1
      - VGW_S3_ENDPOINT=https://s3-ap-northeast-1.amazonaws.com
    ports:
      - "7070:7070"
    restart: unless-stopped
```

FYI: 
AWS CLI allows anonymous connections by using `--no-sign-request` option:
```powershell
aws --endpoint-url $S3_ENDPOINT s3 ls s3://my-bucket/ --no-sign-request
```

benmcclelland:
You can enable this with public ACLs:
authenticated:
```
aws --endpoint-url $S3_ENDPOINT s3api put-bucket-ownership-controls --bucket my-bucket --ownership-controls 'Rules=[{ObjectOwnership=ObjectWriter}]'
aws --endpoint-url $S3_ENDPOINT s3api put-bucket-acl --bucket my-bucket --acl public-read
```
then unsigned requests should work:
```
[...]

PR #1836, opened 2026-02-11
Anonymous S3 backend

andriibratanin:
This PR fixes https://github.com/versity/versitygw/issues/1825

What was changed:
- "access" and "secret" parameters of S3 backend were made optional
- S3proxy Client now uses `aws.AnonymousCredentials` when both values are missing
- example configuration file updated accordingly

**Warning:**  
If only one value ("access" or "secret") is specified the following error message will be printed in logs:
```log
2026/02/11 09:23:39 init s3 backend: both access and secret must be set or none at all
```
it **replaces** previous error message produced by `urfave/cli` flags validator:
```log
2026/02/11 07:14:09 Required flags "access, secret" not set
```


versity-github:

PR #1837, opened 2026-02-11
feat: fiber migration `v2` -> `v3`

niksis02:
Fiber has released a new major version, **v3.0.0**, introducing several breaking changes. As part of the gateway migration from Fiber v2 to v3, the `filesystem` middleware has been deprecated and replaced with the `static` middleware for serving web assets. Currently, Fiber’s `static` middleware contains a bug in how it resolves the filesystem root path. Because of this, `fs.Sub` is used to set the embedded `web` directory as the effective root of the filesystem.


niksis02:
Another PR has been created for the fiber `v3` migration: https://github.com/versity/versitygw/pull/2182

PR #1843, opened 2026-02-12
feat: add multi-address listener for s3/admin/webui

### LINKED ISSUE #1761: [Bug] - `--port localhost:7070` only uses IPv4
**Describe the bug**
`localhost` usually means both IPv4 and v6.  
It should be all the IP addresses returned by the name resolver, both IPv4/A and v6/AAAA queries.

**To Reproduce**
Run versitygw with `--port localhost:7070` and/or `--admin-port localhost:7070`.

**Expected behavior**
```
Admin/S3 service listening on:
  http://127.0.0.1:7070
  http://[::1]:7070
```

**Server Version**
v1.0.20

**Additional context**
N/A

benmcclelland:
This allows specifying the following options more than once: port, admin-port, webui

or using a comma-separated list for the env vars:
e.g., VGW_PORT=:7070,:8080,localhost:9090

This will also expand multiple interfaces from hostnames, for example "localhost" in this case would resolve to both IPv4 and IPv6 interfaces: localhost has address 127.0.0.1
localhost has IPv6 address ::1

This updates the banner to reflect all of the listening interfaces/ports, and starts the service listener on all requested interfaces/ports.

Fixes #1761

### DIFF
--- a/cmd/versitygw/main.go
+++ b/cmd/versitygw/main.go
@@ -39,7 +39,8 @@ import (
 )
 
 var (
-	port, admPort                          string
+	ports                                  []string
+	admPorts                               []string
 	rootUserAccess                         string
 	rootUserSecret                         string
 	region                                 string
@@ -93,7 +94,7 @@ var (
 	ipaUser, ipaPassword                   string
 	ipaInsecure                            bool
 	iamDebug                               bool
-	webuiAddr                              string
+	webuiPorts                             []string
 	webuiCertFile, webuiKeyFile            string
 	webuiNoTLS                             bool
 )
@@ -144,6 +145,13 @@ VersityGW is an open-source project licensed under the Apache 2.0 License. The
 source code is hosted on GitHub at https://github.com/versity/versitygw, and
 documentation can be found in the GitHub wiki.`,
 		Copyright: "Copyright (c) 2023-2024 Versity Software",
+		Before: func(ctx *cli.Context) error {
+			// Initialize global variables from context (including default values)
+			ports = ctx.StringSlice("port")
+			webuiPorts = ctx.StringSlice("webui")
+			admPorts = ctx.StringSlice("admin-port")
+			return nil
+		},
 		Action: func(ctx *cli.Context) error {
 			return ctx.App.Command("help").Run(ctx)
 		},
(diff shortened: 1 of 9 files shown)


cmd/versitygw/main.go
```diff
@@ -673,15 +683,23 @@ func runGateway(ctx context.Context, be backend.Backend) error {
 		return fmt.Errorf("root user access and secret key must be provided")
 	}
 
-	webuiAddr = strings.TrimSpace(webuiAddr)
-	if webuiAddr != "" && isAllDigits(webuiAddr) {
-		webuiAddr = ":" + webuiAddr
+	// Ensure we have at least one port specified
+	if len(ports) == 0 {
+		ports = []string{":7070"}
+	}
```
niksis02, line 729: Seems the default value is already provided with the CLI flag ?

s3api/utils/multi_listener.go
```diff
@@ -0,0 +1,306 @@
...
+	addrs := make([]string, 0, len(ips))
+	for _, ip := range ips {
+		var addr string
+		if ip.To4() != nil {
+			// IPv4
+			addr = net.JoinHostPort(ip.String(), port)
+		} else {
+			// IPv6
+			addr = net.JoinHostPort(ip.String(), port)
+		}
```
niksis02: The check seems to be unnecessary? Anyways both ip V4 and V6 are converted to string and JoinHostPort is called. I think we don't care the ip version

s3api/utils/multi_listener.go
```diff
@@ -0,0 +1,306 @@
...
+			}
+			return
+		case ml.acceptCh <- acceptResult{conn: conn, err: err}:
+			// Connection or error sent successfully
+			if err != nil {
+				return
+			}
+		}
+	}
+}
```
niksis02, line 85: There's a potential deadlock risk here after `ShutDown`.

`MultiListener` uses an unbuffered `acceptCh` and during shutdown, this can happen:

1. An `acceptLoop` goroutine successfully returns from ln.Accept() with a connection.
2. It tries to forward the result into `acceptCh`.
3. Fiber has already stopped receiving from `MultiListener.Accept()`.
4. The send blocks forever.
5. `Close()` calls `ml.wg.Wait()`, waiting for that goroutine to exit.
6. The goroutine can’t exit because it’s blocked on the channel send.

Result: shutdown hangs (deadlock).
A goroutine blocked on a channel send cannot be interrupted by closing another channel.

When shutdown starts, fiber sets some internal “stopping” state and exits the connection `Accept` loop. Even if it then closes the listener right after, there’s still a tiny window,  where the accept loop has already stopped receiving, but our `acceptLoop` goroutine has just accepted a connection and is trying to forward it.

I guess this can be reproduced by opening some multiple connections and making a large number of concurrent requests to the gateway. While sending the requests, simultaneously shutting down the gateway might lead to a deadlock(haven't tried though).

What I recommend here to avoid the daedlock on ShutDown, is to make `acceptCh` buffered so producers don’t require an immediate receiver during shutdown.
For safety reasons I would recommend using `2xlisteners`.
```
acceptCh: make(chan acceptResult, 2*len(listeners))
```

Each underlying listener has its own acceptLoop goroutine and buffering at least `len(listeners)` guarantees each loop can enqueue one result without blocking. Using `2 * len(listeners)` provides additional safety margin against brief scheduling pauses, GC pauses, transient delays in Fiber’s accept loop ...

extra/example.conf
```diff
@@ -81,7 +87,8 @@ ROOT_SECRET_ACCESS_KEY=
 # control of firewall restrictions to the admin endpoint. The certs for this
 # can be different certs than specified for the S3 service. The default when
 # these are not specified is to have the admin server listen on the same
-# endpoint as the S3 service.
+# endpoint as the S3 service. This can be specify mutliple ports and will
```
niksis02: `This can specify` ? `be` is unnecessary ?

extra/example.conf
```diff
@@ -210,7 +217,8 @@ ROOT_SECRET_ACCESS_KEY=
 # users, buckets and objects. The format can be either ':port' to listen on all
 # interfaces (e.g., ':7071') or 'host:port' to listen on a specific interface
 # (e.g., '127.0.0.1:7071' or 'localhost:7071'). When omitted, the Web GUI is
-# disabled.
+# disabled. This can be specify mutliple ports and will resolve hostnames to
```
niksis02: `This can specify` ? `be` is unnecessary ?

extra/example.conf
```diff
@@ -210,7 +217,8 @@ ROOT_SECRET_ACCESS_KEY=
...
+# mutliple addresses the same as VGW_PORT.
```
niksis02: `*multiple`: Misspelled

niksis02 (approved):

Issue #1815, opened 2026-02-04
[Bug] - Application Crashes from thread exhaustion


MCDELTAT:
**Describe the bug**
A while after starting a container of versitygw, it crashes and produces the following error:
```
fatal error: newosproc
runtime: failed to create new OS thread (have 2053 already; errno=11)
runtime: may need to increase max user processes (ulimit -u)
```

I have attached the full stack trace, as it seems all threads unravel when the application begins to crash.
[versity_crash_2026_02_03.log](https://github.com/user-attachments/files/25059905/versity_crash_2026_02_03.log)

**To Reproduce**
Unknown steps to reproduce, happens regularly after starting the versitygw instance.

**Expected behavior**
The application should release threads gracefully and operate within it's ulimits.

**Server Version**

```
Version  : v1.0.19
Build    : 3c2b4c6452cb8bf3068e0e9c48cb21a8e523bcaa
BuildTime: 2025-11-20T21:54:35Z

Linux 2283bf6ef069 5.14.0-611.16.1.el9_7.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Dec 7 05:52:24 EST 2025 x86_64 Linux
```

App is in the official v1.0.19 container with the digest `df47c7f7c5a1`. I have seen no difference in `v1.0.20` and `v1.1.0`.

[df47c7f7c5a1](https://hub.docker.com/layers/versity/versitygw/v1.0.19/images/sha256-df47c7f7c5a17b47ae946e7bc5608e2970237a197763f914e90600e3f0cc2bf3)

**Additional context**
Client is a proprietary storage client made with the AWS S3 SDK. From the access logs it identifies as aws-sdk-cpp/1.7.211

Per the stack trace, I tried to increase the containers ulimits to `noproc: 4096` and `nofiles: 65536`. There was no change on any of the affected versions.

niksis02:
Hi @MCDELTAT .
Thanks for the bug report.

Would you mind providing your docker setup(the run command and volumes)? I see from the stack trace, that the crash reason is a lot of `os.Rename` syscalls. This is a bug which I'm planning to work on, but I also wonder if you used sidecar metadata storage or not? This might also be a reason to a lot of syscalls, which might lead to a panic.

MCDELTAT:
I'm running it in Podman Compose with the following file, pretty basic.
```yaml
version: "3"
services:
  posix:
    image: versity/versitygw:v1.0.19
    volumes:
      - ./log:/log
[...]

PR #1848, opened 2026-02-13
feat: adds fiber max connections and in-flight requests limiter

niksis02:
This is part of the thread exhaustion issue (#1815).

This PR introduces:

* A **maximum Fiber HTTP connections limit**
* A middleware that enforces a **hard limit on in-flight HTTP requests**

When the in-flight request limit is reached, the middleware returns an **S3-compatible `503 SlowDown`** error.

The same mechanism is implemented for the **admin server** (both max connections and max in-flight requests).

All limits are configurable via **CLI flags** and **environment variables**, for both the `s3api` server and the `admin` server.

---

| Setting         | CLI Flag            | Alias | Environment Variable  | Default |
| --------------- | ------------------- | ----- | --------------------- | ------- |
| Max Connections | `--max-connections` | `-mc` | `VGW_MAX_CONNECTIONS` | 250000  |
| Max Requests    | `--max-requests`    | `-mr` | `VGW_MAX_REQUESTS`    | 100000  |

---

| Setting         | CLI Flag                  | Alias  | Environment Variable        | Default |
| --------------- | ------------------------- | ------ | --------------------------- | ------- |
| Max Connections | `--admin-max-connections` | `-amc` | `VGW_ADMIN_MAX_CONNECTIONS` | 250000  |
| Max Requests    | `--admin-max-requests`    | `-amr` | `VGW_ADMIN_MAX_REQUESTS`    | 100000  |

### DIFF
--- a/cmd/versitygw/main.go
+++ b/cmd/versitygw/main.go
@@ -43,6 +43,8 @@ var (
 	rootUserAccess                         string
 	rootUserSecret                         string
 	region                                 string
+	maxConnections, maxRequests            int
+	adminMaxConnections, adminMaxRequests  int
 	corsAllowOrigin                        string
 	admCertFile, admKeyFile                string
 	certFile, keyFile                      string
@@ -217,6 +219,22 @@ func initFlags() []cli.Flag {
 			Destination: &region,
 			Aliases:     []string{"r"},
 		},
+		&cli.IntFlag{
+			Name:        "max-connections",
+			Usage:       "maximum number of concurrent connections s3 api server may serve",
+			EnvVars:     []string{"VGW_MAX_CONNECTIONS"},
+			Value:       250000,
+			Destination: &maxConnections,
+			Aliases:     []string{"mc"},
+		},
+		&cli.IntFlag{
+			Name:        "max-requests",
+			Usage:       "maximum number of in-flight requests s3 api server may serve",
+			EnvVars:     []string{"VGW_MAX_REQUESTS"},
+			Value:       100000,
+			Destination: &maxRequests,
+			Aliases:     []string{"mr"},
+		},
 		&cli.StringFlag{
 			Name:        "cors-allow-origin",
 			Usage:       "default CORS Access-Control-Allow-Origin value (applied when no bucket CORS configuration exists, and for admin APIs)",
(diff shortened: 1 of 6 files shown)


cmd/versitygw/main.go
```diff
@@ -719,7 +753,9 @@ func runGateway(ctx context.Context, be backend.Backend) error {
 		}()
 	}
```
benmcclelland, line 766: maybe some input validation on these?

```go
    if maxConnections < 1 {
        log.Fatal("max-connections must be positive")
    }
    if maxRequests < 1 {
        log.Fatal("max-requests must be positive")
    }
    if maxRequests > maxConnections {
        log.Printf("WARNING: max-requests (%d) exceeds max-connections (%d) which could allow for gateway to panic before throttling requests", 
                  maxRequests, maxConnections)
    }

```

s3api/middlewares/rate-limiter.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"github.com/gofiber/fiber/v2"
+	"github.com/versity/versitygw/s3err"
+)
+
+// RateLimiter hard-limits the number of in-flight requests.
+// If the limit is reached, an immediate SlowDown error is returned
+func RateLimiter(limit int) fiber.Handler {
+	sem := make(chan struct{}, limit)
```
benmcclelland: This limit is likely to be very large which makes a buffered channel fairly memory intesive, I think we can use golang.org/x/sync/semaphore.Weighted for this instead:

```go
    sem := semaphore.NewWeighted(int64(limit))

    return func(ctx *fiber.Ctx) error {
        if !sem.TryAcquire(1) {
            // limit reached
            err := s3err.GetAPIError(s3err.ErrSlowDown)
            ctx.Status(err.HTTPStatusCode)
            return ctx.Send(s3err.GetAPIErrorResponse(err, "", "", ""))
        }
        defer sem.Release(1)
        return ctx.Next()
    }
```

cmd/versitygw/main.go
```diff
@@ -217,6 +219,22 @@ func initFlags() []cli.Flag {
 			Destination: &region,
 			Aliases:     []string{"r"},
 		},
+		&cli.IntFlag{
+			Name:        "max-connections",
+			Usage:       "maximum number of concurrent connections s3 api server may serve: default 250000",
```
benmcclelland: dont need the "default" in the usage because this automatically gets added in, same below on requests

s3api/middlewares/rate-limiter.go
```diff
@@ -0,0 +1,40 @@
...
+
+package middlewares
+
+import (
+	"github.com/gofiber/fiber/v2"
+	"github.com/versity/versitygw/s3err"
+)
+
+// RateLimiter hard-limits the number of in-flight requests.
+// If the limit is reached, an immediate SlowDown error is returned
```
benmcclelland, line 26: I wonder if we should add some metrics for number of rejected requests so that we can add some observability when a gateway it too heavily loaded.

benmcclelland (approved):

PR #1855, opened 2026-02-16
A faster directory walk algorithm 

tonyipm:
Note that because things are done differently, some of the existing walk tests fail with unexpected results. This is because the walk algorithm sorts and discards as it goes along, which is the main reason it has a time advantage over the current implementation. I have run numerous scenarios to ensure that the final json sent back to the client always matches (using md5 after filtering things like date/time) the results returned from pre this change.


versity-github:

benmcclelland:
> Note that because things are done differently, some of the existing walk tests fail with unexpected results. This is because the walk algorithm sorts and discards as it goes along, which is the main reason it has a time advantage over the current implementation. I have run numerous scenarios to ensure that the final json sent back to the client always matches (using md5 after filtering things like date/time) the results returned from pre this change.

I'll take a look at the tests to see if these are making some bad assumptions. I expected that these should all be testing aws expected results. If not, then we need to update the tests.

benmcclelland:
The test failures look like real errors to me. For example:
```
backend/walk_test.go:262: different delimiter 2: unexpected object, got [photo|s/] wanted []
```

here is the test object listing on aws:
```
% aws s3api list-objects-v2 --bucket <bucket> | grep "Key"                          
            "Key": "photo|s/200|6/Januar|y/sampl|e1.jpg",
            "Key": "photo|s/200|6/Januar|y/sampl|e2.jpg",
            "Key": "photo|s/200|6/Januar|y/sampl|e3.jpg",
```

and the test `different delimiter 2` is `--prefix 'photo|' --delimiter '|'`:
```
% aws s3api list-objects-v2 --bucket <bucket> --prefix 'photo|s/200|' --delimiter '|'
{
    "CommonPrefixes": [
        {
            "Prefix": "photo|s/200|"
        }
    ],
    "RequestCharged": null,
    "Prefix": "photo|"
}
```

Notice how `photo|s/200|` is returned as a common prefix, but no objects listed. But the walk test error is showing that `photo|s/` is getting returned as an object. It looks like this walk is still using the / delimiter because of the directory structure even when alternate delimiter are set.

tonyipm:
Hi Ben, thanks for looking at this. The test error you are seeing is what I was talking about - I don't understand why the test is reporting error. If I create a bucket with those object keys and perform the exact same list-objects-v2, I get the exact output you posted, no additional unexpected key, just the common prefix. I also set a breakpoint at line 4725 of backend/posix/posix.c and I do not see any objects listed in the results array. I did spend some time looking at this test but couldn't work out why it was reporting the failure, other than perhaps it is expecting some different behaviour at the functional level due to the way the results are generated. The point for me is, the ultimate result returned to the S3 client is identical to that returned by version 1.0.18 of versitygw, and matches what you posted as the expected output. See output below: 

DEV13 [/mmfs1/data/TIVGW] :>vgaws s3 mb s3://nt3
make_bucket: nt3
DEV13 [/mmfs1/data/TIVGW] :>vgaws s3api  put-object --bucket nt3  --key photo\|s/200\|6/January\|y/sampl\|e1.jpg
{
    "ChecksumCRC64NVME": "AAAAAAAAAAA=",
    "Size": 0
}
DEV13 [/mmfs1/data/TIVGW] :>vgaws s3api  put-object --bucket nt3  --key photo\|s/200\|6/January\|y/sampl\|e2.jpg
{
    "ChecksumCRC64NVME": "AAAAAAAAAAA=",
    "Size": 0
}
DEV13 [/mmfs1/data/TIVGW] :>vgaws s3api  put-object --bucket nt3  --key photo\|s/200\|6/January\|y/sampl\|e3.jpg
{
    "ChecksumCRC64NVME": "AAAAAAAAAAA=",
    "Size": 0
}
DEV13 [/mmfs1/data/TIVGW] :>vgaws s3api list-objects-v2 --bucket nt3 --prefix 'photo|' --delimiter '|'
{
    "CommonPrefixes": [
        {
            "Prefix": "photo|s/200|"
        }
    ],
    "RequestCharged": null,
    "Prefix": "photo|"
}

benmcclelland:
oh, yes, you are right. I'm seeing the same. I'll look to see if I can tell why the fstest.MapFS is not behaving the same as a normal filesystem. its possible this is not correctly mocking a real filesystem in these cases. I'll keep digging into this, and maybe just remove fstest.MapFS for a local directory test instead.

benmcclelland:
@tonyipm I was able to fix the tests to get them to pass. The issue was how we were tracking directories that were implicitly created vs explicitly uploaded as an object in the mock fs. I'm opening a new PR to replace this one so that I can make the changes: https://github.com/versity/versitygw/pull/1889. Lets continue discussions in that new PR.

PR #1862, opened 2026-02-17
test: REST CORS commands - initial checks

lrm25:


tests/drivers/rest.sh
```diff
@@ -253,11 +253,54 @@ send_rest_go_command() {
   return 0
 }
 
+split_params_by_delimiter() {
```
anodelman: I think that this might be cleaner by having two functions, one to read the callback_params and one for go_params.  Then you can do a `read -a callback_params <<< $(get_callback_params)`.  While you do end up procressing the params twice, the information passing becomes a lot simpler and easier to understand and without having to add additional markers to be processed.

tests/test_rest_put_object.sh
```diff
@@ -517,3 +518,27 @@ export RUN_USERS=true
...
+  bucket_name="$output"
+
+  run get_file_name
+  assert_success
+  test_file=$output
+
+  run setup_bucket_v2 "$bucket_name"
+  assert_success
+
+  payload_content="1234567890"
```
anodelman: Any chance this should be randomly generated?  Also, weird/special characters?

tests/test_xml.sh
```diff
@@ -0,0 +1,90 @@
...
+  assert_failure
+  assert_output -p "No XML declaration found"
+}
+
+@test "get_xml_data - valid XML with declaration" {
+  input=$(get_file_name) output_file=$(get_file_name)
+  printf 'HTTP/1.1 200\r\n\r\n<?xml version="1.0"?><Value>OK</Value>' > "$TEST_FILE_FOLDER/$input"
+  run get_xml_data "$TEST_FILE_FOLDER/$input" "$TEST_FILE_FOLDER/$output_file"
+  assert_success
+  grep -q "<Value>OK</Value>" "$TEST_FILE_FOLDER/$output_file"
```
anodelman: Check output on this call and other grep calls.

anodelman (changes requested):

tests/drivers/rest.sh
```diff
@@ -253,11 +253,49 @@ send_rest_go_command() {
...
+
 send_rest_go_command_callback() {
   if ! check_param_count_gt "response code, callback, params" 2 $#; then
     return 1
   fi
-  if ! rest_go_command_perform_send "${@:3}"; then
+
+  local all_params=("${@:3}") no_callback_params=0 go_params go_param_array=() callback_params=()
+  go_params=$(get_go_params "${all_params[@]}") || no_callback_params=$?
+  while IFS= read -r line; do
```
anodelman, line 289: You can avoid having to loop with readarray `readarray -t go_param_array <<<"$(get_go_params "${all_params[@]})"'

lrm25: Yeah, probably time to upgrade bash

anodelman: Oh, if readarray isn't available than your solution is fine.  Didn't know you wouldn't have it.  :(

lrm25: For now, I can leave it, but I should think about eventually upgrading since I've run into lots of these issues.

PR #1866, opened 2026-02-18
fix: 💄 Responsive WebUI

maxlerebourg:
This PR add responsiveness to the WebUI, only with CSS and Tailwind
image
image

### DIFF
--- a/webui/webserver.go
+++ b/webui/webserver.go
@@ -16,8 +16,8 @@ package webui
 
 import (
 	"fmt"
+	"net"
 	"net/http"
-	"strings"
 
 	"github.com/gofiber/fiber/v2"
 	"github.com/gofiber/fiber/v2/middleware/filesystem"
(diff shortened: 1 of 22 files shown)


versity-github:

webui/webserver.go
```diff
@@ -103,8 +103,8 @@ func (s *Server) setupRoutes() {
 	s.app.Use("/", filesystem.New(filesystem.Config{
 		Root:         http.FS(webFS),
 		PathPrefix:   "web",
-		Index:        "index.html",
-		NotFoundFile: "index.html", // SPA fallback
+		Index:        "web/index.html",
+		NotFoundFile: "web/index.html", // SPA fallback
```
niksis02: The `PathPrefix` seems to take care of path resolution, why do we need to update `Index` and `NotFoundFile` with `web` prefix ?

maxlerebourg: Have you really tested ? When I launch the WebUI and go to http://localhost:8080/test, I have a 404

benmcclelland: we were going to use pathprefix for another issue, we can look into test coverage and verify whats going wrong there

maxlerebourg: This was reverted

benmcclelland:
fixing up a couple minor things in a new PR for this https://github.com/versity/versitygw/pull/1890 and will get the revised one merged. Thanks for this fix!

PR #1875, opened 2026-02-19
feat: configuration option to disable ACLs

### LINKED ISSUE #1847: [Feature] - Configuration option to ignore acl headers.

Configuration option to disable checks added in https://github.com/versity/versitygw/pull/1783. 

**Additional context**

I'm using versitygw in a local test environment which doesn't require actual ACLs, but the library code which I use always sends `x-amz-acl` headers. For that use case it would be fine to ignore this headers.

niksis02:
Closes #1847

This PR introduces a global optional gateway CLI flag `--disable-acl` (`VGW_DISABLE_ACL`) to disable ACL handling. When this flag is enabled, the gateway ignores all ACL-related headers, particularly in `CreateBucket`, `PutObject`, `CopyObject`, and `CreateMultipartUpload`.

`GetBucketAcl` behavior is unchanged simply returning the bucket ACL config. There's no change in object ACL actions(`PutObjectACL`, `GetObjectACL`). They return a`NotImplemented` error as before.

A new custom error is added for PutBucketAcl calls when ACLs are disabled at the gateway level. Its HTTP status code and error code match AWS S3’s behavior, with only a slightly different error message.

In the access-control checker, ACL evaluation is fully bypassed. If ACLs are disabled only the bucket owner gets access to the bucket and all grantee checks are ignored.

The PR also includes minor refactoring of the S3 API server and router. The growing list of parameters passed to the router’s Init method has been consolidated into fields within the router struct, initialized during router construction. Parameters not needed by the S3 server are no longer stored in the server configuration and are instead forwarded directly to the router.


auth/access-control.go
```diff
@@ -79,6 +79,7 @@ type AccessOptions struct {
 	Action          Action
 	Readonly        bool
 	IsPublicRequest bool
+	disableACL      bool
```
benmcclelland: does this need to be exported?

s3api/controllers/bucket-put.go
```diff
@@ -480,11 +489,11 @@ func (c S3ApiController) PutBucketAcl(ctx *fiber.Ctx) (*Response, error) {
...
-	grantFullControl := ctx.Get("X-Amz-Grant-Full-Control")
-	grantRead := ctx.Get("X-Amz-Grant-Read")
-	grantReadACP := ctx.Get("X-Amz-Grant-Read-Acp")
-	grantWrite := ctx.Get("X-Amz-Grant-Write")
+	acl := types.BucketCannedACL(c.getAclHeaderValue(ctx, "X-Amz-Acl"))
+	grantFullControl := c.getAclHeaderValue(ctx, "X-Amz-Grant-Full-Control")
+	grantRead := c.getAclHeaderValue(ctx, "X-Amz-Grant-Read")
+	grantReadACP := c.getAclHeaderValue(ctx, "X-Amz-Grant-Read-Acp")
+	grantWrite := c.getAclHeaderValue(ctx, "X-Amz-Grant-Write")
 	grantWriteACP := ctx.Get("X-Amz-Grant-Write-Acp")
```
benmcclelland: missed getAclHeaderValue?

s3api/server.go
```diff
@@ -42,18 +42,16 @@ const (
...
-	maxConnections  int
-	maxRequests     int
+	Router         *S3ApiRouter
+	app            *fiber.App
+	backend        backend.Backend
+	CertStorage    *utils.CertStorage
+	quiet          bool
+	keepAlive      bool
+	health         string
+	virtualDomain  string
```
benmcclelland: appears this is now in Router.virtualDomain, so no longer used

s3api/controllers/bucket-put.go
```diff
@@ -353,6 +353,15 @@ func (c S3ApiController) PutBucketAcl(ctx *fiber.Ctx) (*Response, error) {
 		}, err
 	}
 
+	if c.disableACL {
```
benmcclelland, line 363: is this needed? looks like auth.ParseACL has already run by this point?

cmd/versitygw/main.go
```diff
@@ -326,6 +327,13 @@ func initFlags() []cli.Flag {
 			Destination: &virtualDomain,
 			Aliases:     []string{"vd"},
 		},
+		&cli.BoolFlag{
+			Name:        "disable-acl",
+			Usage:       "disables gateway ACLs, by ignoring all ACL headers",
+			EnvVars:     []string{"VGW_DISABLE_ACL"},
+			Destination: &disableACLs,
+			Aliases:     []string{"da"},
```
benmcclelland: maybe `noacl` would be clearer than `da`?

benmcclelland (changes requested):
Not usre if we need to add any tests for this? Also its maybe surprising that GetBucketAcl still returns ACLs with this disabled. I think that is probably ok, but I wonder if the cli option should note this behavior? I have been using extra/example.conf for some longer descriptions, and then updating the wiki accordingly. This is optional for this PR though, I can update extra/example.conf in a new PR if needed too.

s3api/controllers/bucket-put.go
niksis02: `auth.ParseACL` is run in `middlewares.ParseACL` for all routes to parse buckets ACL which is necessary to retreive bucket owner(for metrics, s3 event, audit logs ...), bucket ACL for access control ... . It doesn't return any error even when ACLs are disabled

s3api/server.go
niksis02: it's also used in server to initialize the `middlewares.HostStyleParser` global middleware.

niksis02:
> Not usre if we need to add any tests for this? Also its maybe surprising that GetBucketAcl still returns ACLs with this disabled. I think that is probably ok, but I wonder if the cli option should note this behavior? I have been using extra/example.conf for some longer descriptions, and then updating the wiki accordingly. This is optional for this PR though, I can update extra/example.conf in a new PR if needed too.

Do you mean integration or unit testing ?

s3api/server.go
benmcclelland: Can't you just use the one in Router for this? I think its confusing to have it in two places when one isnt really used. And it looks like there is a bug with WithHostStyle() setting the wrong one (to confirm the previous concern).
```diff
--- a/s3api/server.go
+++ b/s3api/server.go
@@ -49,7 +49,6 @@ type S3ApiServer struct {
        quiet          bool
        keepAlive      bool
        health         string
-       virtualDomain  string
        maxConnections int
        maxRequests    int
 }
@@ -128,8 +127,8 @@ func New(
        app.Use(controllers.WrapMiddleware(middlewares.DecodeURL, l, mm))
 
        // initialize host-style parser in virtual domain is specified
-       if server.virtualDomain != "" {
-               app.Use(middlewares.HostStyleParser(server.virtualDomain))
+       if server.Router.virtualDomain != "" {
+               app.Use(middlewares.HostStyleParser(server.Router.virtualDomain))
        }
 
        // initialize the debug logger in debug mode
@@ -172,7 +171,7 @@ func WithReadOnly() Option {
 // WithHostStyle enabled host-style bucket addressing on the server
 func WithHostStyle(virtualDomain string) Option {
        return func(s *S3ApiServer) {
-               s.virtualDomain = virtualDomain
+               s.Router.virtualDomain = virtualDomain
                s.Router.virtualDomain = virtualDomain
        }
 }
```

benmcclelland:
> > Not usre if we need to add any tests for this? Also its maybe surprising that GetBucketAcl still returns ACLs with this disabled. I think that is probably ok, but I wonder if the cli option should note this behavior? I have been using extra/example.conf for some longer descriptions, and then updating the wiki accordingly. This is optional for this PR though, I can update extra/example.conf in a new PR if needed too.
> 
> Do you mean integration or unit testing ?

yeah, but only if you think we need to test the behavior. the original concern was that a client setting object ACL was getting an error. maybe we need a test to make sure this no longer errors with this setting set?

benmcclelland (approved):

PR #1887, opened 2026-02-23
fix: store final checksum on CompleteMultipartUpload

niksis02:
Previously, if no object checksum type/algorithm was specified when initiating a multipart upload, the CompleteMultipartUpload request would compute the final object’s CRC64NVME checksum but not persist it. This logic has now been fixed, and in the scenario described above the checksum is stored on the final object. There should no longer be any case where a CompleteMultipartUpload request finishes without persisting the final object checksum.

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -1661,9 +1661,8 @@ func (p *Posix) CompleteMultipartUploadWithCopy(ctx context.Context, input *s3.C
 	// The checksum algorithm should default to CRC64NVME
 	// just for data integrity. It isn't going to be saved
 	// in the final object metadata
-	checksumAlgorithm := types.ChecksumAlgorithmCrc64nvme
-	if checksums.Algorithm != "" {
-		checksumAlgorithm = checksums.Algorithm
+	if checksums.Algorithm == "" {
+		checksums.Algorithm = types.ChecksumAlgorithmCrc64nvme
 	}
 
 	// ChecksumType should be the same as specified on CreateMultipartUpload
@@ -1676,13 +1675,11 @@ func (p *Posix) CompleteMultipartUploadWithCopy(ctx context.Context, input *s3.C
 		return res, "", s3err.GetChecksumTypeMismatchOnMpErr(checksumType)
 	}
 
+	mpChecksumType := checksums.Type
+
 	// The checksum type should default to FULL_OBJECT(crc64nvme)
-	checksumType := checksums.Type
 	if checksums.Type == "" {
-		// do not modify checksums.Type to further not save the checksum
-		// in the final object. As if no checksum has been specified on mp
-		// creation, the final object shouldn't contain the checksum metadata
-		checksumType = types.ChecksumTypeFullObject
+		checksums.Type = types.ChecksumTypeFullObject
 	}
 
 	// check all parts ok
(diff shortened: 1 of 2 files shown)


backend/posix/posix.go
```diff
@@ -1676,23 +1675,21 @@ func (p *Posix) CompleteMultipartUploadWithCopy(ctx context.Context, input *s3.C
 		return res, "", s3err.GetChecksumTypeMismatchOnMpErr(checksumType)
 	}
 
+	mpChecksumType := checksums.Type
```
benmcclelland, line 1678: move this below the blokc below this to line 1684 so that it gets intialized correctly if we set the default type

niksis02: `mpChecksumType` should store the checksum algorithm specified when the multipart upload was initiated. It must retain this original value so that, after computing the final object checksum, it can be compared with the checksum provided in the `CompleteMultipartUpload` request. If no checksum type/algorithm was originally provided, all checksum headers in the Complete request must be ignored (S3 behavior), and the final object checksum should be computed using the default `crc64nvme`.

benmcclelland: ah ok, then the bug is in the switch statement below if this is intended to remain an empty string.

backend/posix/posix.go
```diff
@@ -1676,23 +1675,21 @@ func (p *Posix) CompleteMultipartUploadWithCopy(ctx context.Context, input *s3.C
...
+		checksums.Type = types.ChecksumTypeFullObject
 	}
 
 	// check all parts ok
 	last := len(parts) - 1
 	var totalsize int64
 
 	var composableCRC bool
-	switch checksums.Type {
+	switch mpChecksumType {
```
benmcclelland, line 1690: this should be:
```go
switch checksums.Type {
```
so that we get the override when the incoming request is ""

benmcclelland: I guess I should put the whole problem statement in here. This is what I was chasing down with the comment on the other section of code:

Because mpChecksumType is "", composableCRC stays false. This means even though checksums.Algorithm is CRC64NVME (which is composable per IsChecksumComposable) and checksums.Type is now FullObject, the code takes the hashRdr path instead of the composable path.

niksis02: This is also intentional, becuase if no checksum has been specified on multipart upload initialization per-part checksum is only calculated on `UploadPart` and not stored, because `ListParts` doesn't include the calulcated crc64nvme(default) part checksum and the final complete request shoudln't contain per part checksum in the payload

niksis02: The composble checksum path won't simply work in this case, because parts chcksums are not stored

benmcclelland: that seems like a problem? this is going to half the performance of multipart uploads.

benmcclelland: why cant we just store the checksum as CRC64NVME?

niksis02: We can't because when no checksum algorithm is provided on mp initiation, each UploadPart may use different checksums. Otherwise we should calculate the crc64nvme for all uploaded parts just to store in xattrs for further complete request calculation. Effectively it's the same effort splitted between the parts upload. We only improve the performance when crc64nvme is used on parts upload. On the other hand if for example sha1 is used, we will have double checksum calculation on part uploads

niksis02: Maybe we should mention in wiki that we recommend posix users to use a certian checksum algorithm on mp creation ?

benmcclelland: we are going to keep iterating on the mpu checksum in a new PR, resolving this for now and we can work on efficiency in another PR

PR #1889, opened 2026-02-23
feat: implement quicker backend/posix walk algorithm

### LINKED ISSUE #1845: [Bug] - <Short Description>
**Describe the bug**
The latest update (#1801) to directory walking in versity gateway produces incorrect results as compared to AWS s3 and the earlier versity code

**To Reproduce**

set up a bucket:
aws s3 mb s3://titest3

add objects as follows:
aws s3api put-object --bucket titest3 --key f3
aws s3api put-object --bucket titest3 --key f4
aws s3api put-object --bucket titest3 --key f5
aws s3api put-object --bucket titest3 --key f6

now try this list-objects-v2 request:
vgaws s3api list-objects-v2 --bucket titest3 --delimiter 5 --start-after f3

If you try this with AWS or with versity before the latest change you see:
{
    "Contents": [
        {
            "Key": "f4",
            "LastModified": "2026-02-12T13:55:50+00:00",
            "ChecksumAlgorithm": [
                "CRC64NVME"
            ],
            "Size": 0,
            "StorageClass": "STANDARD"
        },
        {
            "Key": "f6",
            "LastModified": "2026-02-12T13:55:53+00:00",
            "ChecksumAlgorithm": [
                "CRC64NVME"
            ],
            "Size": 0,
            "StorageClass": "STANDARD"
        }
    ],
    "CommonPrefixes": [
        {
            "Prefix": "f5"
        }
    ],
    "RequestCharged": null,
    "Prefix": ""
}

That is, objects f4 and f6 are returned, along with common prefix f5 because it matches the --delimiter 5 argument

However if you try this with latest (post-#1801) versity, the common prefix is not seen. 
This can impact listing of large buckets where multiple calls are required with start-after arguments. 

**Expected behavior**
I expect to see the same behaviour as AWS S3, or pre-1801 versity

**Server Version**
post the #1801 update

**Additional context**
I recently opened enhancement request #1844 offering my own walk algorithm that does not suffer from this bug, and also runs significantly faster on large buckets. This algorithm can be seen here: https://github.com/tonyipm/versitygw_quickerwalk

benmcclelland:
Rewrite the posix Walk implementation to avoid the extra ReadDir per directory that was noted as a TODO in the old code.  The new algorithm holds all traversal state in a walkState struct and uses processDir to interleave sibling entries in correct S3 lexicographic order without a second syscall.

Key changes:
prefix optimisation: jump directly into the deepest matching directory rather than scanning from the root on every call

marker short-circuit: skip entire subtrees that are lexically before the marker, making paginated listing faster

Fixes #1845
Fixes #1855

### DIFF
--- a/backend/walk.go
+++ b/backend/walk.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"path"
 	"slices"
 	"strings"
 	"syscall"
@@ -27,11 +28,36 @@ import (
 	"github.com/versity/versitygw/s3response"
 )
 
+const (
+	pathSeparator = "/"
+	pathDot       = "."
+)
+
 type WalkResults struct {
 	CommonPrefixes []types.CommonPrefix
 	Objects        []s3response.Object
 	Truncated      bool
 	NextMarker     string
+}
+
+// walkState holds the state needed during directory traversal
+type walkState struct {
+	ctx        context.Context
+	fileSystem fs.FS
+	prefix     string
+	delimiter  string
+	marker     string
+	max        int32
+	getObj     GetObjFunc
+	skipdirs   []string
+
+	// Mutable state
+	cpmap     cpMap
+	objects   []s3response.Object
+	pastMax   bool
+	newMarker string
+	truncated bool
+	walkErr   error
 }
 
 type GetObjFunc func(path string, d fs.DirEntry) (s3response.Object, error)
@@ -69,9 +95,6 @@ func (c cpMap) CpArray() []types.CommonPrefix {
 // Walk walks the supplied fs.FS and returns results compatible with list
 // objects responses
 func Walk(ctx context.Context, fileSystem fs.FS, prefix, delimiter, marker string, max int32, getObj GetObjFunc, skipdirs []string) (WalkResults, error) {
-	cpmap := cpMap{}
-	var objects []s3response.Object
-
 	// if max is 0, it should return empty non-truncated result
 	if max == 0 {
 		return WalkResults{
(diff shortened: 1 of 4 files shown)


benmcclelland:
@tonyipm I added a benchmark test in here, and got the following results (but didnt run many times, so these may not be statistically significant):

Benchmark | main (old) | new | Δ
-- | -- | -- | --
WalkFlat | 11.26 ms | 12.12 ms | +7.6% slower
WalkFlatDelim | 11.10 ms | 12.06 ms | +8.6% slower
WalkFlatPaged | 22.94 ms | 26.99 ms | +17.6% slower
WalkDeep | 13.21 ms | 12.34 ms | -6.6% faster
WalkDeepDelim | 87.2 µs | 70.5 µs | -19.1% faster
WalkWide | 28.68 ms | 27.71 ms | -3.4% faster
WalkWideDelim | 154.3 µs | 127.1 µs | -17.6% faster
WalkWithPrefix | 220.2 µs | 269.6 µs | +22.4% slower
WalkWithPrefixDelim | 38.37 µs | 33.95 µs | -11.5% faster
WalkWithMarker | 1.384 ms | 1.029 ms | -25.6% faster
geomean | 1.741 ms | 1.676 ms | -3.7%

Can you try running the tests, or adding to the benchmarks to see if this really is speeding up the cases in question. Also, we should consider if this is slowing down any cases that we might need to address.

benmcclelland:
tested again with some larger directories, these are relative to current main:

Benchmark | main | new | Δ | p-value
-- | -- | -- | -- | --
WalkFlatPaged | 25.0 ms | 23.4 ms | -6.6% | p=0.004 ✓
WalkDeepDelim | 95.9 µs | 77.6 µs | -19.0% | p=0.002 ✓
WalkWideDelim | 171 µs | 135 µs | -21.2% | p=0.002 ✓
WalkWithPrefix | 256 µs | 289 µs | +12.7% | p=0.004 ✓
WalkWithPrefixDelim | 46.7 µs | 39.0 µs | -16.6% | p=0.004 ✓
WalkWithMarker | 1.64 ms | 1.08 ms | -34.4% | p=0.041 ✓
WalkFlat | 12.6 ms | 11.9 ms | ~ | p=0.394 (noise)
WalkFlatDelim | 12.0 ms | 12.0 ms | ~ | p=0.699 (noise)
WalkDeep | 14.1 ms | 13.4 ms | ~ | p=0.485 (noise)
WalkWide | 32.1 ms | 31.4 ms | ~ | p=0.818 (noise)
geomean | 1.95 ms | 1.74 ms | -10.8% |

tonyipm:
Thanks Ben I will take a look - in my own performance testing I found it consistently faster than the old algo and where performance was similar I ran a profile and found nearly all the elapsed time was in os.ReadDir. I will see if I can run up your perf tests and profile them.

benmcclelland:
> Thanks Ben I will take a look - in my own performance testing I found it consistently faster than the old algo and where performance was similar I ran a profile and found nearly all the elapsed time was in os.ReadDir. I will see if I can run up your perf tests and profile them.

I didn't do extensive testing, but just a few cases of "large" directories vs fan out into multiple directories. I should probably fire up some tests with really large directories since thats the real case we are trying to solve here. Under a million files seems too fast on my system for any useful analysis.

benmcclelland:
@tonyipm I wanted to make sure to revisit this PR. I think the speed up in the WalkFlatPaged  and WalkWithMarker test may be worth landing this. I was curious if you had done any more testing and have any further results?

tonyipm:
Thanks for the update Ben. I have done a lot of performance testing with this for our product workload, concentrating on the wide and deep directory structures that our customers have, and I find it consistently faster for those workloads. These cases have run times in seconds and minutes rather than ms and microseconds, and the improvement gets significantly greater the wider / deeper the directory. Thanks for bringing this into the main product code, this will make things a lot easier for us!

benmcclelland:
I found another regression that i added a fix for in this. I need to do a little more testing to make sure nothing else missed here.

benmcclelland:
@tonyipm I think I have all the issues resolved that I could find in this latest version now. Do you want to do a final review of these changes?

Also, I added a bunch of tests based on my regression and new issue testing.

PR #1903, opened 2026-02-27
test: CORS - response header tests

lrm25:


tests/drivers/rest.sh
```diff
@@ -318,16 +318,33 @@ send_rest_go_command_callback() {
...
 }
 
+# return 0 for key match, 1 for no key match, 2 for value mismatch
+check_key_and_value_pair_for_match() {
+  if ! check_param_count_v2 "read key, read value, expected key, expected value" 4 $#; then
+    return 2
+  fi
+  read_lowercase=$(printf '%s' "$1" | tr '[:upper:]' '[:lower:]')
+  expected_lowercase=$(printf '%s' "$3" | tr '[:upper:]' '[:lower:]')
+  if [ "$read_lowercase" == "$expected_lowercase" ]; then
```
anodelman: Can do this comparison in one step:  

if echo "$string1" | grep -iq "^$string2$"; then
    echo "Match found"
fi

or, if you have an advanced enough bash

if [[ "${string1,,}" == "${string2,,}" ]]; then
    echo "The strings match (case-insensitive)"
fi

tests/drivers/rest.sh
```diff
@@ -446,3 +463,177 @@ send_rest_go_command_write_response_to_file() {
...
+
+    local read_key_lc
+    read_key_lc=$(printf '%s' "$key" | tr '[:upper:]' '[:lower:]')
+
+    local idx
+    for ((idx=0; idx<${#expected_pairs[@]}; idx+=2)); do
+      local exp_key="${expected_pairs[$idx]}"
+      local exp_val="${expected_pairs[$((idx+1))]}"
+      local exp_key_lc
+      exp_key_lc=$(printf '%s' "$exp_key" | tr '[:upper:]' '[:lower:]')
```
anodelman: Again, i think that we can do this more efficiently than through tr.  Please look into other means of doing direct case insensitive comparison.

anodelman (changes requested):
Some changes to how to do case insensitive comparison more efficiently.

tests/rest_scripts/command/s3Command.go
```diff
@@ -41,7 +41,7 @@ const (
 
 const SHA256HashZeroBytes = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
 
-type S3RESTCommand struct {
+/*type S3RESTCommand struct {
```
anodelman: If you don't need this anymore just remove it instead of commenting out.  It will remain in the git history if you need it back later.

anodelman (approved):

PR #1920, opened 2026-03-03
fix: change the way object metadata is stored in posix

### LINKED ISSUE #1909: [Bug] - `InternalError` - object metadata long keys
**Describe the bug**
Object metadata in posix is stored in file or directory extended attributes using the following format:
`user.x-amz-meta.<key>: <value>`.

Almost all the filesystems we support — such as ext4, ZFS, XFS, and Btrfs ... — impose size limits on xattr keys and values. Typical limits are:

* **Key:** up to **255 bytes**
* **Value:** up to **64 KB**

On **APFS** (macOS), the key size limit is even smaller, at **127 bytes**.

In contrast, **Amazon S3 imposes no per-key or per-value size limit** on user metadata. S3 only applies a **combined metadata size limit of 2 KB**, which is already tracked under bug report **#1606**.

As a result, when a client uploads an object containing **metadata keys that exceed filesystem xattr limits**, the POSIX layer fails during attribute assignment and returns an internal error. Example debug log:

```
[INTERNAL ERROR]: set user attr "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa": xattr.FSet test/.sgwtmp/2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824.4270512995 user.X-Amz-Meta.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa: file name too long
```

**To Reproduce**
```bash

aws s3api put-object --bucket test --key obj --metadata "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa=val"
```

**Expected behavior**
The object upload should succeed

**Server Version**
Darwin mac 24.6.0 Darwin Kernel Version 24.6.0: Wed Nov  5 21:32:34 PST 2025; root:xnu-11417.140.69.705.2~1/RELEASE_ARM64_T6020 arm64

niksis02:
Fixes #1909

Previously, the mapping between object metadata and posix object was as follows: for each metadata key, we stored a separate xattr with the `user.X-Amz-Meta.<key>` prefix. This resulted in syscall overhead when storing and deleting large numbers of metadata keys. In addition, very long metadata keys caused failures because most posix filesystems limit xattr key lengths to 127–255 bytes, while S3 does not enforce such a per-key limit.

The logic has now been changed so that all object metadata is stored in a single xattr, `user.metadata`, as a JSON key/value object. For backward compatibility, metadata GET operations still fall back to the old mechanism (`metadata key -> xattr key`) when `user.metadata` is not present.

A new CLI utility has been added to convert all legacy object metadata to the new metadata format within the provided directory.

**Example usage:**

```
versitygw utils convert-xattr-metadata path/to/bucket
```

or

```
versitygw utils cxm path/to/bucket
```

It is recommended to run this command on bucket directories to convert all legacy metadata for every object in the bucket.


backend/posix/posix.go
```diff
@@ -2182,10 +2183,29 @@ func (p *Posix) loadObjectMetaData(f *os.File, bucket, object string, fi *os.Fil
...
+
 	return result
 }
 
-func (p *Posix) storeObjectMetadata(f *os.File, bucket, object string, m objectMetadata) error {
+// storeObjectMetadata parses the given metadata map[string]string
+// to raw json and stores in file/directory xattrs
+func (p *Posix) storeObjectMetadata(f *os.File, bucket, object string, meta map[string]string) error {
+	if len(meta) == 0 {
+		return nil
```
benmcclelland, line 2211: do we need to delete all previously set metadata here?

backend/posix/posix.go
```diff
@@ -4623,20 +4631,9 @@ func (p *Posix) CopyObject(ctx context.Context, input s3response.CopyObjectInput
...
-		// Store the new metadata
-		for k, v := range input.Metadata {
-			err := p.meta.StoreAttribute(nil, dstBucket, dstObject,
-				fmt.Sprintf("%v.%v", metaHdr, k), []byte(v))
-			if err != nil {
-				return s3response.CopyObjectOutput{}, fmt.Errorf("set user attr %q: %w", k, err)
-			}
+		err = p.meta.DeleteAttribute(dstBucket, dstObject, metadataHdr)
+		if err != nil && !errors.Is(err, meta.ErrNoSuchKey) {
+			return s3response.CopyObjectOutput{}, fmt.Errorf("delete object metadata: %s", err)
```
benmcclelland: use %w for error wrapping type

backend/posix/posix.go
```diff
@@ -2109,41 +2092,59 @@ func (p *Posix) checkUploadIDExists(bucket, object, uploadID string) ([32]byte,
...
-			b, err := p.meta.RetrieveAttribute(f, bucket, object, e)
-			if err != nil {
-				continue
+// loadObjectMetadata loads the given object metadata, if it fails to load
+// it falls back to the old metadata key -> xattr key mapping mechansim
+func (p *Posix) loadObjectMetadata(f *os.File, bucket, object string) map[string]string {
+	m, err := p.meta.RetrieveAttribute(f, bucket, object, metadataHdr)
+	if err != nil {
+		if errors.Is(err, meta.ErrNoSuchKey) {
+			// fallback to the deprecated mechanism
```
benmcclelland, line 2112: i wonder if these should be migrated to the new format when read? we might also need a cli utility to run across a directory converting the metadata format for all files?

backend/posix/posix.go
```diff
@@ -2109,41 +2092,59 @@ func (p *Posix) checkUploadIDExists(bucket, object, uploadID string) ([32]byte,
...
-	if m != nil {
-		for _, e := range ents {
-			if !isValidMeta(e) {
-				continue
-			}
-			b, err := p.meta.RetrieveAttribute(f, bucket, object, e)
-			if err != nil {
-				continue
+// loadObjectMetadata loads the given object metadata, if it fails to load
+// it falls back to the old metadata key -> xattr key mapping mechansim
```
benmcclelland: mechanism

PR #1922, opened 2026-03-03
test: some util file updates, changes, removals

lrm25:


tests/drivers/list_parts/list_parts_rest.sh
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env bash
+
+# Copyright 2024 Versity Software
```
anodelman: 2026

tests/drivers/list_parts/list_parts_rest.sh
```diff
@@ -0,0 +1,130 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+check_part_list_rest_etags() {
+  if ! check_param_count_gt "data file name, etags" 2 $#; then
+    return 1
+  fi
+  if ! etags=$(xmllint --xpath '//*[local-name()="ETag"]/text()' "$1" | tr '\n' ' ' 2>&1); then
```
anodelman: Just a note on this that since the output redirect happens after the pipe it will only absorb errors from tr, which errors very rarely.  You may want to move that before the pipe to stop data leaking into the output.

anodelman (changes requested):

Issue #980, opened 2024-12-10
[Feature] - HELM Chart


ElGigi:
My request is for the creation of an official HELM Chart that would allow the gateway to be deployed in Kubernetes.

Example of options:

```yaml
replicas: 3

volumes:
  - name: vol
    hostPath:
      path: /any/path/it/will/be/replaced

versitygw:
    # Path of backend directory
    path: "/any/path/it/will/be/replaced"
    # Secret name for access token
    access_token: ""
    # Secret name for secret key
    secret_key: ""
```

jacksgt:
Hi @ElGigi , if you are still interested, I put together a Helm chart, please take a look at it: https://github.com/versity/versitygw/pull/1924

ElGigi:
It looks good, thank you for the work.

shoce:
is there a chance to have extraVolumes and extraVolumeMounts OR persistence.hostPath so we can use a hostPath for persistence?

jacksgt:
Certainly! Would you like to contribute a PR for it?

shoce:
> Certainly! Would you like to contribute a PR for it?

Sure, here it is https://github.com/versity/versitygw/pull/1964

PR #1924, opened 2026-03-04
feat: add Helm chart for versitygw

jacksgt:
Hi everyone!

I'm really excited about Versity GW, looks like a perfect replacement for my Minio setup. 

Since I'm deploying Versity GW on Kubernetes, I put together a Helm chart (with the most common configuration settings and options) - I thought I'd share it here for the benefit of everyone.

I also added a GitHub action that will automatically release a new version of the chart (i.e. upload a new OCI image to GHCR) whenever the `version` in the `Chart.yaml` file is updated.

Closes https://github.com/versity/versitygw/issues/980

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -1,0 +1,211 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "versitygw.fullname" . }}
+  labels:
+    {{- include "versitygw.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      {{- include "versitygw.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      annotations:
+        {{- if not .Values.auth.existingSecret }}
+        checksum/credentials: {{ cat .Release.Name .Values.auth.accessKey .Values.auth.secretKey | sha256sum | trunc 12 }}
+        {{- end }}
+      {{- with .Values.podAnnotations }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "versitygw.selectorLabels" . | nindent 8 }}
+        {{- with .Values.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+    spec:
+      {{- with .Values.image.pullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "versitygw.serviceAccountName" . }}
+      automountServiceAccountToken: false
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          env:
+            # Backend selection — consumed by docker-entrypoint.sh
+            - name: VGW_BACKEND
+              value: {{ .Values.gateway.backend.type | quote }}
+            - name: VGW_BACKEND_ARGS
+              value: {{ .Values.gateway.backend.args | quote }}
+            # Root credentials — sourced from a Kubernetes Secret
+            - name: ROOT_ACCESS_KEY_ID
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "versitygw.credentialsSecretName" . }}
+                  key: rootAccessKeyId
+            - name: ROOT_SECRET_ACCESS_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "versitygw.credentialsSecretName" . }}
+                  key: rootSecretAccessKey
+            # Gateway server settings
+            - name: VGW_PORT
+              value: ":{{ .Values.gateway.port }}"
+            - name: VGW_REGION
+              value: {{ .Values.gateway.region | quote }}
+            - name: VGW_HEALTH
+              value: "/_/health"
+            {{- if .Values.gateway.virtualDomain }}
+            - name: VGW_VIRTUAL_DOMAIN
+              value: {{ .Values.gateway.virtualDomain | quote }}
+            {{- end }}
+            {{- if .Values.gateway.readOnly }}
+            - name: VGW_READ_ONLY
+              value: "true"
+            {{- end }}
+            {{- if .Values.gateway.quiet }}
+            - name: VGW_QUIET
+              value: "true"
+            {{- end }}
+            {{- if .Values.gateway.debug }}
+            - name: VGW_DEBUG
+              value: "true"
+            {{- end }}
+            {{- if .Values.gateway.accessLog }}
+            - name: VGW_ACCESS_LOG
+              value: {{ .Values.gateway.accessLog | quote }}
+            {{- end }}
+            {{- if .Values.gateway.corsAllowOrigin }}
+            - name: VGW_CORS_ALLOW_ORIGIN
+              value: {{ .Values.gateway.corsAllowOrigin | quote }}
+            {{- end }}
+            {{- if .Values.gateway.disableStrictBucketNames }}
+            - name: VGW_DISABLE_STRICT_BUCKET_NAMES
+              value: "true"
+            {{- end }}
+            # Admin API server
+            {{- if .Values.admin.enabled }}
+            - name: VGW_ADMIN_PORT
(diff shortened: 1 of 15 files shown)


versity-github:

benmcclelland:
Oh wow, thank you for putting this together! I'll check it out and see if we can get this merged.

chart/templates/deployment.yaml
```diff
@@ -0,0 +1,208 @@
...
+            - name: ROOT_ACCESS_KEY_ID
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "versitygw.credentialsSecretName" . }}
+                  key: rootAccessKeyId
+            - name: ROOT_SECRET_ACCESS_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "versitygw.credentialsSecretName" . }}
+                  key: rootSecretAccessKey
```
crabique, line 55: In cases where these secrets are defined in `values.yaml` as plaintext, changing only them and redeploying the chart would not actually trigger a deployment restart (unless, of course, there's Reloader controller also installed in the cluster and the corresponding annotations are added).

Such scenario could be covered by a hash annotation on the pod template, something like this would work (salted with the release name just in case):
```
{{ cat .Release.Name .Values.auth.accessKey .Values.auth.secretKey | sha256sum | trunc 12 }}
```

Any change to the pod template will trigger a rollout restart for the deployment.

jacksgt: That's an excellent point, and one I had not considered before because almost all config parameters are passed via environment variables directly, where this is not an issue.

Thanks for the suggestion, I have updated the PR.

chart/templates/deployment.yaml
```diff
@@ -0,0 +1,211 @@
...
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "versitygw.serviceAccountName" . }}
+      automountServiceAccountToken: false
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
```
crabique, line 37: Sorry for bike-shedding 🙂

Wouldn't it make sense to rely on `| default .Chart.AppVersion` here as the default, while keeping `.Values.image.tag` empty instead of implicitly defaulting to `latest`? 

Both the `.Chart.Version` and `.Chart.AppVersion` could be tracking the current app version, and then aligned via an automatic commit as a part of the regular release workflow action run. 

After that, the workflow being added in this PR would pick up the change based on the path filter and publish the new chart version, where all versions would be aligned.

This is more or less the standard approach for repos that have both the application source and the chart source in the same place: the chart version lags one release behind in every tag, but the pushed Helm OCI charts are always spot on version-wise, and there's no extra Helm-induced burden for the maintainers.

Helm-only hotfixes could still be added by manually bumping the `.Chart.Version` to a patch semver number, so that a new OCI artifact is produced, but the versions would align upon the next application release.

chart/templates/deployment.yaml
```diff
@@ -0,0 +1,211 @@
...
+              mountPath: /certs
+              readOnly: true
+            {{- end }}
+      volumes:
+        - name: data
+          {{- if .Values.persistence.enabled }}
+          persistentVolumeClaim:
+            claimName: {{ include "versitygw.pvcName" . }}
+          {{- else }}
+          emptyDir: {}
```
crabique, line 192: Wouldn't this be a potential footgun when `.Values.replicaCount ≥ 1` and `.Values.gateway.backend.type == "posix"`?

chart/values.yaml
```diff
@@ -0,0 +1,253 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+# Standard Helm chart settings
+nameOverride: ""
+fullnameOverride: ""
+
+# Number of pod replicas for the deployment
+# Note that when using the "posix" backend with a PVC,
+# you can only increase the number of replicas if you have a "ReadWriteMany" volume!
```
crabique, line 11: It is also (usually) possible with `ReadWriteOnce` (but not `ReadWriteOncePod`), however the caveat is that all replicas will have to be scheduled onto the same node, as kubelet will be able to share the CSI mount to multiple pods attaching the same volume claim.

This is generally achievable with no configuration overhead via a strict inter-pod affinity constraints on all replicas attaching the volume, and is generally useful when the PVCs are block/local volumes, but the user wants to have more processes handling the workload.

benmcclelland:
@crabique thank you for the help with the review, I'm not as experienced with the helm charts

chart/templates/deployment.yaml
jacksgt: > Wouldn't it make sense to rely on | default .Chart.AppVersion here as the default, while keeping .Values.image.tag empty instead of implicitly defaulting to latest?

Defaulting the Helm chart installation to latest is consistent with other getting started instructions for versitygw, i.e.:

```
helm install versitygw oci://ghcr.io/versity/versitygw/chart
```
picks up the latest version of the Helm chart and of Versity itself, similar to:
```
docker run --rm versity/versitygw:latest --version
```
(documented in the README of the project)

Clearly, for production use cases the user should pin a specific Helm chart version and App version.

> Both the .Chart.Version and .Chart.AppVersion could be tracking the current app version, and then aligned via an automatic commit as a part of the regular release workflow action run.

I'm generally not really a fan of coupling the Helm chart version with Application version, at least for use cases where there is no strong coupling needed. Exceptions here are stuff like Kubernetes operators, where it frequently happens that a new version of the operator also needs a new RBAC permissions and therefore an updated Helm chart.
In the case of versitygw, I don't see a need to publish a new Helm chart version for each release of the main application.

chart/templates/deployment.yaml
jacksgt: Yes, not reading the values before deploying a Helm chart is a footgun :-)

Are you suggesting to have `persistence.enabled=true` by default? We can do that.

chart/values.yaml
jacksgt: Your remark is very true, and my comment is only a simplification because I did not want to go into all the details regarding Kubernetes volume access modes in my comment. Do you want me to insert your text instead?

chart/templates/deployment.yaml
crabique: I'm rather suggesting explicitly erroring out in this scenario and stating in the values/readme that any ≥1 replicas setup with POSIX backend requires persistence.

Specifically with `emptyDir` multiple replicas would result in a "working" deployment with different files in each replica: e.g. `PutObject` would only create a file in one replica, but then a `GetObject` for it would go to a different replica.

chart/values.yaml
crabique: I was thinking whether it'd make sense to automagically set the pod affinity under the hood based on the specified volume mode. With that, the readme may just state that when `ReadWriteOnce` is used all pods will be co-located on a single node.

But yeah, now that I read this, it may be a bit too much magic 🙂

chart/templates/deployment.yaml
crabique: It just makes life a lot easier with things like Renovate: applications that treat the chart as part of the distribution will get both the chart and the image updates with just one chart-level integration.

But this was just a question, I understand your point as well.

chart/templates/deployment.yaml
jacksgt: Sure, I can add a check for that.

chart/values.yaml
jacksgt: > But yeah, now that I read this, it may be a bit too much magic 🙂

We can elaborate on this in the chart's README :-)

For now this is a first working version, and we can always improve in the future based on the user's feedback and needs.

chart/templates/deployment.yaml
jacksgt: Renovate can also update the `image.tag` inside a Helm values.yaml just fine :-)

I expect/predict that at the beginning we will need some iteration on the Helm chart (hence the `0.1.0` version). Once most of bugs have been fixed and features added (to the Helm chart), we can go to a stable `1.x` release or align the releases with versitygw application itself.

crabique: Makes sense, though this chart is already of a much better quality than most projects ever get, you did a tremendous job with this 🙂

benmcclelland:
This chart looks like its working pretty well from my tests. I'll go ahead and merge this, and we can track any further requested changes in new issues/PRs. thanks!

jacksgt:
@crabique Thanks for your review and inputs, very much appreciated! I have created a follow-up PR that addresses the concerns you have raised: https://github.com/versity/versitygw/pull/1939

PR #1931, opened 2026-03-04
fix: add request headers and metadata headers limit

### LINKED ISSUE #1606: [Bug] - Object metadata size should be limited
**Describe the bug**
When uploading an object(PutObject, CopyObject, CreateMultipartUpload), its metadat size is limited in S3. Currently the gateway doesn't have any limitations on request or specifically on metadat size.

S3 docs say:
>The PUT request header is limited to 8 KB in size. Within the PUT request header, the user-defined metadata is >limited to 2 KB in size. The size of user-defined metadata is measured by taking the sum of the number of bytes >in the UTF-8 encoding of each key and value.

**To Reproduce**
```go
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/credentials"
	"github.com/aws/aws-sdk-go-v2/service/s3"
	"github.com/aws/smithy-go/logging"
)

var (
	accessKeyId      = "<access_key_id>"
	secret           = "<secret_access_key>"
	region           = "us-east-1"
	bucket           = "niksis02-bucket-6"
	object           = "my-obj"
)

func main() {
	meta := map[string]string{}

	for i := range 10000 {
		meta[fmt.Sprintf("meta-%v", i)] = "hello"
	}

	client := s3.NewFromConfig(aws.Config{
		Credentials: credentials.NewStaticCredentialsProvider(accessKeyId, secret, ""),
		Region:      "us-east-1",
	})
	_, err := client.PutObject(context.Background(), &s3.PutObjectInput{
		Bucket:         &bucket,
		Key:            &object,
		Metadata:       meta,
	}, func(o *s3.Options) {
		o.Logger = logging.NewStandardLogger(os.Stdout)
		o.ClientLogMode = aws.LogResponseWithBody
	})

	fmt.Println(err)
}
```

**Expected behavior**
By running this script, the gateway has to return a raw 400 Bad Request error response.
Debug log output:
```
SDK 2025/10/29 23:07:25 DEBUG Response
HTTP/1.1 400 Bad Request
Connection: close
Content-Length: 0
Date: Wed, 29 Oct 25 19:07:25 GMT
Server: AmazonS3
X-Amz-Id-2: iZi1Vhv5xoXAYjMafVpKNl1enjdFt6d6MhxXVtp+t/NKlStGnaW6vOMn0qBZpudXjk3K6n4uCS1xsL4lOPHgGYKho5LkqLMQOk1S60Tk8qP1CICAWarZ+qsywQsFayEi
X-Amz-Request-Id: 61BFF61F2C29A04E

operation error S3: PutObject, https response error StatusCode: 400, RequestID: 61BFF61F2C29A04E, HostID: iZi1Vhv5xoXAYjMafVpKNl1enjdFt6d6MhxXVtp+t/NKlStGnaW6vOMn0qBZpudXjk3K6n4uCS1xsL4lOPHgGYKho5LkqLMQOk1S60Tk8qP1CICAWarZ+qsywQsFayEi, api error BadRequest: Bad Request
```

niksis02:
Fixes #1606

According to AWS documentation:
> *“The PUT request header is limited to 8 KB in size. Within the PUT request header, the user-defined metadata is limited to 2 KB in size. The size of user-defined metadata is measured by taking the sum of the number of bytes in the UTF-8 encoding of each key and value.”*

Based on this, object metadata size is now limited to **2 KB** for all object upload operations (`PutObject`, `CopyObject`, and `CreateMultipartUpload`).

It also fixes the handling of metadata HTTP headers when the same header appears multiple times with different casing or even if they are identical. According to S3 behavior, these headers must be merged into a single lower-cased metadata key, with values concatenated using commas.

Example:

```
x-amz-meta-Key: value1
x-amz-meta-kEy: value2
x-amz-meta-keY: value3
```

Translated to:

```
key: value1,value2,value3
```

This PR also introduces an **8 KB limit for request headers**. Although the S3 documentation explicitly mentions the 8 KB limit only for **PUT requests**, in practice this limit applies to **all requests**.

To enforce the header size limit, the Fiber configuration option `ReadBufferSize` is used. This parameter defines the maximum number of bytes read when parsing an incoming request. Note that this limit does not apply strictly to request headers only, since request parsing also includes other parts of the request line (e.g., the HTTP method, protocol string, and version such as `HTTP/1.1`). So `ReadBufferSize` is effectively a limit for request headers size, but not the exact limit.

### DIFF
--- a/s3api/server.go
+++ b/s3api/server.go
@@ -91,6 +91,10 @@ func New(
 		DisableStartupMessage: true,
 		ErrorHandler:          globalErrorHandler,
 		Concurrency:           server.maxConnections,
+		// Sets buffer limit to read/parse incoming requests
+		// if the limit is reached, fiber/fasthttp will throw an error
+		// in the global error handler
+		ReadBufferSize: 8 * 1024, // 8 KB
 	})
 
 	server.app = app
@@ -259,13 +263,20 @@ func globalErrorHandler(ctx *fiber.Ctx, er error) error {
 		// handle the fiber specific errors
 		var fiberErr *fiber.Error
 		if errors.As(er, &fiberErr) {
+			if errors.Is(fiberErr, fiber.ErrRequestHeaderFieldsTooLarge) {
+				debuglogger.Logf("total request headers size exceeds the allowed 8KB")
+				ctx.Status(http.StatusBadRequest)
+				return nil
+			}
 			if strings.Contains(fiberErr.Message, "cannot parse Content-Length") {
+				debuglogger.Logf("failed to parse Content-Length")
 				ctx.Status(http.StatusBadRequest)
 				return nil
 			}
 			if strings.Contains(fiberErr.Message, "error when reading request headers") {
 				// This error means fiber failed to parse the incoming request
 				// which is a malfoedmed one. Return a BadRequest in this case
+				debuglogger.Logf("failed to parse the http request")
 				err := s3err.GetAPIError(s3err.ErrCannotParseHTTPRequest)
 				ctx.Status(err.HTTPStatusCode)
 				return ctx.Send(s3err.GetAPIErrorResponse(err, "", "", ""))
--- a/s3api/utils/utils.go
+++ b/s3api/utils/utils.go
@@ -55,20 +55,52 @@ func SetBucketNameValidationStrict(strict bool) {
 	strictBucketNameValidation.Store(strict)
 }
 
-func GetUserMetaData(headers *fasthttp.RequestHeader) (metadata map[string]string) {
-	metadata = make(map[string]string)
+// maximum allowed size (2 KB) for all user-defined
+// object metadata combined, excluded the 'x-amz-meta-' prefix
+const maxMetadataSize = 2048
+
+// GetUserMetaData extracts user metadata from headers with the "x-amz-meta-" prefix.
+// Keys are normalized to lowercase and duplicate headers are merged as
+// comma-separated values. The total metadata size is validated against the
+// 2 KB S3 limit.
+func GetUserMetaData(headers *fasthttp.RequestHeader) (map[string]string, error) {
+	metadata := make(map[string]string)
+	var metadataSize int
+
 	headers.DisableNormalizing()
+	defer headers.EnableNormalizing()
+
 	for key, value := range headers.AllInOrder() {
 		hKey := string(key)
 		if strings.HasPrefix(strings.ToLower(hKey), "x-amz-meta-") {
 			trimmedKey := strings.ToLower(hKey[11:])
 			headerValue := string(value)
-			metadata[trimmedKey] = headerValue
+
+			if existingVal, ok := metadata[trimmedKey]; ok {
+				// S3 combines multiple metadata headers with the same key.
+				// For example:
+				//   x-amz-meta-Key: value1
+				//   x-amz-meta-kEy: value2
+				//   x-amz-meta-keY: value3
+				//
+				// These are merged into a single lowercased key with the values
+				// joined as a comma-separated list:
+				//   key: value1,value2,value3
+				metadataSize += len(headerValue) + 1
+				metadata[trimmedKey] = existingVal + "," + headerValue
+			} else {
+				metadataSize += len(trimmedKey) + len(headerValue)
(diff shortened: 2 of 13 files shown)


s3api/utils/utils.go
```diff
@@ -55,20 +55,31 @@ func SetBucketNameValidationStrict(strict bool) {
...
+
 	headers.DisableNormalizing()
+	defer headers.EnableNormalizing()
+
 	for key, value := range headers.AllInOrder() {
 		hKey := string(key)
 		if strings.HasPrefix(strings.ToLower(hKey), "x-amz-meta-") {
 			trimmedKey := strings.ToLower(hKey[11:])
 			headerValue := string(value)
+			metadataSize += len(trimmedKey) + len(headerValue)
```
benmcclelland: It looks like according to the docs that we should be including the prefix size in this calculation.  This is currently undercounting by 11 bytes per header I think.

s3api/utils/utils.go
```diff
@@ -55,20 +55,31 @@ func SetBucketNameValidationStrict(strict bool) {
...
+
+			if metadataSize > maxMetadataSize {
```
benmcclelland, line 96: Do we know the aws behavior for requests with repeated headers of the same key? Does it accumulate all like this is doing, or does it count it after the de-duplication?

tests/integration/group-tests.go
benmcclelland, line 841: Missing TestServer(ts) here?

tests/integration/server.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+
+		defer resp.Body.Close()
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return err
+		}
+
+		if len(body) != 0 {
+			return fmt.Errorf("expected emtpy response body, instead got %s", body)
```
benmcclelland: emtpy -> empty

s3api/utils/utils.go
niksis02: The `x-amz-meta-` prefix is ignored by S3, when counting the length

s3api/utils/utils.go
niksis02: I was just testing the behavior and truns out, S3 combines multiple same meta header values into a single meta key as a comma separated list.
e.g.
```
x-amz-meta-Key: value1
x-amz-meta-kEy: value2
x-amz-meta-keY: value3
```
Is transleted to:

`key: value1,value2,value3`

niksis02: I'm working on the fix

s3api/utils/utils.go
```diff
@@ -55,20 +55,53 @@ func SetBucketNameValidationStrict(strict bool) {
...
+				//   x-amz-meta-Key: value1
+				//   x-amz-meta-kEy: value2
+				//   x-amz-meta-keY: value3
+				//
+				// These are merged into a single lowercased key with the values
+				// joined as a comma-separated list:
+				//   key: value1,value2,value3
+				metadataSize += len(headerValue) + 1
+				metadata[trimmedKey] = existingVal + "," + headerValue
+				continue
```
benmcclelland: intended that the size check is skipped here?

s3api/server.go
```diff
@@ -259,13 +263,20 @@ func globalErrorHandler(ctx *fiber.Ctx, er error) error {
 		// handle the fiber specific errors
 		var fiberErr *fiber.Error
 		if errors.As(er, &fiberErr) {
+			if errors.Is(fiberErr, fiber.ErrRequestHeaderFieldsTooLarge) {
+				debuglogger.Logf("total request headers size exceeds the allowed 8KB")
+				ctx.Status(http.StatusBadRequest)
```
benmcclelland, line 268: do we need to be sending an aws compatible error back for this case? or is that for a future issue?

s3api/utils/utils.go
niksis02: Ahh, thanks for noticing. `continue` isn't intended to be there

s3api/server.go
niksis02: This is the exact aws behavior: blank 400 response: no response body or specific headers

PR #1932, opened 2026-03-05
feat: add unix domain socket listener support to port option

benmcclelland:
This adds the ability to specify unix domain socket paths for the service listener with the --port <path> option. Where <path> can be either a path to a file in a filesystem or prefixed with @ for an abstract socket name.

### DIFF
--- a/cmd/versitygw/main.go
+++ b/cmd/versitygw/main.go
@@ -156,6 +156,19 @@ documentation can be found in the GitHub wiki.`,
 			admPorts = ctx.StringSlice("admin-port")
 			webuiGateways = ctx.StringSlice("webui-gateways")
 			webuiAdminGateways = ctx.StringSlice("webui-admin-gateways")
+
+			// Resolve relative UNIX socket paths to absolute before any backend
+			// (e.g. posix) can change the working directory via os.Chdir.
+			var err error
+			if ports, err = utils.AbsSocketPaths(ports); err != nil {
+				return err
+			}
+			if admPorts, err = utils.AbsSocketPaths(admPorts); err != nil {
+				return err
+			}
+			if webuiPorts, err = utils.AbsSocketPaths(webuiPorts); err != nil {
+				return err
+			}
 			return nil
 		},
 		Action: func(ctx *cli.Context) error {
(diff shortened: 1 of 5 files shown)


cmd/versitygw/main.go
```diff
@@ -181,7 +181,7 @@ func initFlags() []cli.Flag {
 		},
 		&cli.StringSliceFlag{
 			Name:    "port",
-			Usage:   "gateway listen address <ip>:<port> or :<port> (can be specified multiple times for listening on multiple addresses)",
+			Usage:   "gateway listen address: <ip>:<port>, :<port>, /path/to/socket for file-backed UNIX sockets, or @name for Linux abstract namespace sockets (can be specified multiple times for listening on multiple addresses)",
```
niksis02, line 197: Update `example.conf` to add info about unix sockets ?

s3api/utils/multi_listener.go
```diff
@@ -210,7 +237,25 @@ func resolveHostnameAddrs(address string) ([]string, error) {
...
+// UNIX domain socket forms are also supported:
+//   - "/path/to/socket" or "./rel/socket" — file-backed socket; any stale
+//     socket file is removed before binding.
+//   - "@name" — Linux abstract namespace socket; no file is created or removed.
 func NewMultiAddrListener(network, address string) (net.Listener, error) {
+	if IsUnixSocketPath(address) {
+		// For file-backed sockets, remove any stale file so re-binding works cleanly.
+		// Abstract sockets (@name) have no filesystem entry; skip os.Remove for them.
+		if !isAbstractSocket(address) {
+			os.Remove(address)
```
crabique: Probably also need to verify that it's `fs.ModeSocket` that's being removed, someone might accidentally use it as an `rm` 😄

s3api/utils/multi_listener.go
```diff
@@ -135,10 +137,31 @@ func (ml *MultiListener) Addr() net.Addr {
 	return nil
 }
 
+// IsUnixSocketPath reports whether addr is a UNIX domain socket path.
+// Recognised forms:
+//   - "/path/to/socket" or "./relative/socket" — file-backed socket
+//   - "@name" — Linux abstract namespace socket (no filesystem entry)
+func IsUnixSocketPath(addr string) bool {
+	return strings.HasPrefix(addr, "/") ||
+		strings.HasPrefix(addr, "./") ||
```
crabique: the relative path seems to be a problematic use-case: with posix backend the CWD is set to the data directory _before_ the listener is initialized, so the socket is being created there instead of the CWD of the shell invoking the gateway startup

s3api/utils/multi_listener.go
```diff
@@ -135,10 +138,69 @@ func (ml *MultiListener) Addr() net.Addr {
...
+		strings.HasPrefix(addr, "@")
+}
```
niksis02, line 149: I think we are missing a case when the unix socket file path is provided, which doesn't meet any of these conditions.
e.g. `--port app.sock` or `--port run/app.sock`

These examples are valid unix socket file paths, but they will be treated as tcp `address:port`.

As tcp forms follow some strict rules, my recommendation is to first try to parse/validate the input string as a tcp form and treat everything else as UNIX socket path.

benmcclelland: thats a good point, i'll check to see if there are any issues with hostnames though

PR #1935, opened 2026-03-06
feat(chart): add HTTPRoute configuration

aclerici38:
Adds HTTPRoute configuration to the chart in place of ingress for those who have adopted the Gateway API. 

**NOTE**: In a HTTPRoute you're unable to route to separate backends with separate hostnames in a single resource like you can with ingress. I chose to add `webui.httpRoute` and `admin.httpRoute`, however that may lead to some confusion since there are no ingress fields under the same keys. I considered adding a global `httpRoutes` key with each route under that but thought this was better ux reading through the chart values. Let me know if you would prefer that or another approach.

Thanks!

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -203,7 +203,7 @@ webui:
   # Kubernetes because it uses the internal pod IP addresses.
   adminGateways: []
   # - s3-admin.example.com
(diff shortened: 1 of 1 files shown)


versity-github:

PR #1940, opened 2026-03-06
fix(chart): replace buggy chart-releaser action

aclerici38:
Updates the workflow to only run on changes to Chart.yaml and swaps to manual logic for checking if the chart's version has been updated. 

The previously used bitdeps/helm-oci-charts-releaser action seems buggy IMHO and not created for the use-case here. On a commit that didn't touch the chart [it looks like it still packaged and pushed it](https://github.com/versity/versitygw/actions/runs/22778993128/job/66079589190) (I think this is because it checks git tags and there is no tag created for the chart) and throughout the action's repo there's some inconsistencies between `released_charts` and `changed_charts` for the output variable that leads to it never being set, meaning the cosign steps here never ran.

The manual logic uses `helm show` on the version in chart.yaml. If it errors, the version doesn't exist in GHCR and the chart will be pushed. if it succeeds the version already exists and the rest of the steps are skipped

### DIFF
--- a/.github/workflows/helm-chart.yml
+++ b/.github/workflows/helm-chart.yml
@@ -4,40 +4,48 @@ on:
   push:
     branches:
       - main
+    paths:
+      - "chart/Chart.yaml"
 
 jobs:
   release:
     runs-on: ubuntu-latest
     permissions:
-      contents: write
       packages: write
       id-token: write
     steps:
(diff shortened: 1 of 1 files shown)


versity-github:

benmcclelland (approved):

PR #1941, opened 2026-03-07
fix(chart): fix multidoc separator when 2 routes are enabled

aclerici38:
The multidoc separators were before the conditional, so when 2 routes were enabled it renders `---apiVersion`.

Sorry for missing this, I tested with 1 and all 3, not 2 😄


versity-github:

benmcclelland:
No problem, thanks for the quick fix

PR #1943, opened 2026-03-07
feat: add option to change webui path prefix

benmcclelland:
New option webui-path-prefix that specifies a prefix for the webui endpoint. The default is the old behavior at the root.

### DIFF
--- a/cmd/versitygw/main.go
+++ b/cmd/versitygw/main.go
@@ -100,6 +100,7 @@ var (
 	webuiNoTLS                             bool
 	webuiGateways                          []string
 	webuiAdminGateways                     []string
+	webuiPathPrefix                        string
 	disableACLs                            bool
 )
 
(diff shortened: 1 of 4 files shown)


cmd/versitygw/main.go
```diff
@@ -219,6 +221,12 @@ func initFlags() []cli.Flag {
 			Usage:   "override auto-detected admin gateway URLs for WebUI (e.g. 'http://localhost:7080', 'https://admin.example.com'; can be specified multiple times)",
 			EnvVars: []string{"VGW_WEBUI_ADMIN_GATEWAYS"},
 		},
+		&cli.StringFlag{
+			Name:        "webui-path-prefix",
+			Usage:       "mount the WebUI under a path prefix (e.g. '/ui'); must start with '/' and must not end with '/'",
+			EnvVars:     []string{"VGW_WEBUI_PATH_PREFIX"},
+			Destination: &webuiPathPrefix,
+		},
```
niksis02, line 242: Maybe adding a validation for `webuiPathPrefix` would make sense ?

benmcclelland: i'll check to see how this fails, if its not obvious then yeah we need some validation

benmcclelland: added validation check

PR #1944, opened 2026-03-07
fix(chart): correct cosign-installer tag, revert path filter

aclerici38:
https://github.com/versity/versitygw/actions/runs/22793758976
The cosign-installer action doesn't have a major version tag for v4. There is one for v3 but [they recommend upgrading](https://github.com/sigstore/cosign-installer/releases/tag/v3.10.1) and I think they stopped updating the v3 tag, [it points to v3.9.1 from June](https://github.com/sigstore/cosign-installer/tags).

I also reverted the path filter. I don't think it makes a ton of sense to have it run each push to main but with the version check it's harmless and it's the original behavior here.

thanks!

### DIFF
--- a/.github/workflows/helm-chart.yml
+++ b/.github/workflows/helm-chart.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - main
-    paths:
-      - "chart/Chart.yaml"
 
 jobs:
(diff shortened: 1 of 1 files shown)


versity-github:

PR #1945, opened 2026-03-08
feat: option to host webui on s3 port

benmcclelland:
This adds the webui-s3-prefix option to specify a prefix and host the webui on the same port as the s3 service. Like the health endpoint, this will mask any bucket with the same name as the webui prefix.

The benefit of hosting this on the same interface as the s3 service is no longer needing the CORS headers for the browser access if the webui and s3 access are on the same IP:PORT.

### DIFF
--- a/cmd/versitygw/main.go
+++ b/cmd/versitygw/main.go
@@ -101,6 +101,7 @@ var (
 	webuiGateways                          []string
 	webuiAdminGateways                     []string
 	webuiPathPrefix                        string
+	webuiS3Prefix                          string
 	disableACLs                            bool
 )
 
@@ -239,6 +240,12 @@ func initFlags() []cli.Flag {
 			Usage:       "mount the WebUI under a path prefix (e.g. '/ui'); must be single segment path that starts with '/'",
 			EnvVars:     []string{"VGW_WEBUI_PATH_PREFIX"},
 			Destination: &webuiPathPrefix,
+		},
+		&cli.StringFlag{
+			Name:        "webui-s3-prefix",
+			Usage:       "mount the WebUI on the S3 port at the given path prefix (e.g. '/ui'); must start with '/', must not be '/', and must not end with '/'",
+			EnvVars:     []string{"VGW_WEBUI_S3_PREFIX"},
+			Destination: &webuiS3Prefix,
 		},
 		&cli.StringFlag{
 			Name:        "access",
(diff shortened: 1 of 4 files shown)


cmd/versitygw/main.go
```diff
@@ -817,6 +824,19 @@ func runGateway(ctx context.Context, be backend.Backend) error {
...
+			return fmt.Errorf("--webui-s3-prefix must start with '/'")
+		}
+		if webuiS3Prefix == "/" {
+			return fmt.Errorf("--webui-s3-prefix must not be '/' (use a specific path like '/ui')")
+		}
+		if strings.HasSuffix(webuiS3Prefix, "/") {
+			return fmt.Errorf("--webui-s3-prefix must not end with '/'")
+		}
+	}
+
```
niksis02: DRY: Maybe the same `validateWebUIPathPrefix` can be used to validate `webui-s3-prefix` ?

niksis02 (commented):
When both `--webui-s3-prefix` and `--webui` flags are specified, the web ui is hosted twice: on s3api with prefix and as a separate process with --webui port. The primary concern is if we should allow the webui to be hosted twice? Is that intentional ?

benmcclelland:
> When both `--webui-s3-prefix` and `--webui` flags are specified, the web ui is hosted twice: on s3api with prefix and as a separate process with --webui port. The primary concern is if we should allow the webui to be hosted twice? Is that intentional ?

Yes, I think it should be fine to allow this. I originally prevented this, but took out the check since it seems fine if someone really wants to do this.

cmd/versitygw/main.go
```diff
@@ -813,10 +820,19 @@ func runGateway(ctx context.Context, be backend.Backend) error {
...
 		return err
 	}
 
+	// Validate webui-s3-prefix
+	if webuiS3Prefix != "" {
+		err := validateWebUIPathPrefix("--webui-s3-prefix", webuiS3Prefix)
+		if err != nil {
+			return err
+		}
+	}
```
niksis02: nit: double check for empty string: once before calling `validateWebUIPathPrefix`, once inside `validateWebUIPathPrefix`. Remove the check before calling `validateWebUIPathPrefix` ?

cmd/versitygw/main.go
```diff
@@ -1644,35 +1727,41 @@ func validateGatewayURLs(urls []string, urlType string) ([]string, error) {
...
 // Accepted format is a single path segment like "/ui".
-func validateWebUIPathPrefix(prefix string) error {
+func validateWebUIPathPrefix(option, prefix string) error {
 	if prefix == "" {
 		return nil
 	}
 
 	if strings.TrimSpace(prefix) != prefix {
-		return fmt.Errorf("invalid --webui-path-prefix %q: must not contain leading or trailing whitespace", prefix)
+		return fmt.Errorf("invalid -%v %q: must not contain leading or trailing whitespace",
```
niksis02: Extra `-`(dash) ? Either `--webui-s3-prefix` or `--webui-path-prefix` is passed to the funcation, the `dash` seems to be unnecessary. `-%v` -> `%v`

cmd/versitygw/main.go
```diff
@@ -1644,35 +1727,41 @@ func validateGatewayURLs(urls []string, urlType string) ([]string, error) {
...
 
 	if strings.ContainsAny(prefix, "?#") {
-		return fmt.Errorf("invalid --webui-path-prefix %q: query strings and fragments are not allowed", prefix)
+		return fmt.Errorf("invalid %v %q: query strings and fragments are not allowed",
+			option, prefix)
 	}
 
 	if strings.Contains(prefix, "\\") {
-		return fmt.Errorf("invalid --webui-path-prefix %q: backslashes are not allowed", prefix)
+		return fmt.Errorf("invalid -%v %q: backslashes are not allowed",
```
niksis02: `-%v` - > `%v`

PR #1947, opened 2026-03-09
fix(chart): use docker login for helm and cosign

aclerici38:
Cosign does not take credentials from helm login but does from docker login, and helm does as well. Verified here https://github.com/aclerici38/pocket-id-operator/actions/runs/22818484620/job/66187482832

Sorry again for all the noise I've caused the last few days. This should have been a couple very simple additions to the new, great chart.

Thanks!!


versity-github:

Issue #1957, opened 2026-03-10
[Bug] - Helm chart always tries to create PVC


aclerici38:
**Describe the bug**
In the chart, `persistence.claimName` is supposed to mount an existing claim https://github.com/versity/versitygw/blob/aebe09182be8715244917a71cfaea2f278f91366/chart/values.yaml#L285-L286 But the chart seems to try to create a PVC no matter what, just with the name from claimName if it's set. 
~This could be destructive as I think helm now "owns" the PVC and will try to delete it on uninstall or persistence disabled, which probably isn't intended when mounting an existing claim. 
As for fixing it, I'm not sure how without risking data loss for anyone who currently has `claimName` set...~

**To Reproduce**
```
helm template versitygw oci://ghcr.io/versity/versitygw/charts/versitygw:0.2.1 --set auth.accessKey=test --set auth.secretKey=test --set persistence.enabled=true --set persistence.claimName=test --show-only templates/pvc.yaml
Pulled: ghcr.io/versity/versitygw/charts/versitygw:0.2.1
Digest: sha256:31b4bb229808b0779cbc14e8faad44e3fbb7e58744b27f6774a393bcdfac6e25
---
# Source: versitygw/templates/pvc.yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: test
  labels:
    helm.sh/chart: versitygw-0.2.1
    app.kubernetes.io/name: versitygw
    app.kubernetes.io/instance: versitygw
    app.kubernetes.io/managed-by: Helm
spec:
  accessModes:
    - "ReadWriteOnce"
  resources:
    requests:
      storage: 10Gi
```

**Expected behavior**
with persistence enabled and claimName set the chart should not try to create a PVC

**Chart Version**
0.2.1

**Server Version**
v1.3.1

**Additional context**
<!-- Describe s3 client and version if applicable.

aclerici38:
So that PR is the obvious fix but A) doesn't handle deleting the PVC for anyone already using `claimName` B) seems like it's an ai bot? joined yesterday with 38 forks and a suspiciously quick PR to my issue @jsodf78h823f

edit: looks like the bot got banned. anyways...

jacksgt:
You are correct, I missed this. Will open a PR ASAP.

> This could be destructive as I think helm now "owns" the PVC and will try to delete it on uninstall or persistence disabled

Not sure if I fully understand you, but: Helm only owns the resources it also created.
[...]

PR #1962, opened 2026-03-10
fix: allow Helm chart users to use their own, existing PVC

jacksgt:
closes https://github.com/versity/versitygw/issues/1957

the following configurations are now possible:
```
# create a PVC with the Helm chart name
persistence:
  enabled: true # default: true
  create: true # default: true


# create a PVC with a custom name
persistence:
  enabled: true
  create: true
  claimName: foobar


# re-use an existing PVC
persistence:
  enabled: true
  create: false
  claimName: foobar


# do not use a PVC at all
persistence:
  enabled: false
```


versity-github:

PR #1964, opened 2026-03-12
feat: add persistence.hostPath to Helm chart

shoce:
https://github.com/versity/versitygw/issues/980

This adds `persistence.hostPath` to `values.yaml`. I tried to make it so it does not need to set `persistence.create` to false to use `hostPath` instead of creating a pvc.

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -89,7 +89,11 @@ The `gateway.backend.type` value selects the storage backend. Use `gateway.backe
 
 ## Scaling and Persistence
 
-By default, this chart enables persistence via a `PersistentVolumeClaim` (PVC) to ensure data consistency and prevent data loss. To use a hostPath volume set `persistence.hostPath`.
+By default, this chart enables persistence via a `PersistentVolumeClaim` (PVC) to ensure data consistency and prevent data loss. 
+
+Alternatively, you may use [hostPath volume](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) by setting the `persistence.hostPath` value.
+As a general rule, this setup should only be used if all nodes in the cluster have access to the same data (e.g. NFS share is mounted on all nodes) or for single-node use cases.
+Special care must be taken particularly when using multiple replicas with such a setup, since Versity does not perform internal data replication ("clustering").
 
 ### Horizontal Scaling (replicas > 1)
 


versity-github:

benmcclelland:
@jacksgt for review for this

chart/README.md
```diff
@@ -84,12 +84,12 @@ The `gateway.backend.type` value selects the storage backend. Use `gateway.backe
...
 | **WebUI** | `webui.enabled=true` — browser-based management UI on `webui.port` (default `8080`); set `webui.apiGateways` and `webui.adminGateways` to your externally reachable endpoints |
 | **IAM** | `iam.enabled=true` — flat-file identity and access management stored alongside backend data |
-| **Persistence** | `persistence.enabled=true` — provisions a PVC for backend data and IAM storage; defaults to `10Gi` |
+| **Persistence** | `persistence.enabled=true` — provisions a PVC for backend data and IAM storage; defaults to `10Gi`, or uses a hostPath volume specified by `persistence.hostPath` |
 | **NetworkPolicy** | `networkPolicy.enabled=true` — restricts ingress to selected pods/namespaces; allows all egress |
 
 ## Scaling and Persistence
 
-By default, this chart enables persistence via a `PersistentVolumeClaim` (PVC) to ensure data consistency and prevent data loss.
+By default, this chart enables persistence via a `PersistentVolumeClaim` (PVC) to ensure data consistency and prevent data loss. To use a hostPath volume set `persistence.hostPath`.
```
crabique: Some documentation concerns:

- People looking for a replacement for their MinIO/Garage/etc deployments may not fully understand that VersityGW does _not_ do any clustering/cross-talk between instances.

  So if they don't have a storage class configured they may just think "`hostPath` is easier to use, let's just use that", which with multiple replicas would result in inconsistent state just like with `emptyDir`.

  Would it make sense to add a disclaimer that the `hostPath` needs to be a network filesystem available to all replicas when there's ≥1 replica? 

- Even with just 1 replica and non-network filesystem, for it to be actual persistence there needs to be a strong node affinity guarantee.

benmcclelland: I was thinking the same thing. It looks like it would be likely for people to end up with inconsistent replicas if they don't understand this. There could be a check for >1 replicas, but is there a scenario where this host path could be cluster storage?

shoce: should we add a check that hostPath is used with not more than one replica? and a short comment saying it is the restriction because there is no sense in using hostPath with more than one replica?

jacksgt: It can definitely make sense to use a hostPath volume with multiple replicas: for example if you have several nodes, mount an NFS share at /data on each of them and want to serve this data with Versity from multiple nodes.

I think a disclaimer in the documentation (Helm chart README) should be sufficient.

benmcclelland: I think that would be best, mainly to prevent the case where people think that this can somehow create a cluster storage environment. Also, claimName and hostPath set should probably fail as well. I think these are mutually exclusive?

shoce: honestly i am not sure how to phrase that warning about using hostPath with more than one replice, can you please help with it?

shoce: > Also, claimName and hostPath set should probably fail as well. I think these are mutually exclusive?

yes, it is already there in `deployment.yaml`:
```
{{- if and .Values.persistence.claimName .Values.persistence.hostPath }}
  {{- fail "only persistence.claimName or persistence.hostPath can be set" }}
{{- end }}
```

jacksgt:
```suggestion
By default, this chart enables persistence via a `PersistentVolumeClaim` (PVC) to ensure data consistency and prevent data loss. 

Alternatively, you may use [hostPath volume](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) by setting the `persistence.hostPath` value.
As a general rule, this setup should only be used if all nodes in the cluster have access to the same data (e.g. NFS share is mounted on all nodes) or for single-node use cases.
Special care must be taken particularly when using multiple replicas with such a setup, since Versity does not perform internal data replication ("clustering").
```

jacksgt: > honestly i am not sure how to phrase that warning about using hostPath with more than one replice, can you please help with it?

Sure. I have left a comment with a suggestion.

jacksgt: @shoce have you considered using a [`local` volume](https://kubernetes.io/docs/concepts/storage/volumes/#local) for your setup?

> Using the hostPath volume type presents many security risks. If you can avoid using a hostPath volume, you should. For example, define a [local PersistentVolume](https://kubernetes.io/docs/concepts/storage/volumes/#local), and use that instead.

https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

shoce: > @shoce have you considered using a [`local` volume](https://kubernetes.io/docs/concepts/storage/volumes/#local) for your setup?

i did not know about local volumes before, i will definitely take a deeper look, at the first glance they look like a better option than hostPath, but for now hostPath is kind of standard solution in where i work and i would like to first deploy it "the standard way" and later migrate to local volume if it really suits. thanks a lot, i will definitely learn about local volumes.

crabique: I agree local volumes are a much better solution for cases where `hostPath` isn't NFS/EFS/SMB/CephFS etc as it takes care of the required scheduling shenanigans automatically.

With actual node-local `hostPath` something like this could easily happen:
- you start a 1-replica deployment, create some buckets and put some data on it
- something happens that causes pod recreation (could even be chart/image upgrade with `rollingUpdate`)
- the pod starts on a different node that doesn't have the previous state under that hostPath
- at best there's no data in your S3 now

So it's essentially a hard requirement to have one of the following params pointing to a single cluster node:
- `nodeAffinity. requiredDuringSchedulingIgnoredDuringExecution`
- `nodeSelector`
- `nodeName`

PR #1966, opened 2026-03-13
chore: remove the unnecessary AmazonS3.xsd from s3response.

niksis02:


benmcclelland:
fix? feels like this should be chore instead. its not really fixing anything.

PR #1975, opened 2026-03-18
Do not expose directories that can't be buckets

jvansanten:
If a directory name doesn't pass bucket name validation, don't bother returning it from ListBuckets.


versity-github:

PR #1976, opened 2026-03-18
Add flag to disable io.Copy

jvansanten:
First of all, I've found this project extremely useful!

In my application the storage backend (dCache) is mounted over NFSv4.2, but does not support server-side copy. Instead of replying to copy_file_range requests with an error, though, it silently drops them, causing CompleteMultiPartUpload to hang forever.

This bug has been reported and [fixed](https://github.com/dCache/nfs4j/commit/260a6949a64bc228e6cd9c5313060e8a9f2cf474), but has yet to make it into a release. This PR adds a `--disable-copy-file-range` flag and corresponding `VGW_DISABLE_COPY_FILE_RANGE` environment variable to instruct the posix backend to explicitly use block copies to reassemble files in CompleteMultiPartUpload so the gateway can be used with unpatched dCache clusters or other storage with incomplete server-side copy support.

### DIFF
--- a/cmd/versitygw/posix.go
+++ b/cmd/versitygw/posix.go
@@ -139,15 +139,15 @@ func runPosix(ctx *cli.Context) error {
 	}
 
 	opts := posix.PosixOpts{
-		ChownUID:            chownuid,
-		ChownGID:            chowngid,
-		BucketLinks:         bucketlinks,
-		VersioningDir:       versioningDir,
-		NewDirPerm:          fs.FileMode(dirPerms),
-		ForceNoTmpFile:      forceNoTmpFile,
+		ChownUID:             chownuid,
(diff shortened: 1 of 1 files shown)


versity-github:

benmcclelland:
I don't think you need to implement your own copy, the io.Copy() will fallback to normal copy if you prevent either the source io.Reader or the destination io.Writer from implementing the io.WriterTo or io.ReaderFrom interfaces. For example:
```go
type onlyRead struct {
	io.Reader
}

if disableCopyFileRange {
	 io.Copy(dstFile, &onlyRead{srcFile})
}
```

benmcclelland:
the current test failure is from gofmt, its just wants the changes to be formatted in a gofmt compatible way

PR #1977, opened 2026-03-18
Ensure tempfiles are removed

jvansanten:
When a client provides an invalid or incomplete body for a single-part upload, the handler returns before the link stage.  Factor tempfile removal into cleanup() to catch all cases.


versity-github:

backend/posix/with_otmpfile.go
benmcclelland: should this one be removed as well since the cleanup() should handle this?

PR #1978, opened 2026-03-18
Update scoutfs glacier stage constants for ongoing requests

57-Wolve:
Fix issue with incompatible S3 response for offline and staging status.

Resolves issue with MinIO SDK returning "unexpected restore header" and causing restore requests to fail.


versity-github:

PR #1982, opened 2026-03-20
feat: add browser-based POST object upload support

### LINKED ISSUE #1648: [Feature] - Browser-Based Uploads Using POST
> Amazon S3 supports HTTP POST requests so that users can upload content directly to Amazon S3. By using POST, end users can authenticate requests without having to pass data through a secure intermediary node that protects your credentials. Thus, HTTP POST has the potential to reduce latency.

> The POST operation adds an object to a specified bucket by using HTML forms. POST is an alternate form of PUT that enables browser-based uploads as a way of putting objects in buckets. Parameters that are passed to PUT through HTTP headers are instead passed as form fields to POST in the multipart/form-data encoded message body. To add an object to a bucket, you must have WRITE access on the bucket.

The idea is to add this `POST` request support in the gateway.

**Additional context**
[S3 POST Object](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html)
[Browser-Based Uploads Using POST (AWS Signature Version 4)](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-UsingHTTPPOST.html)

### LINKED ISSUE #1980: [Bug] - EntityTooSmall error message
**Describe the bug**
The `EntityTooSmall` error message slightly differes from s3.
**s3**: `Your proposed upload is smaller than the minimum allowed size`
**versitygw**: `Your proposed upload is smaller than the minimum allowed object size.`

The same is for `EntityTooLarge`:
**s3**: `Your proposed upload exceeds the maximum allowed size`
**versitygw**: `Your proposed upload exceeds the maximum allowed object size.`

**To Reproduce**
```bash

aws s3api create-multipart-upload --bucket test --key mp

aws s3api upload-part --bucket test --key mp --upload-id <upload-id> --part-number 1 --body file.bin // file.bin should be less than 5mb
aws s3api upload-part --bucket test --key mp --upload-id <upload-id> --part-number 2 --body file.bin // file.bin should be less than 5mb

aws s3api complete-multipart-upload --bucket test --key mp --upload-id <upload-id> --multipart-upload file://mp.json
```

**Expected behavior**
The following error should be returned:
`An error occurred (EntityTooSmall) when calling the CompleteMultipartUpload operation: Your proposed upload is smaller than the minimum allowed size`

### LINKED ISSUE #1981: [Bug] - PutObjectTagging request body structure
**Describe the bug**
The `PutObjectTagging` request body syntax is the following according to s3 docs.
```xml
<Tagging xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
   <TagSet>
      <Tag>
         <Key>string</Key>
         <Value>string</Value>
      </Tag>
   </TagSet>
</Tagging>
```

Which means that the go sdk `types.Tagging` doesn't not match to what s3 expects and any request which uses `types.Tagging` as request body type to send should fail with `MalformedXML` error. The `PutBucketTagging_success_status` integration test should fail with `MalformedXML` error.

**To Reproduce**
```bash
./versitygw test -a <access> -s <secret> -e http://127.0.0.1:7070 PutBucketTagging_success_status
```

**Expected behavior**
The test should not succeed and it should fail with a `MalformedXML` error.

niksis02:
Closes #1648
Fixes #1980
Fixes #1981

This PR implements browser-based POST object uploads for S3-compatible form uploads. It adds support for handling `multipart/form-data` object uploads submitted from browsers, including streaming multipart parsing so file content is not buffered in memory, POST policy decoding and evaluation, SigV4-based form authorization, and integration with the existing `PutObject` backend flow. The implementation covers the full browser POST upload path, including validation of required form fields, credential scope and request date checks, signature verification, metadata extraction from `x-amz-meta-*` fields, checksum field parsing, object tagging conversion from XML into the query-string format expected by `PutObject`, and browser-compatible success handling through `success_action_status` and `success_action_redirect`. It also wires the new flow into the router and metrics layer and adds POST-specific error handling and debug logging across policy parsing, multipart parsing, and POST authorization. AWS S3 also accepts the `redirect` form field alongside `success_action_redirect`, but since AWS has marked `redirect` as deprecated and is planning to remove it, this gateway intentionally does not support it.

### DIFF
--- a/auth/post_policy.go
+++ b/auth/post_policy.go
@@ -28,6 +28,11 @@ import (
 	"github.com/versity/versitygw/s3err"
 )
 
+const (
+	policyFieldExpiration = "expiration"
+	policyFieldConditions = "conditions"
+)
+
 // POSTPolicy is the parsed browser-based upload policy document.
 type POSTPolicy struct {
 	expiration time.Time
@@ -49,12 +54,7 @@ type postPolicyCondition interface {
 	validate() error
 	match(PostPolicyEvalInput) error
 	coveredField() string
-}
-
-// rawPolicy mirrors the JSON structure of an incoming POST policy document.
-type rawPolicy struct {
-	Expiration string            `json:"expiration"`
-	Conditions []json.RawMessage `json:"conditions"`
+	isBucketCondition() bool
 }
 
 // ParsePOSTPolicyBase64 decodes and validates a base64-encoded POST policy.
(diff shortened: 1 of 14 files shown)


auth/post_policy.go
```diff
@@ -0,0 +1,503 @@
...
+func rawScalarToString(raw json.RawMessage) (string, error) {
+	var v any
+	if err := json.Unmarshal(raw, &v); err != nil {
+		return "", err
+	}
+
+	switch x := v.(type) {
+	case string:
+		return x, nil
+	case json.Number:
```
benmcclelland, line 550: unreachable case, all JSON numbers are unmarshaled into float64. you can enable this with a decoder option though:
```go
decoder := json.NewDecoder(bytes.NewReader(raw))
decoder.UseNumber()
decoder.Decode(&v)
```

s3api/utils/multipart_reader.go
```diff
@@ -0,0 +1,349 @@
...
+			return nil, s3err.GetAPIError(s3err.ErrMalformedPOSTRequest)
+		}
+
+		_, hasFilename := params["filename"]
+		isFilePart := name == "file" || hasFilename
+
+		// At this point, headers + blank line have already been consumed,
+		// so bytesRead points exactly at the first byte of file content.
+		if isFilePart {
+			fileContentLength := mp.requestContentLength - mp.bytesRead - int64(len(mp.boundary)) - 8
```
benmcclelland: add a const to clarify the value here:
```go
const finalBoundaryDelimiterLen = 8 // len("\r\n--") + len("--\r\n")
```

If the client sends a multipart body with an epilogue after the final boundary, this calculation would produce a wrong value.

The finalFileReader correctly ignores epilogue bytes during streaming, but the pre-computed ContentLength passed to the backend will be wrong in those cases.

s3api/middlewares/object-post-auth.go
```diff
@@ -0,0 +1,187 @@
...
+
+			// Parse the date and check the date validity
+			tdate, err := time.Parse(iso8601Format, amzDate)
+			if err != nil {
+				debuglogger.Logf("invalid POST object x-amz-date %q: %v", amzDate, err)
+				return s3err.GetAPIError(s3err.ErrInvalidDateHeader)
+			}
+
+			// Validate the dates difference
+			// TODO: Seems s3 doesn't validate this
```
benmcclelland: should we be calling this validate if s3 doesnt validate the date?

niksis02: I'm currently working on this

s3api/utils/utils.go
```diff
@@ -986,10 +1071,12 @@ func GenerateObjectLocation(ctx *fiber.Ctx, virtualDomain, bucket, object string
 	scheme := ctx.Protocol()
 	host := ctx.Hostname()
 
+	fmt.Println(host, virtualDomain, bucket, object)
```
benmcclelland: extra println in here

PR #1993, opened 2026-03-26
fix: serialize concurrent CompleteMultipartUpload calls via rename

benmcclelland:
When two requests raced to complete the same multipart upload, the first caller to finish would remove the part files and upload directory. The second caller, already past the initial existence check, would then fail mid-flight with confusing errors such as ErrInvalidPart or an I/O error when trying to open a part that no longer exists.

Fix this by atomically renaming the upload directory from <uploadID> to <uploadID>.inprogress at the very start of CompleteMultipartUploadWithCopy, before any part data is read. A concurrent caller will now find the original directory absent and receive a clean NoSuchUpload error. A deferred rename restores the original name if the complete does not succeed, allowing the client to retry.

ListMultipartUploads is updated to skip any directories whose name ends in .inprogress so in-flight completes do not appear as pending uploads.

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -106,6 +106,7 @@ const (
 	MetaTmpDir          = ".sgwtmp"
 	MetaTmpMultipartDir = MetaTmpDir + "/multipart"
 	onameAttr           = "objname"
+	inProgressSuffix    = ".inprogress"
 	tagHdr              = "X-Amz-Tagging"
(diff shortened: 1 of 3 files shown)


benmcclelland:
> That would actually be very nice to have an integration test case for this.

I'll work on one

benmcclelland:
> > That would actually be very nice to have an integration test case for this.
> 
> I'll work on one

done:
```
PASS CompleteMultipartUpload_racey_data_integrity
```

PR #1994, opened 2026-03-27
test: remove eval, go command generation overhaul

lrm25:


tests/drivers/rest.sh
```diff
@@ -175,17 +179,22 @@ send_rest_command_expect_success_callback() {
...
+  if ! curl_command=$(go run ./tests/rest_scripts/generateCommand.go -awsAccessKeyId "$AWS_ACCESS_KEY_ID" -awsSecretAccessKey "$AWS_SECRET_ACCESS_KEY" -awsRegion "$AWS_REGION" -url "$AWS_ENDPOINT_URL" "-writeXMLPayloadToFile" "$xml_file" "$@" 2>&1); then
     log 2 "error: $curl_command"
     return 1
   fi
-  local full_command="send_command $curl_command"
-  log 5 "full command: $full_command"
-  if ! result=$(eval "${full_command[*]}" 2>&1); then
+  curl_command=$(echo -n "$curl_command" | tr -d '\n')
+  mapfile -t curl_command_array < <(
+    python3 -c 'import shlex, sys; print("\n".join(shlex.split(sys.argv[1])))' "$curl_command"
```
anodelman: use:
echo -n "$curl_command" | python3 -c 'import shlex, sys; [print(arg) for arg in shlex.split(sys.stdin.read())]'

you are passing "$curl_command" as a shell argument to Python. If that string contains a double quote, it can prematurely terminate the argument or cause a shell syntax error

PR #1997, opened 2026-03-30
fix: enforce 5gb copy source object size threshold.

### LINKED ISSUE #1896: [Bug] - `CopyObject` 5GB+ object copy
**Describe the bug**
When attempting to `CopyObject` a `5GB+` object, S3 returns an error, while versitygw successfully copies the object. Note that a 5GB+ object can be uploaded only with multipart uploads.

**To Reproduce**
```bash

// create a 5GB+ object with multipart uplod
aws s3 cp large-file.bin s3://test/obj // large-file.bin is 5GB+

aws s3api copy-object --bucket test --key dst-obj --copy-source test/obj
```

**Expected behavior**
The following error has to be returned:
```
An error occurred (InvalidRequest) when calling the CopyObject operation: The specified copy source is larger than the maximum allowable size for a copy source: 5368709120
```

niksis02:
Fixes #1896

Enforces the S3 `5 GiB` copy source size limit across the posix and azure backends for `CopyObject` and `UploadPartCopy`, returning `InvalidRequest` when the source object exceeds the threshold.

The limit is now configurable via `--copy-object-threshold` (`VGW_COPY_OBJECT_THRESHOLD`, default 5 GiB).
A new `--mp-max-parts flag` (`VGW_MP_MAX_PARTS`, default `10000`) has been added to make multipart upload parts number limit configurable.

No integration test has been added, as GitHub Actions cannot reliably handle large objects.

### DIFF
--- a/backend/azure/azure.go
+++ b/backend/azure/azure.go
@@ -95,16 +95,17 @@ func (key) Table() map[string]struct{} {
 type Azure struct {
 	backend.BackendUnsupported
 
-	client         *azblob.Client
-	sharedkeyCreds *azblob.SharedKeyCredential
-	defaultCreds   *azidentity.DefaultAzureCredential
-	serviceURL     string
-	sasToken       string
+	client              *azblob.Client
+	sharedkeyCreds      *azblob.SharedKeyCredential
+	defaultCreds        *azidentity.DefaultAzureCredential
+	serviceURL          string
+	sasToken            string
+	copyObjectThreshold int64
 }
 
 var _ backend.Backend = &Azure{}
 
-func New(accountName, accountKey, serviceURL, sasToken string) (*Azure, error) {
+func New(accountName, accountKey, serviceURL, sasToken string, copyObjectThreshold int64) (*Azure, error) {
 	url := serviceURL
 	if serviceURL == "" && accountName != "" {
 		// if not otherwise specified, use the typical form:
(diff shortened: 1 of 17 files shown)


backend/azure/azure.go
benmcclelland: missing copyObjectThreshold

benmcclelland: missing copyObjectThreshold

backend/posix/posix.go
```diff
@@ -3079,6 +3091,9 @@ func (p *Posix) UploadPartCopy(ctx context.Context, upi *s3.UploadPartCopyInput)
 	if err != nil {
 		return s3response.CopyPartResult{}, fmt.Errorf("stat object: %w", err)
 	}
+	if fi.Size() > p.copyObjectThreshold {
```
benmcclelland: I think this should be using length from backend.ParseCopySourceRange() instead of fi.Size()?

backend/posix/posix.go
```diff
@@ -235,6 +246,7 @@ func New(rootdir string, meta meta.MetadataStorer, opts PosixOpts) (*Posix, erro
 		forceNoCopyFileRange: opts.ForceNoCopyFileRange,
 		validateBucketName:   opts.ValidateBucketNames,
 		actionLimiter:        semaphore.NewWeighted(int64(concurrencyOrDefault(opts.Concurrency))),
+		copyObjectThreshold:  opts.CopyObjectThreshold,
```
benmcclelland, line 249: I think we need to wire this through for scoutfs as well in scoutfs_compat.go otherwise it will be setting this to 0.

PR #1998, opened 2026-03-30
test: more ListObjects REST coverage

lrm25:


tests/drivers/file.sh
```diff
@@ -288,3 +289,342 @@ get_file_names() {
...
+  for ((i=0;i<$2;i++)); do
+    if ! file_name=$(create_test_file_with_prefix "$1" 2>&1); then
+      log 2 "error creating test file: $file_name"
+      return 1
+    fi
+    file_names+=("$file_name")
+  done
+  echo "${file_names[*]}"
+  return 0
+}
```
anodelman, line 325: Don't need two function here, can do with one with an optional count argument:

```
# Usage: create_test_files_with_prefix <prefix> [count]
# Returns: Space-separated list of created filenames
create_test_files_with_prefix() {
  local prefix="$1"
  local count="${2:-1}"  # Default to 1 if not provided
  local file_names=()
  local file_name
  local error

  if [[ -z "$prefix" ]]; then
    log 2 "error: prefix is required"
    return 1
  fi

  for ((i=0; i<count; i++)); do
    # Generate the name
    if ! file_name=$(get_file_name_with_prefix "$prefix" 2>&1); then
      log 2 "error getting file name: $file_name"
      return 1
    fi

    # Create the file
    if ! error=$(create_test_file "$file_name" 2>&1); then
      log 2 "error creating test file: $error"
      return 1
    fi

    file_names+=("$file_name")
  done

  echo "${file_names[*]}"
  return 0
}
```

tests/drivers/file.sh
```diff
@@ -288,3 +289,342 @@ get_file_names() {
...
+  fi
+  if [ "$file_size" -eq 0 ]; then
+    return 0
+  fi
+  if ! error=$(dd if=/dev/urandom of="$TEST_FILE_FOLDER/$1" bs=1 count="$file_size" 2>&1); then
+    log 2 "error adding data to file: $error"
+    return 1
+  fi
+  return 0
+}
```
anodelman, line 379: Can be combined into a single function:
```
# Combined function to setup environment and create test files
# Params: filename1 [filename2 ...] 
# Note: Uses $FILE_SIZE if set, otherwise defaults to 10 bytes.
create_test_files() {
  log 6 "Initializing test file creation..."

  if [[ $# -lt 1 ]]; then
    log 2 "'create_test_files' requires at least one filename"
    return 1
  fi

  if [[ -z "$TEST_FILE_FOLDER" ]]; then
    export TEST_FILE_FOLDER=$(mktemp -d -t test_files_XXXXXX)
    log 6 "Created temporary test folder: $TEST_FILE_FOLDER"
  fi

  mkdir -p "$TEST_FILE_FOLDER"

  local file_size="${FILE_SIZE:-10}" # Use global $FILE_SIZE or default to 10
  local error

  for filename in "$@"; do
    local full_path="$TEST_FILE_FOLDER/$filename"

    # Clean up existing file if present
    if [[ -e "$full_path" ]]; then
      if ! error=$(rm -f "$full_path" 2>&1); then
        log 2 "error removing existing file $filename: $error"
        return 1
      fi
    fi

    # Create the file with random data
    if [[ "$file_size" -eq 0 ]]; then
      touch "$full_path"
    else
      # Use dd for specific size creation
      if ! error=$(dd if=/dev/urandom of="$full_path" bs=1 count="$file_size" conv=notrunc 2>&1); then
        log 2 "error adding $file_size bytes to $filename: $error"
        return 1
      fi
    fi
    
    log 6 "Created: $full_path ($file_size bytes)"
  done

  return 0
}
```

tests/drivers/file.sh
```diff
@@ -288,3 +289,342 @@ get_file_names() {
...
+    fi
+  fi
+  echo "$file_size"
+}
+
+# split file into pieces to test multipart upload
+# param: file location
+# return 0 for success, 1 for error
+split_file() {
+  if ! check_param_count_v2 "file name, number of pieces" 2 $#; then
```
anodelman, line 457: Can simplify the logic here:

```
# -n l/K : Split into K pieces without breaking lines (or use 'K' for raw bytes)
  # -d     : Use numeric suffixes
  # -a 2   : Allow up to 100 pieces (00-99)
  if ! error=$(split -a 2 -d -n "$num_pieces" "$file_path" "${file_path}-" 2>&1); then
    log 2 "error splitting file: $error"
    return 1
  fi
```
Then you don't need to calculate part size.

tests/drivers/file.sh
```diff
@@ -288,3 +289,342 @@ get_file_names() {
...
+    if ! file_one_md5=$(md5 -q "$1" 2>&1); then
+      log 2 "error getting md5 for '$1': $file_one_md5"
+      return 2
+    fi
+    if ! file_two_md5=$(md5 -q "$2" 2>&1); then
+      log 2 "error getting md5 for '$2': $file_two_md5"
+      return 2
+    fi
+  else
+    if ! file_one_md5=$(md5sum "$1" | cut -d " " -f 1 2>&1); then
```
anodelman: This is going to give you the error status of the cut command and not the checksum calculation.

More streamlined code:
```
compare_files() {
   if ! check_param_count_v2 "two files" 2 $#; then
    return 2
  fi
  log 5 "comparing files '$1' and '$2'"

  local file1="$1"
  local file2="$2"
  local md5_cmd
  local f1_sum f2_sum

  if [[ "$(uname)" == "Darwin" ]]; then
    md5_cmd="md5 -q"
  else
    md5_cmd="md5sum"
  fi

  if ! f1_raw=$($md5_cmd "$file1" 2>&1); then
    log 2 "error getting md5 for '$file1': $f1_raw"
    return 2
  fi  
  # Clean the output (extract just the hex hash)
  f1_sum=$(echo "$f1_raw" | awk '{print $1}')

  if ! f2_raw=$($md5_cmd "$file2" 2>&1); then
    log 2 "error getting md5 for '$file2': $f2_raw"
    return 2
  fi
   # Clean the output (extract just the hex hash)
  f2_sum=$(echo "$f2_raw" | awk '{print $1}')

  if [[ "$f1_sum" == "$f2_sum" ]]; then
    return 0
  fi

  log 2 "MD5 mismatch: $f1_sum vs $f2_sum"
  return 1
}
```

tests/drivers/file.sh
```diff
@@ -288,3 +289,342 @@ get_file_names() {
...
+  if ! check_param_count_v2 "file name, size in MB" 2 $#; then
+    return 1
+  fi
+  filesize=$(($2*1024*1024))
+  if ! error=$(dd if=/dev/urandom of="$TEST_FILE_FOLDER"/"$1" bs=1024 count=$((filesize/1024)) 2>&1); then
+    log 2 "error adding data to large file: $error"
+    return 1
+  fi
+  return 0
+}
```
anodelman, line 556: Can combine create_large_file()/create_large_file_v2()/create_large_file_with_size() into a simpler single function.  This probably throws off the param count check, since we have optional params, but is easier to use overall.

```
# Usage: create_large_file [filename] [size_in_mb]
# If filename is omitted, it generates one. Defaults to 160MB.
create_large_file() {
  local file_name="$1"
  local size_mb="${2:-160}"
  local error

  if [[ -z "$file_name" ]]; then
    if ! file_name=$(get_file_name 2>&1); then
      log 2 "error generating automatic file name: $file_name"
      return 1
    fi
  fi

  log 6 "Creating ${size_mb}MB file: $file_name"

  mkdir -p "${TEST_FILE_FOLDER:?}"

  # bs=1M is significantly faster than bs=1024 for large files
  if ! error=$(dd if=/dev/urandom of="${TEST_FILE_FOLDER}/${file_name}" bs=1M count="$size_mb" 2>&1); then
    log 2 "error creating ${size_mb}MB file at ${file_name}: $error"
    return 1
  fi

  echo "$file_name"
  return 0
}
```

anodelman (changes requested):

tests/drivers/file.sh
lrm25: Done, but leaving a `create_test_file` stub function for now to avoid changing too much this PR

tests/drivers/file.sh
```diff
@@ -288,3 +290,345 @@ get_file_names() {
...
+        return 1
+      fi
+    fi
+
+    # Create the file with random data
+    if [[ "$file_size" -eq 0 ]]; then
+      touch "$full_path"
+    else
+      # Use dd for specific size creation
+      if ! error=$(dd if=/dev/urandom of="$full_path" bs=1 count="$file_size" conv=notrunc 2>&1); then
```
anodelman: this tiny block size is very inefficient (to make a 1KB file it would make 1024 calls of 1 byte each) - is this intended?  any reason for this?

tests/drivers/file.sh
```diff
@@ -288,3 +290,345 @@ get_file_names() {
...
+    return 1
+  fi
+  return 0
+}
+
+create_file_single_char() {
+  if ! check_param_count_v2 "filename, size, char" 3 $#; then
+    return 1
+  fi
+  if [[ -e "$TEST_FILE_FOLDER/$1" ]]; then
```
anodelman: could just do an rm -f, eats the error if the file doesn't exist

tests/drivers/file.sh
```diff
@@ -288,3 +290,345 @@ get_file_names() {
...
+  local file_size="${FILE_SIZE:-10}" # Use global $FILE_SIZE or default to 10
+  local error
+
+  log 5 "file size: $file_size"
+  for filename in "$@"; do
+    local full_path="$TEST_FILE_FOLDER/$filename"
+
+    # Clean up existing file if present
+    if [[ -e "$full_path" ]]; then
+      if ! error=$(rm -f "$full_path" 2>&1); then
```
anodelman: don't need to do the file exists checks here, rm -f will eat the failure if it doesn't exist.

tests/drivers/file.sh
```diff
@@ -288,3 +290,345 @@ get_file_names() {
...
+delete_test_files() {
+  if ! check_param_count_gt "filenames" 1 $#; then
+    return 1
+  fi
+  if [ -z "$TEST_FILE_FOLDER" ]; then
+    log 2 "no test file folder defined, not deleting"
+    return 1
+  fi
+  for name in "$@"; do
+    rm -rf "${TEST_FILE_FOLDER:?}"/"${name:?}" || rm_result=$?
```
anodelman: Since this is a single file you don't need -r.

tests/drivers/file.sh
```diff
@@ -288,3 +290,345 @@ get_file_names() {
...
+    return 0
+  fi
+
+  log 2 "MD5 mismatch: $f1_sum ($file1) vs $f2_sum ($file2)"
+  return 1
+}
+
+# Usage: create_large_file [filename] [size_in_mb]
+# If filename is omitted, it generates one. Defaults to 160MB.
+create_large_file() {
```
anodelman, line 526: I do think that you could merge all these file creation methods, using something like this block:

```
if [[ "$requested_size" -eq 0 ]]; then
      touch "$full_path"
    elif [[ "$requested_size" -lt "$block_size" ]]; then
      # For small files, use the size as the block size (1 write)
      dd if=/dev/urandom of="$full_path" bs="$requested_size" count=1 conv=notrunc 2>&1
    else
      # For large files, use 1MB chunks to save RAM
      count=$(( requested_size / block_size ))
      remainder=$(( requested_size % block_size ))
      
      # Write the big 1MB chunks
      dd if=/dev/urandom of="$full_path" bs="$block_size" count="$count" conv=notrunc 2>&1
      
      # If there's a remainder (e.g., 1.5MB), append the last few bytes
      if [[ $remainder -gt 0 ]]; then
        dd if=/dev/urandom of="$full_path" bs=1 count="$remainder" oflag=append conv=notrunc 2>&1
      fi
    fi
```

tests/drivers/file.sh
```diff
@@ -288,3 +290,345 @@ get_file_names() {
...
+    ls_result=$(ls "$TEST_FILE_FOLDER/file_*")
+    log 5 "$ls_result"
+  fi
+  return 0
+}
+
+download_and_compare_file_with_user() {
+  if ! check_param_count_gt "original file, bucket, key, destination, username, password, chunk size (optional)" 6 $#; then
+    return 1
+  fi
```
anodelman, line 581: Should probably clear the dest dir before download_file to ensure that you don't end up comparing old data if download_file_with_user misbehaves.

anodelman (changes requested):

tests/drivers/file.sh
lrm25: I can do this in the next PR, since this one has gotten so large already

anodelman (approved):

PR #2005, opened 2026-04-01
feat: implement x-amz-source-expected-bucket-owner for CopyObject and UploadPartCopy

### LINKED ISSUE #1897: [Feature] - CopyObject `x-amz-source-expected-bucket-owner`
As `x-amz-expected-bucket-owner` is supported in the gateway, it makes sense to add the `x-amz-source-expected-bucket-owner` support in `CopyObject`.

**AWS docs say**
>`x-amz-source-expected-bucket-owner`
>The account ID of the expected source bucket owner. If the account ID that you provide does not match the actual owner of the source bucket, the request fails with the HTTP status code 403 Forbidden (access denied).

**Additional context**
[AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html#API_CopyObject_RequestSyntax)

niksis02:
Closes #1897

Extract the `X-Amz-Source-Expected-Bucket-Owner` header for CopyObject and UploadPartCopy. Verify the source bucket owner in the backend and if the provided access key id doesn't match, return an `AccessDenied` error.

### DIFF
--- a/s3api/controllers/object-put.go
+++ b/s3api/controllers/object-put.go
@@ -360,6 +360,7 @@ func (c S3ApiController) UploadPartCopy(ctx *fiber.Ctx) (*Response, error) {
 	key := strings.TrimPrefix(ctx.Path(), fmt.Sprintf("/%s/", bucket))
 	copySource := strings.TrimPrefix(ctx.Get("X-Amz-Copy-Source"), "/")
 	copySrcRange := ctx.Get("X-Amz-Copy-Source-Range")
+	expectedSrcBucketOwnerUPC := ctx.Get("X-Amz-Source-Expected-Bucket-Owner")
 	partNumber := int32(ctx.QueryInt("partNumber", -1))
 	uploadId := ctx.Query("uploadId")
 	// context locals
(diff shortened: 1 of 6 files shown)


s3api/controllers/object-put.go
```diff
@@ -429,6 +430,7 @@ func (c S3ApiController) UploadPartCopy(ctx *fiber.Ctx) (*Response, error) {
 			CopySourceIfNoneMatch:       preconditionHdrs.IfNoneMatch,
 			CopySourceIfModifiedSince:   preconditionHdrs.IfModSince,
 			CopySourceIfUnmodifiedSince: preconditionHdrs.IfUnmodeSince,
+			ExpectedSourceBucketOwner:   &expectedSrcBucketOwnerUPC,
```
benmcclelland, line 433: s3.UploadPartCopyInput has ExpectedBucketOwner, is this expected to be added in here too?
```go
    // The account ID of the expected destination bucket owner. If the account ID that
    // you provide does not match the actual owner of the destination bucket, the
    // request fails with the HTTP status code 403 Forbidden (access denied).
    ExpectedBucketOwner *string
```

benmcclelland: actually, if this is an issue we can take care of this in a new PR. current PR looks good scoped to source owner

PR #2006, opened 2026-04-01
fix: retry link on ENOENT caused by racing DeleteObject

### LINKED ISSUE #1988: [Bug] - ExistingObjectIsDirectory during uploading an object
**Describe the bug**

I was benchmarking versitygw (POSIX backend) with a script that uploads and deletes objects with unique filenames, and uses 12 concurrent threads to upload objects.  From time to time (every few thousand requests or so) versitygw returns HTTP 409 ExistingObjectIsDirectory. 

I added a few debug statements in the versitygw source, and found out the error is thrown here:

https://github.com/versity/versitygw/blob/271d7377dd84983faecf4b6d50390c7ec5cd928a/backend/posix/posix.go#L3563

In this location the code does not check what type of error err is, so `ErrExistingObjectIsDirectory` may be misleading. I added the following debug statement before this line:

```
fmt.Printf("oh no 3! %v\n", err.Error())
```

And the output was:

```
oh no 3! link tmpfile (fd "14" as "pings/33e95e4c-56cb-4051-a117-ec339dad5a0b/ugcbdec-378657"): no such file or directory
```

I found out the POSIX backend has a `--disableotmp` flag. As an experiment I ran versitygw with this flag and could not reproduce the issue. I ran into a different issue (at one point the filesystem permissions of the object's parent directory got messed up) but I will file that separately if I can reliably reproduce it.

**To Reproduce**

I have a convoluted testing setup that I cannot easily share. I can try to put together a self-contained script. 

**Expected behavior**

No 409 responses.

**Server Version**

1.3.1

**Additional context**

OS: Ubuntu 24.04
versitygw start command: 

```
ROOT_ACCESS_KEY="testuser" ROOT_SECRET_KEY="V7f1CwQqAcwo80UEIJEjc5gVQUSSx5ohQ9GSrr12" versitygw  --port :10000 posix /tmp/vgw
```

The `/tmp` directory on this system appears to be a normal directory under the root filesystem, it is not using tmpfs or anything like that. The root filesystem is ext4. The filesystem 43% full, and uses 16% of available inodes.

benmcclelland:
A concurrent PutObject and DeleteObject on the same prefix directory can race:
PutObject opens an O_TMPFILE in MetaTmpDir (not yet visible in the fs) DeleteObject removes the last visible object in the prefix directory and calls removeParents(), which rmdir's the now-empty prefix directory
PutObject's link() tries to link the fd into a parent directory that no longer exists

Fix by detecting ENOENT in the final link step (Linkat, Rename, and MoveFile) and retrying after recreating the parent directory.

Also extract linkatOTmpfile() to consolidate the Linkat+EEXIST→Renameat logic that was previously inline in link().

Fixes #1988

### DIFF
--- a/backend/posix/with_otmpfile.go
+++ b/backend/posix/with_otmpfile.go
@@ -43,7 +43,7 @@ type tmpfile struct {
 	objname    string
 	isOTmp     bool
 	size       int64
-	needsChown bool
+	doChown    bool
 	uid        int
 	gid        int
 	newDirPerm fs.FileMode
@@ -87,7 +87,7 @@ func (p *Posix) openTmpFile(dir, bucket, obj string, size int64, acct auth.Accou
 		objname:    obj,
 		isOTmp:     true,
 		size:       size,
-		needsChown: doChown,
+		doChown:    doChown,
 		uid:        uid,
 		gid:        gid,
 		newDirPerm: p.newDirPerm,
@@ -126,13 +126,13 @@ func (p *Posix) openMkTemp(dir, bucket, obj string, size int64, dofalloc bool, u
 		return nil, err
 	}
 	tmp := &tmpfile{
-		f:          f,
-		bucket:     bucket,
-		objname:    obj,
-		size:       size,
-		needsChown: doChown,
-		uid:        uid,
-		gid:        gid,
+		f:       f,
+		bucket:  bucket,
+		objname: obj,
+		size:    size,
+		doChown: doChown,
+		uid:     uid,
+		gid:     gid,
 	}
 	// falloc is best effort, its fine if this fails
 	if size > 0 && dofalloc {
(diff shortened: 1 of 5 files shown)


backend/posix/with_otmpfile.go
```diff
@@ -244,33 +277,54 @@ func (tmp *tmpfile) fallbackLink() error {
...
+		}
+	}
 	if err != nil {
 		// rename only works for files within the same filesystem
 		// if this fails fallback to copy
-		return backend.MoveFile(tempname, objPath, fs.FileMode(defaultFilePerm))
+		backoffMs := initialBackoffMs
+		for range maxDirRecreateRetries {
+			// Add jitter to avoid synchronized retry waves.
+			sleepWithJitter(backoffMs)
```
niksis02: Sleeping at the very beginning of the for loop would add an unnecessary latency for the cross-device rename (EXDEV) for the first attempt. I think the sleep should come after `MoveFile`, so that we are sure that we put the code on sleep after the first failed `MoveFile`, because most of the cases it will succeed and the delay is not needed.

backend/posix/with_otmpfile.go
```diff
@@ -159,6 +159,38 @@ func (tmp *tmpfile) falloc() error {
...
+		if errors.Is(err, syscall.EEXIST) && retries < maxTmpFileNameRetries {
+			continue
+		}
+		if err != nil {
+			return fmt.Errorf("cannot find free temporary file: %w", err)
+		}
+		err = unix.Renameat(dirFd, tmpName, dirFd, basename)
+		if err != nil {
+			return fmt.Errorf("overwriting renameat failed: %w", err)
+		}
```
niksis02, line 191: If `Linkat` succeeds but `Renameat` does not, an orphaned `".%s.sgwtmp.%d"` file is left in the target directory. Maybe an `Unlinkat` is needed in `Renameat` error path?
`unix.Unlinkat(dirFd, tmpName, 0)`

backend/posix/without_otmpfile.go
```diff
@@ -30,11 +30,20 @@ import (
...
-	objname string
-	size    int64
+	f          *os.File
+	bucket     string
+	objname    string
+	size       int64
+	newDirPerm fs.FileMode
+	uid        int
+	gid        int
+	doChown    bool
```
niksis02, line 46: nit: in `with_otmpfile` it's called `needsChown`. It might be confusing to have different names. Keep both either `needsChown` or `doChown` ?

PR #2011, opened 2026-04-02
fix: check PutObjectTagging/LegalHold/Retention permissions on PutObject,CopyObject and CreateMultipartUpload

### LINKED ISSUE #1986: [Bug] - PutObject/CreateMultipartUpload permissions
**Describe the bug**
When creating an object using `PutObject` or `CreateMultipartUpload`, a client can specify tagging, legal hold, or retention settings. Each of these requires distinct bucket policy permissions:

* tagging → `s3:PutObjectTagging`
* legal hold → `s3:PutObjectLegalHold`
* retention → `s3:PutObjectRetention`

Therefore, these permissions must be explicitly granted to the requesting user (not just an admin) in order to allow access to these features.

Currently, `versitygw` only checks for the `s3:PutObject` permission and allows modification of tagging, legal hold, and retention without validating the corresponding specific permissions.

**To Reproduce**
```bash

// run as 'testuser'
aws s3api put-object --bucket test --key obj --tagging "Key1=Value1&Key2=Value2"
aws s3api put-object --bucket test --key obj --object-lock-legal-hold-status ON
aws s3api put-object --bucket test --key obj --object-lock-mode GOVERNANCE --object-lock-retain-until-date "2026-12-31T00:00:00Z"
```

**Expected behavior**
S3 in this case returns something like:
```
An error occurred (AccessDenied) when calling the PutObject operation: User: arn:aws:iam::792168558830:user/test is not authorized to perform: s3:PutObjectLegalHold on resource: "arn:aws:s3:::niksis02-bucket-1/green" because no identity-based policy allows the s3:PutObjectLegalHold action
```

niksis02:
Fixes #1986
    
When a client includes tagging, legal hold, or retention headers in a PutObject, CopyObject or CreateMultipartUpload request, the corresponding bucket policy permissions must be verified in addition to s3:PutObject:
    
`X-Amz-Tagging` - `s3:PutObjectTagging`
`X-Amz-Object-Lock-Legal-Hold` - `s3:PutObjectLegalHold`
`X-Amz-Object-Lock-Mode` - `s3:PutObjectRetention`
    
Previously, only s3:PutObject was checked, allowing users to set tagging, legal hold, and retention without having the required permissions. Now each action permission is check, if user tries to add them.
    
For CopyObject these permissions are checked on destination object.

### DIFF
--- a/auth/bucket_policy.go
+++ b/auth/bucket_policy.go
@@ -234,7 +234,11 @@ func ValidatePolicyDocument(policyBin []byte, bucket string, iam IAMService) err
 	return nil
 }
 
-func VerifyBucketPolicy(policy []byte, access, bucket, object string, action Action) error {
+func VerifyBucketPolicy(policy []byte, access, bucket, object string, actions ...Action) error {
+	if len(actions) == 0 {
+		return s3err.GetAPIError(s3err.ErrAccessDenied)
+	}
+
 	var bucketPolicy BucketPolicy
 	if err := json.Unmarshal(policy, &bucketPolicy); err != nil {
 		return fmt.Errorf("failed to parse the bucket policy: %w", err)
@@ -245,8 +249,10 @@ func VerifyBucketPolicy(policy []byte, access, bucket, object string, action Act
 		resource += "/" + object
 	}
 
-	if !bucketPolicy.isAllowed(access, action, resource) {
-		return s3err.GetAPIError(s3err.ErrAccessDenied)
+	for _, action := range actions {
+		if !bucketPolicy.isAllowed(access, action, resource) {
+			return s3err.GetAPIError(s3err.ErrAccessDenied)
+		}
 	}
 
 	return nil
--- a/s3api/controllers/object-post.go
+++ b/s3api/controllers/object-post.go
@@ -48,7 +48,7 @@ func (c S3ApiController) RestoreObject(ctx *fiber.Ctx) (*Response, error) {
 			Acc:             acct,
 			Bucket:          bucket,
 			Object:          key,
-			Action:          auth.RestoreObjectAction,
+			Actions:         []auth.Action{auth.RestoreObjectAction},
 			IsPublicRequest: isBucketPublic,
 			DisableACL:      c.disableACL,
 		})
@@ -100,7 +100,7 @@ func (c S3ApiController) SelectObjectContent(ctx *fiber.Ctx) (*Response, error) 
 			Acc:             acct,
 			Bucket:          bucket,
 			Object:          key,
-			Action:          auth.GetObjectAction,
+			Actions:         []auth.Action{auth.GetObjectAction},
 			IsPublicRequest: isBucketPublic,
 			DisableACL:      c.disableACL,
 		})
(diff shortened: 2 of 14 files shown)


auth/bucket_policy.go
```diff
@@ -234,7 +234,7 @@ func ValidatePolicyDocument(policyBin []byte, bucket string, iam IAMService) err
 	return nil
 }
 
-func VerifyBucketPolicy(policy []byte, access, bucket, object string, action Action) error {
+func VerifyBucketPolicy(policy []byte, access, bucket, object string, actions ...Action) error {
 	var bucketPolicy BucketPolicy
```
benmcclelland, line 242: should we add a check in here to make sure actions slice not nil?
```go
    if len(actions) == 0 {
        return s3err.GetAPIError(s3err.ErrAccessDenied)
    }
```

s3api/controllers/object-put.go
```diff
@@ -660,6 +662,17 @@ func (c S3ApiController) PutObject(ctx *fiber.Ctx) (*Response, error) {
...
 	}
 
+	actions := []auth.Action{auth.PutObjectAction}
+	if tagging != "" {
+		actions = append(actions, auth.PutObjectTaggingAction)
+	}
+	if legalHoldHdr != "" {
+		actions = append(actions, auth.PutObjectLegalHoldAction)
+	}
+	if lockModeHdr != "" {
```
benmcclelland: should we also have a check if `X-Amz-Object-Lock-Retain-Until-Date` set?

s3api/controllers/object-post.go
```diff
@@ -154,11 +154,25 @@ func (c S3ApiController) CreateMultipartUpload(ctx *fiber.Ctx) (*Response, error
...
 	parsedAcl := utils.ContextKeyParsedAcl.Get(ctx).(auth.ACL)
 
+	actions := []auth.Action{auth.PutObjectAction}
+	if tagging != "" {
+		actions = append(actions, auth.PutObjectTaggingAction)
+	}
+	if legalHoldHdr != "" {
+		actions = append(actions, auth.PutObjectLegalHoldAction)
+	}
+	if lockModeHdr != "" {
```
benmcclelland: same question here if we need to consider `X-Amz-Object-Lock-Retain-Until-Date`

s3api/controllers/object-put.go
```diff
@@ -508,7 +508,7 @@ func (c S3ApiController) CopyObject(ctx *fiber.Ctx) (*Response, error) {
 			Acc:           acct,
 			Bucket:        bucket,
 			Object:        key,
-			Action:        auth.PutObjectAction,
+			Actions:       []auth.Action{auth.PutObjectAction},
```
benmcclelland: are these checks needed for copy object?
```go
	if tagging != "" {
		actions = append(actions, auth.PutObjectTaggingAction)
	}
	if legalHoldHdr != "" {
		actions = append(actions, auth.PutObjectLegalHoldAction)
	}
	if lockModeHdr != "" {
		actions = append(actions, auth.PutObjectRetentionAction)
	}
```

tests/integration/Access_Control.go
```diff
@@ -453,3 +454,289 @@ func AccessControl_copy_object_with_starting_slash_for_user(s *S3Conf) error {
...
+			Bucket:                    &bucket,
+			Key:                       &obj,
+			ObjectLockMode:            types.ObjectLockModeGovernance,
+			ObjectLockRetainUntilDate: &date,
+		})
+		cancel()
+		if err != nil {
+			return err
+		}
+		return err
```
benmcclelland, line 824: simplify to just `return err`

tests/integration/Access_Control.go
```diff
@@ -453,3 +454,556 @@ func AccessControl_copy_object_with_starting_slash_for_user(s *S3Conf) error {
...
+			return err
+		}
+
+		userClient := s.getUserClient(testuser)
+		ctx, cancel := context.WithTimeout(context.Background(), shortTimeout)
+		_, err = userClient.CopyObject(ctx, &s3.CopyObjectInput{
+			Bucket:                    &bucket,
+			Key:                       &dstObj,
+			CopySource:                getPtr(fmt.Sprintf("%s/%s", bucket, srcObj)),
+			ObjectLockLegalHoldStatus: types.ObjectLockLegalHoldStatusOn,
```
benmcclelland: I think this test should use governance instead?:
ObjectLockMode: types.ObjectLockModeGovernance
ObjectLockRetainUntilDate: ...

niksis02: Ahh, right. Thanks for noticing.

benmcclelland (approved):

PR #2012, opened 2026-04-03
fix: fix azure multipart upload objects masking

niksis02:
As multipart uploads are translated to blobs in azure blob storage, they were visible in ListObjects(V2) as complete objects. Now the blobs with multipart prefix are filtered out during listing.

The listing logic is rewritten client-side to implement proper S3 semantics: flat blob enumeration with manual delimiter handling, correct truncation (IsTruncated only set when more items genuinely exist beyond maxKeys), and StartAfter/Marker/ContinuationToken applied via the lexicographic max of both constraints in ListObjectsV2.

For the same reason bucket deletion was not allowed. Now multipart objects are explicitly checked on bucket deletion and any pending multipart upload doesn't block the bucket deletion anymore.

### DIFF
--- a/backend/azure/azure.go
+++ b/backend/azure/azure.go
@@ -304,15 +304,21 @@ func (az *Azure) HeadBucket(ctx context.Context, input *s3.HeadBucketInput) (*s3
 func (az *Azure) DeleteBucket(ctx context.Context, bucket string) error {
 	pager := az.client.NewListBlobsFlatPager(bucket, nil)
 
-	pg, err := pager.NextPage(ctx)
-	if err != nil {
-		return azureErrToS3Err(err)
+	for pager.More() {
+		pg, err := pager.NextPage(ctx)
+		if err != nil {
+			return azureErrToS3Err(err)
+		}
+
+		for _, item := range pg.Segment.BlobItems {
+			// ignore temp multipart objects when determining if bucket non-empty
+			if !strings.HasPrefix(backend.GetStringFromPtr(item.Name), string(metaTmpMultipartPrefix)) {
+				return s3err.GetAPIError(s3err.ErrBucketNotEmpty)
+			}
+		}
 	}
 
-	if len(pg.Segment.BlobItems) > 0 {
-		return s3err.GetAPIError(s3err.ErrBucketNotEmpty)
-	}
-	_, err = az.client.DeleteContainer(ctx, bucket, nil)
+	_, err := az.client.DeleteContainer(ctx, bucket, nil)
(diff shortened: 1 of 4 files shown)


benmcclelland:
In the loop fetching pages, the no-StartAfter path exits after the first page. If a page contains only hidden multipart entries, the response can still be empty while real objects exist on later pages. We might need to continue paging until collected visible results (or exhaust pages), and count only visible entries against MaxKeys.

The tmp multipart namespace can still leak via CommonPrefixes. With delimiter-based listing, users can still see .sgwtmp prefixes I think.

Is there a way to add a prefix filter to mask sgwtmp in the backend request?

backend/azure/azure.go
```diff
@@ -659,92 +665,123 @@ func (az *Azure) ListObjects(ctx context.Context, input *s3.ListObjectsInput) (s
...
+				cpSet[cpKey] = struct{}{}
+				candidateMarker = cpKey
+				totalFound++
+				if totalFound == maxKeys {
+					pastMax = true
+				}
+			} else {
+				if pastMax {
+					isTruncated = true
+					break
```
benmcclelland: I think this needs to break the outer loop here?

backend/azure/azure.go
```diff
@@ -659,92 +665,123 @@ func (az *Azure) ListObjects(ctx context.Context, input *s3.ListObjectsInput) (s
...
+					continue
+				}
+				// Deduplicate: multiple blobs can map to the same common prefix
+				if _, exists := cpSet[cpKey]; exists {
+					continue
+				}
+				// If we already reached maxKeys, this new unique CP means truncation
+				if pastMax {
+					isTruncated = true
+					break
```
benmcclelland: also break outer loop here?

backend/azure/azure.go
```diff
@@ -304,15 +304,21 @@ func (az *Azure) HeadBucket(ctx context.Context, input *s3.HeadBucketInput) (*s3
...
+	for pager.More() {
+		pg, err := pager.NextPage(ctx)
+		if err != nil {
+			return azureErrToS3Err(err)
+		}
 
-	if len(pg.Segment.BlobItems) > 0 {
-		return s3err.GetAPIError(s3err.ErrBucketNotEmpty)
+		for _, item := range pg.Segment.BlobItems {
+			// the bucket is not empty, if any no multipart upload object is present
```
benmcclelland: slightly confusing wording here, maybe something like:
// ignore temp multipart objects when determining if bucket non-empty

benmcclelland (approved):

PR #2054, opened 2026-04-16
fix: make CompleteMultipartUpload idempotent and add part-number support to GetObject/HeadObject

### LINKED ISSUE #1064: [Feature] - return success for successive complete multipart upload calls
The `HeadObject` and `GetObjectAttributes` actions return some data (e.g., `PartsCount`) about completed multipart uploads.  

A `CompleteMultipartUpload` call with the same `key` and `uploadId` on a completed multipart upload returns a successful response instead of throwing an error (as the gateway does). This suggests that the metadata of completed multipart uploads is being stored.  

The goal is to design and implement a persistent/temporary storage solution for completed multipart upload metadata. One option is to store this metadata in the extended attributes (`xattrs`) of the created object.

niksis02:
Closes #1064

Use the multipart ETag as the in-progress directory suffix instead of the static `.inprogress` marker so that concurrent CompleteMultipartUpload calls for the same upload ID are all treated as successful (idempotent) rather than racing, where only one succeeded and the rest returned NoSuchUpload.

After finalizing the multipart upload, store an `mp-metadata` xattr on the assembled object that records the upload ID and cumulative byte offsets for each part. GetObject and HeadObject now use this metadata to serve individual part ranges via the `partNumber` query parameter, returning a successful response instead of returning NotImplemented.

Add two new S3 error codes:
- `ErrInvalidPartNumberRange` (416 RequestedRangeNotSatisfiable) — returned when the requested part number exceeds the number of parts in the upload.
- `ErrRangeAndPartNumber` (400 BadRequest) — returned when both a Range header and a partNumber query parameter are specified on the same request.

### DIFF
--- a/backend/azure/azure.go
+++ b/backend/azure/azure.go
@@ -73,6 +73,10 @@ const (
 	// keyMpZeroBytesParts tracks zero-byte upload parts in the sgwtmp metadata.
 	// Azure StageBlock rejects Content-Length: 0, so zero-byte parts are stored here.
 	keyMpZeroBytesParts key = "Zerobytesparts"
+	// keyMpMetadata stores multipart upload part-offset metadata on the final
+	// committed blob so that GetObject/HeadObject can serve individual parts
+	// by part-number.
+	keyMpMetadata key = "Mpmetadata"
 
 	defaultListingMaxKeys = 1000
 )
@@ -89,6 +93,7 @@ func (key) Table() map[string]struct{} {
 		"objectlegalhold":   {},
 		"objname":           {},
 		".sgwtmp/multipart": {},
+		"mpmetadata":        {},
 	}
 }
 
@@ -457,11 +462,6 @@ func (az *Azure) DeleteBucketTagging(ctx context.Context, bucket string) error {
 }
 
 func (az *Azure) GetObject(ctx context.Context, input *s3.GetObjectInput) (*s3.GetObjectOutput, error) {
-	if input.PartNumber != nil {
-		// querying an object with part number is not supported
-		return nil, s3err.GetAPIError(s3err.ErrNotImplemented)
-	}
-
 	client, err := az.getBlobClient(*input.Bucket, *input.Key)
 	if err != nil {
 		return nil, err
@@ -486,8 +486,53 @@ func (az *Azure) GetObject(ctx context.Context, input *s3.GetObjectInput) (*s3.G
 	}
 
 	var opts *azblob.DownloadStreamOptions
-	if *input.Range != "" {
-		offset, count, isValid, err := backend.ParseObjectRange(*resp.ContentLength, *input.Range)
+	var partsCount *int32
+	var contentRange *string
+
+	if input.PartNumber != nil {
+		// Serve a specific part if the object has multipart upload metadata.
+		// For non-multipart objects (no mp-metadata), partNumber=1 returns the
+		// full object with no Content-Range; any other partNumber is out of range.
+		if mpMetaStr, ok := resp.Metadata[string(keyMpMetadata)]; ok && mpMetaStr != nil {
+			var mpMeta backend.MpUploadMetadata
+			if err := json.Unmarshal([]byte(*mpMetaStr), &mpMeta); err != nil {
+				return nil, fmt.Errorf("parse object multipart metadata: %w", err)
+			}
+
+			partNum := *input.PartNumber
+			totalParts := int32(len(mpMeta.Parts))
+			partsCount = &totalParts
+			if partNum > totalParts {
+				return nil, s3err.GetAPIError(s3err.ErrInvalidPartNumberRange)
+			}
+
+			var startOffset int64
+			if partNum > 1 {
+				startOffset = mpMeta.Parts[partNum-2]
+			}
+			length := mpMeta.Parts[partNum-1] - startOffset
+			var objSize int64
+			if resp.ContentLength != nil {
+				objSize = *resp.ContentLength
+			}
+			contentRange = backend.GetPtrFromString(fmt.Sprintf("bytes %d-%d/%d", startOffset, startOffset+length-1, objSize))
+			opts = &azblob.DownloadStreamOptions{
+				Range: blob.HTTPRange{
+					Offset: startOffset,
+					Count:  length,
+				},
+			}
+		} else if *input.PartNumber > 1 {
+			return nil, s3err.GetAPIError(s3err.ErrInvalidPartNumberRange)
+		}
+		// partNumber=1 on a non-multipart object: fall through and serve the
+		// full object without a range (opts remains nil, contentRange stays nil).
+	} else if *input.Range != "" {
+		var objSize int64
+		if resp.ContentLength != nil {
+			objSize = *resp.ContentLength
+		}
+		offset, count, isValid, err := backend.ParseObjectRange(objSize, *input.Range)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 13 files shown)


backend/posix/posix.go
```diff
@@ -4370,15 +4404,49 @@ func (p *Posix) GetObject(ctx context.Context, input *s3.GetObjectInput) (*s3.Ge
...
+				return nil, s3err.GetAPIError(s3err.ErrInvalidPartNumberRange)
+			}
+
+			// Parts holds cumulative sizes: Parts[i] = sum of sizes 1..i+1
+			if partNum > 1 {
+				startOffset = mpMeta.Parts[partNum-2]
+			}
+			length = mpMeta.Parts[partNum-1] - startOffset
+			contentRange = fmt.Sprintf("bytes %d-%d/%d", startOffset, startOffset+length-1, objSize)
+		}
```
benmcclelland, line 4451: missing handling of non multipart upload objects here. if part number specified but there is no mpMetaKey because uploaded with put-object instead of multipart upload, aws will download the whole file if part-number is 1, otherwise it returns
An error occurred (InvalidPartNumber) when calling the GetObject operation: The requested partnumber is not satisfiable

backend/azure/azure.go
```diff
@@ -486,8 +486,47 @@ func (az *Azure) GetObject(ctx context.Context, input *s3.GetObjectInput) (*s3.G
...
+				objSize = *resp.ContentLength
+			}
+			contentRange = fmt.Sprintf("bytes %d-%d/%d", startOffset, startOffset+length-1, objSize)
+			opts = &azblob.DownloadStreamOptions{
+				Range: blob.HTTPRange{
+					Offset: startOffset,
+					Count:  length,
+				},
+			}
+		}
```
benmcclelland, line 527: also need the check here for partnumber != 1 for a non-multipart upload. and if partnumber == 1, set to full object size.

backend/azure/azure.go
```diff
@@ -523,18 +564,14 @@ func (az *Azure) GetObject(ctx context.Context, input *s3.GetObjectInput) (*s3.G
 		LastModified:       blobDownloadResponse.LastModified,
 		Metadata:           parseAndFilterAzMetadata(blobDownloadResponse.Metadata),
 		TagCount:           &tagcount,
-		ContentRange:       blobDownloadResponse.ContentRange,
+		ContentRange:       &contentRange,
```
benmcclelland: I think this needs to be nil for non-range requests

backend/common.go
```diff
@@ -406,6 +406,16 @@ func md5String(data []byte) string {
...
 }
 
+// MpUploadMetadata is stored alongside the final object after a multipart
+// upload completes. It records the uploadId and the cumulative byte offsets
+// of each part: Parts[i] = sum of sizes of parts 1..i+1. This allows O(1)
+// total size lookup (Parts[last]) and O(1) part-start/size derivation for
+// GetObject/HeadObject part-number requests.
+type MpUploadMetadata struct {
+	UploadID string  `json:"uploadId"`
+	Parts    []int64 `json:"Parts"`
```
benmcclelland: the json tag should be lowercase "parts" here

backend/posix/posix.go
```diff
@@ -1690,25 +1690,66 @@ func (p *Posix) CompleteMultipartUploadWithCopy(ctx context.Context, input *s3.C
...
+	// the same ETag, so it will either find the directory still present (still
+	// processing) or gone (already completed) and react accordingly.
+	sum := sha256.Sum256([]byte(object))
 	objdirFull := filepath.Join(bucket, MetaTmpMultipartDir, fmt.Sprintf("%x", sum))
 	uploadIDDir := filepath.Join(objdirFull, uploadID)
-	activeUploadName := uploadID + inProgressSuffix
+	// Calculate s3 compatible md5sum for complete multipart.
+	s3MD5 := backend.GetMultipartMD5(parts)
+	activeUploadName := uploadID + strings.Trim(s3MD5, "\"")
 	uploadIDInProgress := filepath.Join(objdirFull, activeUploadName)
```
benmcclelland, line 1707: line 2567 is looking for a "." in the name to know which uploads to skip in teh listing, but I dont think we have one in there now. Should we keep a suffix on here with a "." so that we can trim these from the listing easier?

tests/integration/utils.go
```diff
@@ -2700,3 +2700,16 @@ func sendPostObject(input PostRequestConfig) (*http.Response, error) {
 
 	return input.s3Conf.httpClient.Do(req)
 }
+
+func getEtagBytes(etag string) []byte {
+	decode, err := hex.DecodeString(strings.ReplaceAll(etag, string('"'), ""))
+	if err != nil {
+		return []byte(etag)
```
benmcclelland: nit: this might be problematic in future uses if its falling back to just []byte(etag) here, might be better to return or log an error here?

backend/posix/posix.go
```diff
@@ -4569,15 +4630,45 @@ func (p *Posix) HeadObject(ctx context.Context, input *s3.HeadObjectInput) (*s3.
...
-	startOffset, length, isValid, err := backend.ParseObjectRange(size, getString(input.Range))
-	if err != nil {
-		return nil, err
-	}
-
 	var contentRange string
-	if isValid {
-		contentRange = fmt.Sprintf("bytes %v-%v/%v",
-			startOffset, startOffset+length-1, size)
+	var startoffset, length int64
```
benmcclelland: minor nit, make startoffset camel case here for consisntency

```suggestion
	var startOffset, length int64
```

benmcclelland (changes requested):

backend/posix/posix.go
niksis02: Good catch, I haven't even tested this case

backend/posix/posix.go
```diff
@@ -4403,11 +4483,8 @@ func (p *Posix) GetObject(ctx context.Context, input *s3.GetObjectInput) (*s3.Ge
 	}
 
 	// using an os.File allows zero-copy sendfile via io.Copy(os.File, net.Conn)
-	var body io.ReadCloser = f
-	if startOffset != 0 || length != objSize {
-		rdr := io.NewSectionReader(f, startOffset, length)
-		body = &backend.FileSectionReadCloser{R: rdr, F: f}
-	}
+	rdr := io.NewSectionReader(f, startOffset, length)
+	body := &backend.FileSectionReadCloser{R: rdr, F: f}
```
benmcclelland, line 4487: it looks like we lost the zero-copy sendfile case here for when a range is not requested.  for this to work, we need to set body to f when we dont need the range reader.

benmcclelland (changes requested):

backend/posix/posix.go
```diff
@@ -4408,6 +4488,8 @@ func (p *Posix) GetObject(ctx context.Context, input *s3.GetObjectInput) (*s3.Ge
 		rdr := io.NewSectionReader(f, startOffset, length)
 		body = &backend.FileSectionReadCloser{R: rdr, F: f}
 	}
+	rdr := io.NewSectionReader(f, startOffset, length)
+	body = &backend.FileSectionReadCloser{R: rdr, F: f}
```
benmcclelland: why is this needed? this is already done in the conditional above. and body should just be f if not reading from a range.

benmcclelland (changes requested):

backend/posix/posix.go
niksis02: My bad (. Thanks for noticing.

benmcclelland (approved):

PR #2061, opened 2026-04-20
fix: properly encode path and query parameters.

### LINKED ISSUE #2060: [Bug] - WebUI: Invalid encoding in signature computation
**Describe the bug**

It seems that the `signRequest()` method in `api.js` incorrectly encodes the values included in the query string, leaving some unencoded characters that subsequently fail validation.

**To Reproduce**

Create a directory named, for example, `invalid)` (with a parenthesis) and try to access it. The response will be 403 because of an invalid signature.

This also makes such a directory undeletable from the WebUI.

The WebUI uses `encodeURIComponent()` to encode values used in the query string. This is not sufficient, since it leaves some characters unencoded, as far as AWS defines them.

**Expected behavior**

We should be able to access the newly created directory.

**Server Version**

v1.4.0

**Additional context**

My example gives both parentheses as problematic, but really, the issue most probably also appears with other characters that `encodeURIComponent()` ignores, such as `!`, `'`, or `*`.

_EDIT:_ from MDN, the missing, unencoded characters would be `! ~ * ' ( )`.

If requested, I can submit a pull request that could try and fix this.

apognu:
Fixes #2060.

[AWS Signature V4](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html) defines that all characters except for `[A-Za-z0-9-_.~]` should be hex-encoded. The current use of `encodeURIComponent()` misses a few characters, which would make signatures for those prefixes invalid.

This PR extends that method with the proper encoding for the mising characters.

---

Excerpt from the documentation:

```
URI encode every byte. UriEncode() must enforce the following rules:

    URI encode every byte except the unreserved characters: 'A'-'Z', 'a'-'z', '0'-'9', '-', '.', '_', and '~'.

    The space character is a reserved character and must be encoded as "%20" (and not as "+").

    Each URI encoded byte is formed by a '%' and the two-digit hexadecimal value of the byte.

    Letters in the hexadecimal value must be uppercase, for example "%1A".

    Encode the forward slash character, '/', everywhere except in the object key name. For example, if the object key name is photos/Jan/sample.jpg, the forward slash in the key name is not encoded. 
```

### DIFF
--- a/webui/web/js/api.js
+++ b/webui/web/js/api.js
@@ -109,7 +109,7 @@ const hasSubtleCrypto = typeof crypto !== 'undefined' && typeof crypto.subtle !=
  */
 function encodeS3Key(key) {
   if (!key) return '';
-  return key.split('/').map(segment => encodeURIComponent(segment)).join('/');
+  return key.split('/').map(segment => awsUriEncode(segment)).join('/');
 }
 
 class VersityAPI {
@@ -161,7 +161,7 @@ class VersityAPI {
     // Sort query params for canonical request
     const sortedParams = [...url.searchParams.entries()].sort((a, b) => a[0].localeCompare(b[0]));
     const canonicalQueryString = sortedParams
-      .map(([k, v]) => `${encodeURIComponent(k)}=${encodeURIComponent(v)}`)
+      .map(([k, v]) => `${awsUriEncode(k)}=${awsUriEncode(v)}`)
       .join('&');
 
     const canonicalHeaders = `host:${host}\n`;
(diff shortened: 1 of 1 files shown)


versity-github:

webui/web/js/api.js
```diff
@@ -2059,5 +2059,17 @@ ${tagsXml}
   }
 }
 
+/**
+ * Encode path and query params as defined as the AWS SigV4 specification: any
+ * character other than A-Za-z0-9-_. should be hex-encoded.
+ */
+function awsUriEncode(text) {
```
benmcclelland:
```suggestion
/**
 * URI-encode a string per the AWS SigV4 specification.
 * All characters except the unreserved set (A-Za-z0-9 - _ . ~) are percent-encoded,
 * with hexadecimal digits in uppercase (e.g. %2F, not %2f).
 *
 * @param {string} text - The string to encode.
 * @returns {string} The percent-encoded string.
 */
 function awsUriEncode(text) {
```

benmcclelland (changes requested):
suggestions for jsdoc and using  codePointAt instead of charCodeAt

benmcclelland (approved):

PR #2062, opened 2026-04-20
test: test tagging, shellcheck fix, speedup

lrm25:


tests/tags/get_tests.sh
```diff
@@ -0,0 +1,150 @@
...
+}
+
+run_tests() {
+  if ! check_param_count_v2 "test lines" 1 $#; then
+    return 1
+  fi
+
+  local i=0 files=() names=() tags=()
+
+  while IFS=$'\t' read -r loc t name; do
```
anodelman: add || [ -n "$loc" ] to the while condition to ensure the very last test in the list isn't skipped.


`while IFS=$'\t' read -r loc t name || [ -n "$loc" ]; do`

tests/tags/list_tags.py
```diff
@@ -0,0 +1,139 @@
...
+            print(f"\t{tag}: {desc}")
+
+
+def print_tag_description(data: dict, tag: str) -> int:
+    tags = data.get("tags", {})
+    for g in tags:
+        group_tags = tags.get(g, {}) or {}
+        g_lower = {k.lower(): v for k, v in group_tags.items()}
+        if tag.lower() in g_lower:
+            desc = (group_tags.get(tag, {}) or {}).get("desc", "")
```
anodelman: If the user types listbuckets (lowercase) but the YAML has ListBuckets, tag.lower() in g_lower will be true, but group_tags.get(tag) will return None.

```
tags = data.get("tags", {})
    target = tag.lower()
    for group_name, group_tags in tags.items():
        for tag_key, tag_info in group_tags.items():
            if tag_key.lower() == target:
                desc = tag_info.get("desc", "")
                print(f"{tag_key}: {desc}")
                return 0
    return 1
```

anodelman (changes requested):
Just a couple of minor fixes.

anodelman (approved):

PR #2063, opened 2026-04-20
feat: history back/forward actions on explorer and modals.

apognu:
Most actions within each page is stateless (show modals) or change the URL hash. As it is, those are not tracked and using the back button has no effect.

This commits implements two things:

 - Tracking of the URL hash in the explorer to move from bucket/folders on history change.
 - Add a history state when a modal is open, so the back button closes the modal.

---

**Note:** The modal behavior, as implemented above, leaves behind an "empty" forward state when closing a modal (either through a button, or by using back). This means that, on modal close, the user is able to go one page forward, with no effect.

I did not find a way to prevent this, and figured it was still a good feature to have and what users expect.

### DIFF
--- a/webui/web/buckets.html
+++ b/webui/web/buckets.html
@@ -221,7 +221,7 @@ under the License.
   </div>
 
   <!-- Change Owner Modal -->
-  <div id="owner-modal" class="hidden fixed inset-0 z-50">
+  <div id="owner-modal" class="modal hidden fixed inset-0 z-50">
     <div class="modal-backdrop absolute inset-0" onclick="closeModal('owner-modal')"></div>
     <div class="absolute inset-0 flex items-center justify-center p-4">
       <div class="bg-white rounded-xl shadow-2xl w-full max-w-md relative">
@@ -281,7 +281,7 @@ under the License.
   </div>
 
   <!-- Create Bucket Modal -->
-  <div id="create-bucket-modal" class="hidden fixed inset-0 z-50">
+  <div id="create-bucket-modal" class="modal hidden fixed inset-0 z-50">
     <div class="modal-backdrop absolute inset-0" onclick="closeModal('create-bucket-modal')"></div>
     <div class="absolute inset-0 flex items-center justify-center p-4">
       <div class="bg-white rounded-xl shadow-2xl w-full max-w-md relative">
@@ -353,6 +353,15 @@ under the License.
     let allBuckets = [];
     let allUsers = [];
     let selectedBucket = null;
+    
+    // Catch the back button to close the open modal, if any is open.
+    window.addEventListener('popstate', (e) => {
+      Array.from(document.getElementsByClassName('modal')).forEach((modal) => {
+        if (!modal.classList.contains('hidden')) {
+          closeModal(modal.getAttribute('id'), false);
+        }
+      })
+    });
 
     // ============================================
     // Custom Dropdown Functions
--- a/webui/web/buckets.html
+++ b/webui/web/buckets.html
@@ -354,15 +354,6 @@ under the License.
     let allUsers = [];
     let selectedBucket = null;
     
-    // Catch the back button to close the open modal, if any is open.
-    window.addEventListener('popstate', (e) => {
-      Array.from(document.getElementsByClassName('modal')).forEach((modal) => {
-        if (!modal.classList.contains('hidden')) {
-          closeModal(modal.getAttribute('id'), false);
-        }
-      })
-    });
-
     // ============================================
     // Custom Dropdown Functions
     // ============================================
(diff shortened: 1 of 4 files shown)


versity-github:

webui/web/js/app.js
```diff
@@ -129,13 +129,23 @@ function openModal(modalId) {
...
-function closeModal(modalId) {
+// Close the currently opened modal, manually popping the modal
+// history state if the modal was closed manually (default). In the
+// case where the modal was closed due to navigating back, the state
+// is already popped and we can skip it.
+function closeModal(modalId, popState = true) {
   const modal = document.getElementById(modalId);
   if (modal) {
     modal.classList.add('hidden');
+    if (popState) {
```
benmcclelland, line 146:
```suggestion
    if (popState && history.state?.modal) {
```

If closeModal is ever called without a corresponding openModal, history.back() will navigate the user away from the page entirely. The guard should check that the current history.state actually corresponds to a modal before calling back():

webui/web/js/app.js
```diff
@@ -129,13 +129,23 @@ function openModal(modalId) {
...
+      history.back();
```
benmcclelland, line 147: history.back() returns immediately but the popstate event fires later. If the user quickly opens another modal after closing one, the deferred popstate will fire and close the new modal unexpectedly. Maybe add a guard to protect against htis case:
```javascript
let navigatingBack = false;

function closeModal(modalId, popState = true) {
  const modal = document.getElementById(modalId);
  if (modal) {
    modal.classList.add('hidden');
    if (popState && history.state?.modal) {
      navigatingBack = true;
      history.back();
    }
  }
}

// In the popstate handler:
window.addEventListener('popstate', (e) => {
  if (navigatingBack) { navigatingBack = false; return; }
  // ... close open modals
});
```

webui/web/explorer.html
```diff
@@ -998,6 +998,9 @@ <h2 class="text-xl font-semibold text-charcoal">Multipart Uploads</h2>
     if (!requireAuth()) {
       // Will redirect to login
     } else {
+      // Clear history state on startup (can happen when reloading with popup open)
+      if (history.state) history.back();
```
benmcclelland, line 1002:
```suggestion
      if (history.state?.modal) history.back();
```

history.state can be non-null for reasons unrelated to modals (e.g. the browser restoring state after a normal reload). This would silently trigger an unintended back-navigation. It should be scoped to modal state specifically

webui/web/explorer.html
```diff
@@ -1021,10 +1024,27 @@ <h2 class="text-xl font-semibold text-charcoal">Multipart Uploads</h2>
...
+    });
+    
+    // Catch the back button to close the open modal, if any is open.
+    window.addEventListener('popstate', (e) => {
+      Array.from(document.getElementsByClassName('modal')).forEach((modal) => {
+        if (!modal.classList.contains('hidden')) {
+          closeModal(modal.getAttribute('id'), false);
+        }
+      })
+    });
```
benmcclelland, line 1048: When the user navigates back through a hash change, both hashchange and popstate are dispatched. This means init() could run twice. The popstate handler should check if the navigation was hash-based (i.e., no modal state) and skip the modal-close logic, or the hashchange listener should be gated to prevent double-execution.

webui/web/buckets.html
```diff
@@ -353,6 +353,15 @@ <h2 class="text-xl font-semibold text-charcoal">Create Bucket</h2>
...
     let selectedBucket = null;
+    
+    // Catch the back button to close the open modal, if any is open.
+    window.addEventListener('popstate', (e) => {
+      Array.from(document.getElementsByClassName('modal')).forEach((modal) => {
+        if (!modal.classList.contains('hidden')) {
+          closeModal(modal.getAttribute('id'), false);
+        }
+      })
+    });
```
benmcclelland, line 365: Move all of these duplicated handlers (from buckets, explorer, and users) to app.js:
```javascript
window.addEventListener('popstate', () => {
  document.querySelectorAll('.modal:not(.hidden)').forEach((modal) => {
    closeModal(modal.id, false);
  });
});
```

webui/web/explorer.html
apognu: Can you expand a bit on this? I don't see how `init()` can run a second time because of the `popstate` handler?

It is only called in `hashchange`, which arguably is what we want, because if the hash changes, it means the prefix changed and we want to reload the objects. And if the prefix changes, we would want the modal to be closed because the context changed.

Happy to change it, but there must be something I am not seeing.

Pushing a commit for your other comments.

benmcclelland: actually I was mistaken about the double init(), this looks correct

benmcclelland (approved):

PR #2066, opened 2026-04-21
feat: add server-side pagination for webui object explorer

### LINKED ISSUE #2055: [Feature] - <suggest more than 1000 on webUI listingPage, eg: VGW_S3_MAX_KEYS=2000>

suggest more than 1000 on webUI listingPage, eg: VGW_S3_MAX_KEYS=2000
or list objects by page

benmcclelland:
The object listing now fetches pages from S3 directly using ListObjectsV2 continuation tokens rather than loading all objects at once. Users can navigate forward and back with first/prev/next buttons and choose how many rows to show per page (10/20/50/100/1000, defaulting to 10), which keeps the listing fast and responsive even in buckets with thousands of objects.

Pagination resets automatically when navigating into a folder, running a search, or changing page size. While a search is active, forward navigation is disabled since search filters within the current page only, and the item count shows "(filtered)" to make that clear. When versioning is enabled, delete-marker rows are scoped to the current page's key range so they don't bleed in from adjacent pages.

Fixes #2055

### DIFF
--- a/webui/web/explorer.html
+++ b/webui/web/explorer.html
@@ -285,6 +285,42 @@ under the License.
                   </tbody>
                 </table>
               </div>
+              <!-- Pagination Bar -->
+              <div id="objects-pagination" class="hidden flex items-center justify-between px-4 py-3 border-t border-gray-100">
+                <div class="flex items-center gap-2">
+                  <span class="text-sm text-charcoal-300" id="pagination-info"></span>
+                </div>
+                <div class="flex items-center gap-4">
+                  <div class="flex items-center gap-2">
+                    <span class="text-sm text-charcoal-300">Rows per page:</span>
+                    <select id="page-size-select" onchange="changePageSize(parseInt(this.value, 10))" class="text-sm border border-gray-200 rounded-lg px-2 py-1.5 text-charcoal focus:outline-none focus:border-accent bg-white">
+                      <option value="10" selected>10</option>
+                      <option value="20">20</option>
+                      <option value="50">50</option>
+                      <option value="100">100</option>
+                      <option value="1000">1000</option>
+                    </select>
+                  </div>
+                  <div class="flex items-center gap-1">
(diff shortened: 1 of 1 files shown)


webui/web/explorer.html
```diff
@@ -1510,8 +1577,15 @@ <h2 class="text-xl font-semibold text-charcoal">Multipart Uploads</h2>
         return;
       }
 
+      // Items are already server-paginated via maxKeys; just alias for rendering
+      const pagedFolders = filteredFolders;
+      const pagedObjects = filteredObjects;
+      const pagedDeletedObjects = filteredDeletedObjects;
```
niksis02, line 1583: Unnecessary aliasing ? These are pure reassignments with no transformation. It would be better directly using `filteredFolders`, `filteredObjects`, `filteredDeletedObjects` ?

webui/web/explorer.html
```diff
@@ -2880,6 +3006,7 @@ <h2 class="text-xl font-semibold text-charcoal">Multipart Uploads</h2>
       if (searchInput) {
         searchInput.addEventListener('input', debounce((e) => {
           searchTerm = e.target.value.toLowerCase();
+          currentPage = 1;
           const clearBtn = document.getElementById('clear-search');
           if (searchTerm) {
             clearBtn.classList.remove('hidden');
```
niksis02, line 3012: When a user is on page N > 1 and types a search query, currentPage is reset to 1 for display but loadObjects() is never called — the in-memory data remains page N's content. This means the pagination bar shows "Page 1 (filtered)" while actually filtering over page N's objects. The problem compounds when the user clears the search: clearSearch() calls renderObjects() without resetting currentPage, so the bar now shows "Page 1" with page N's unfiltered data, and clicking "Next" loads continuationTokens[1] (page 2's token) instead of continuing from page N — silently jumping the user backward in the listing. The fix is either to call loadObjects() (no token) when search is applied from page N > 1 so a true page-1 fetch occurs, or to not reset currentPage at all in the search handler since no data reload happens and the current page label stays accurate.

niksis02 (approved):

PR #2076, opened 2026-04-22
fix: add explicit sidecar metadata cleanup on object/bucket deletion

benmcclelland:
Unlike xattr metadata which is tied to the filesystem object and removed automatically, sidecar metadata lives in a parallel directory tree and must be deleted explicitly. Add DeleteAttributes calls after removing bucket directories, null-version files, and versioned object files.

Also add an os.Stat existence check in GetObjectTagging and PutObjectTagging when no versionId is given, since sidecar's StoreAttribute/RetrieveAttribute do not naturally return ErrNotExist for missing objects the way xattr operations do.

### DIFF
--- a/backend/meta/sidecar.go
+++ b/backend/meta/sidecar.go
@@ -145,15 +145,24 @@ func (s SideCar) ListAttributes(bucket, object string) ([]string, error) {
 }
 
 // DeleteAttributes removes all attributes for an object or a bucket.
+// When object is empty the entire bucket sidecar directory is removed,
+// cleaning up any orphaned object or multipart metadata within it.
 func (s SideCar) DeleteAttributes(bucket, object string) error {
-	metadir := filepath.Join(s.dir, bucket, object, sidecarmeta)
 	if object == "" {
-		metadir = filepath.Join(s.dir, bucket, sidecarmeta)
+		// Remove the entire bucket sidecar directory so that orphaned
+		// object/multipart metadata does not accumulate after DeleteBucket.
+		bucketDir := filepath.Join(s.dir, bucket)
+		err := os.RemoveAll(bucketDir)
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("failed to remove bucket attributes: %w", err)
+		}
+		return nil
 	}
 
+	metadir := filepath.Join(s.dir, bucket, object, sidecarmeta)
 	err := os.RemoveAll(metadir)
 	if err != nil && !errors.Is(err, os.ErrNotExist) {
-		return fmt.Errorf("failed to remove attributes: %v", err)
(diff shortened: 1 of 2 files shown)


backend/meta/sidecar.go
```diff
@@ -145,12 +145,21 @@ func (s SideCar) ListAttributes(bucket, object string) ([]string, error) {
...
 func (s SideCar) DeleteAttributes(bucket, object string) error {
-	metadir := filepath.Join(s.dir, bucket, object, sidecarmeta)
 	if object == "" {
-		metadir = filepath.Join(s.dir, bucket, sidecarmeta)
+		// Remove the entire bucket sidecar directory so that orphaned
+		// object/multipart metadata does not accumulate after DeleteBucket.
+		bucketDir := filepath.Join(s.dir, bucket)
+		err := os.RemoveAll(bucketDir)
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("failed to remove bucket attributes: %v", err)
```
niksis02: `%w` for error wrapping ?

backend/posix/posix.go
```diff
@@ -641,6 +641,10 @@ func (p *Posix) DeleteBucket(ctx context.Context, bucket string) error {
 	if err != nil {
 		return fmt.Errorf("remove bucket: %w", err)
 	}
+	err = p.meta.DeleteAttributes(bucket, "")
+	if err != nil {
+		return fmt.Errorf("delete bucket attributes: %w", err)
+	}
```
niksis02: If DeleteAttributes fails, the function returns an error even though the bucket data is already deleted. Callers may treat this as a failed deletion. The sidecar metadata is now orphaned with no way to retry cleanup. Maybe the error should simply be ignored(or logged and ignored)?

backend/posix/posix.go
```diff
@@ -1042,6 +1050,18 @@ func (p *Posix) ensureNotDeleteMarker(bucket, object, versionId string) error {
 		return nil
 	}
 
+	// With path-based metadata backends (e.g. sidecar), RetrieveAttribute
+	// returns ErrNoSuchKey whether the sidecar attribute is absent OR the
+	// data file simply doesn't exist — the two cases are indistinguishable
+	// from metadata alone.  Verify the data file directly so callers
+	// receive the correct NoSuchVersion / NoSuchKey error.
+	if _, statErr := os.Stat(filepath.Join(bucket, object)); errors.Is(statErr, fs.ErrNotExist) {
```
niksis02: `syscall.ENOTDIR` error check is also necessary for cases like: `foo/bar` where foo is a regular file object

niksis02 (commented):
Looks good !

niksis02 (approved):

PR #2089, opened 2026-04-22
 @lrm25 test: GetObject - add query tests, go/bats improved header/object rep…

lrm25:


tests/drivers/complete_multipart_upload/complete_multipart_upload_rest.sh
```diff
@@ -264,3 +239,34 @@ upload_check_parts() {
...
+  if ! response=$(create_multipart_upload_rest "$1" "$2" "" "parse_upload_id" 2>&1); then
+    log 2 "error creating multipart upload: $response"
+    return 1
+  fi
+  upload_id="$response"
+
+  local parts_payload="" idx=1
+  for part in "${@:3}"; do
+    if ! response=$(upload_part_rest "$1" "$2" "$upload_id" "$idx" "$part" 2>&1); then
+      log 2 "error uploading part 1: $response"
```
anodelman: Just a nit - should be 'part $idx' instead of 'part 1'

tests/drivers/get_object/get_object_rest.sh
```diff
@@ -30,4 +30,71 @@ get_object_success_or_access_denied() {
...
+    log 2 "error setting up bucket and file: $response"
+    return 1
+  fi
+  read -r bucket_name file_name <<< "$response"
+
+  if ! result=$(put_object_rest "$TEST_FILE_FOLDER/$file_name" "$bucket_name" "$file_name" 2>&1); then
+    log 2 "error putting object: $result"
+    return 1
+  fi
+  if ! result=$(send_rest_go_command_callback "200" "check_for_header_key_and_value" "-bucketName" "$bucket_name" "-objectKey" "$file_name" "-query" "response-$1=$2" \
```
anodelman, line 49: Just another note here, want to be sure that you use urlencoding so that things work correctly for potential urls with spaces or special characters.

```
  local encoded_val
  encoded_val=$(python3 -c "import urllib.parse, sys; print(urllib.parse.quote(sys.argv[1]))" "$2")
```

tests/drivers/get_object/get_object_rest.sh
```diff
@@ -30,4 +30,71 @@ get_object_success_or_access_denied() {
...
+
+  local response
+  if ! response=$(get_file_size "$3" 2>&1); then
+    log 2 "error getting part size: $response"
+    return 1
+  fi
+  part_size="$response"
+
+  starting_byte=$((($2-1)*part_size))
+  ending_byte=$(($2*part_size-1))
```
anodelman: Your current calculation for ending_byte assumes every part is exactly part_size - the ending_byte must be calculated based on the actual size of the downloaded chunk ($3) relative to the starting position.

```
  local header_data="$1"
  local part_num="$2"
  local downloaded_file="$3"
  local original_part_file="$4"
  local total_file_size="$5"
  local response result

  if ! response=$(get_file_size "$downloaded_file" 2>&1); then
    log 2 "error getting part size: $response"
    return 1
  fi
  local actual_chunk_size="$response"

  local nominal_part_size
  nominal_part_size=$(get_file_size "$original_part_file")

  local starting_byte=$(( (part_num - 1) * nominal_part_size ))
  # The ending byte is the start + how much we actually got - 1
  local ending_byte=$(( starting_byte + actual_chunk_size - 1 ))
```

tests/drivers/get_object/get_object_rest.sh
```diff
@@ -30,4 +30,71 @@ get_object_success_or_access_denied() {
...
+  if ! check_param_count_v2 "bucket name, key, part number, part data file, original file size" 5 $#; then
+    return 1
+  fi
+  if ! response=$(get_file_names 2 2>&1); then
+    log 2 "error getting file names: $response"
+    return 1
+  fi
+  read -r header_file output_file <<< "$response"
+  log 5 "output file: $output_file"
+  if ! send_rest_go_command_callback "206" "validate_partial_content_response" "-bucketName" "$1" "-objectKey" "$2" "-query" "partNumber=$3" \
```
anodelman, line 90: There is a mismatch in the arguments you are passing to the callback versus what validate_partial_content_response expects.

Your Helper expects:

Header data file

Part number

Downloaded part file

Original part file

Original file size

Your Current Call passes:
"--" "$3" "$4" "$TEST_FILE_FOLDER/$output_file" "$5"

The Discrepancy: Your helper expects the Header File as the first argument after the --. Currently, you are passing the Part Number ($3) as the first argument.

anodelman (changes requested):

tests/drivers/get_object/get_object_rest.sh
lrm25: Query values are encoded on the go side, e.g., a command sent for `dummy+ one`:  `curl -ksi https://localhost:7070/versity-gwtest-bucket-one-1-7c588470-80ab-4b52-b367-a9126c670b2/test-file-FEC41EF8-EB4B-42EC-86A6-9EC59E298FAB?response-expires=dummy%2B%20one`

tests/drivers/get_object/get_object_rest.sh
lrm25: Good catch, changed the test for this scenario

tests/drivers/get_object/get_object_rest.sh
lrm25: the header data is passed as the first parameter by default to curl/go callback functions right now

anodelman (approved):

PR #2090, opened 2026-04-22
test: add windows build/test for functional tests

benmcclelland:
We have been getting a few bug reports on windows, so this is an attempt to have some test coverage for windows.

### DIFF
--- a/cmd/versitygw/test.go
+++ b/cmd/versitygw/test.go
@@ -44,6 +44,7 @@ var (
 	azureTests        bool
 	tlsStatus         bool
 	parallel          bool
+	windowsTests      bool
 	sidecarTests      bool
 )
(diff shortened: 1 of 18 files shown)


cmd/versitygw/test.go
niksis02, line 520: `windows-test-mode` is missing in `extractIntTests` to run a single windows test.
```suggestion
				&cli.BoolFlag{
					Name:        "windows-test-mode",
					Usage:       "Skips tests that are not supported on Windows",
					Destination: &windowsTests,
					Aliases:     []string{"windows"},
				},
```

niksis02 (changes requested):

niksis02 (approved):

PR #2094, opened 2026-04-25
feat: replace webui client-side name filter with server-side prefix filter

### LINKED ISSUE #2091: [Feature] - <suggest search s3 object not in page number 1>

suggest search s3 object not in page number 1
<1> we have a object="function" in second 1000-page
Image
<2> but we can't search object="function" in the first 1000-page
Image

**Additional context**

benmcclelland:
Remove the client-side search that filtered already-loaded objects by name. Replace it with a prefix input that is appended to the current path prefix and passed directly to the S3 ListObjectsV2 API, so filtering is performed server-side and works correctly across all pages.

Fixes #2091

### DIFF
--- a/webui/web/explorer.html
+++ b/webui/web/explorer.html
@@ -167,7 +167,7 @@ under the License.
             <svg class="w-4 h-4 text-charcoal-300 absolute left-3 top-1/2 transform -translate-y-1/2" fill="none" stroke="currentColor" viewBox="0 0 24 24">
               <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M21 21l-6-6m2-5a7 7 0 11-14 0 7 7 0 0114 0z"/>
             </svg>
-            <input type="text" id="search-input" class="pl-9 pr-8 py-1.5 w-64 border border-gray-200 rounded-lg text-sm focus:outline-none focus:border-accent" placeholder="Filter by name...">
+            <input type="text" id="search-input" class="pl-9 pr-8 py-1.5 w-64 border border-gray-200 rounded-lg text-sm focus:outline-none focus:border-accent" placeholder="Filter by prefix...">
             <button id="clear-search" class="hidden absolute right-2 top-1/2 transform -translate-y-1/2 text-charcoal-300 hover:text-charcoal" onclick="clearSearch()">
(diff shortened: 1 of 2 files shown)


webui/web/explorer.html
```diff
@@ -1700,61 +1699,27 @@ <h2 class="text-xl font-semibold text-charcoal">Multipart Uploads</h2>
...
-          return name.toLowerCase().includes(searchTerm);
-        });
-        filteredDeletedObjects = deletedObjectsWithVersions.filter(o => {
-          const name = getFileName(o.key);
-          return name.toLowerCase().includes(searchTerm);
-        });
-      }
+      const filteredFolders = folders;
+      const filteredObjects = objects;
+      const filteredDeletedObjects = deletedObjectsWithVersions;
```
niksis02: Dead assignments: The rendering code could use `folders`, `objects`, and `deletedObjectsWithVersions` directly.

niksis02 (changes requested):
Versioned, deleted objects bypass the prefix filter. ListObjectsV2 uses currentPrefix + searchPrefix, but version metadata still loads with only currentPrefix, and deletedObjectsWithVersions is rendered unfiltered. In a versioned bucket, filtering by foo can still show deleted keys outside foo, with restore/delete actions available.

webui/web/explorer.html
```diff
@@ -3221,21 +3165,20 @@ <h2 class="text-xl font-semibold text-charcoal">Multipart Uploads</h2>
...
-            reloadFirstPageForSearch();
-            return;
-          }
+          if (!currentBucket) return;
 
-          renderObjects();
-        }, 200));
+          currentPage = 1;
+          continuationTokens = [null];
+          loadObjects();
```
niksis02, line 3186: We have 300 milliseconds debounce, but still not guaranteed, that any previous pending request will not overwrite the state.
e.g
1. User types
2. 300 ml passes
3. A request is sent
4. User types again
5. after 300mls the second request is sent, but the first is still pengin
6. If the second response arrives sooner, then first, the first response will overwrite the state, based on first search input.

To resolve this, it would be nice to have `AbortController` for `loadObjects`. This is a common practice in FE

niksis02 (changes requested):

niksis02 (approved):

PR #2096, opened 2026-04-27
feat: add new ErrNoSpaceLeftOnDevice API error for ENOSPC errors

### LINKED ISSUE #2093: [Bug]: ENOSPC (No space left on device) in POSIX mode incorrectly mapped to generic InternalError instead of S3-compatible InsufficientStorage
**Describe the bug**
When the underlying POSIX filesystem runs out of space, VersityGW logs no space left on device but returns an S3 InternalError (HTTP 500) with the message "Unable to execute HTTP request: Broken pipe". Standard S3 behavior for storage exhaustion should ideally return InsufficientStorage (HTTP 507) or at least a more descriptive error string in the XML body to allow clients to handle the exception properly.

**To Reproduce**

- Set up VersityGW in POSIX mode pointing to a small partition or a directory with a strict disk quota.
- Fill the disk/quota to 100%.
- Attempt a PutObject or Multipart Upload via the Java SDK (or any S3 client).
- Observe the client receives a 500 InternalError and potentially a Broken Pipe during the data stream.

**Expected behavior**
The gateway should catch the ENOSPC syscall error and map it to an S3-compliant InsufficientStorage error code, or provide the OS error detail in the XML response.

**Server Version**
VersityGW Version: 1.4.1

benmcclelland:
Add a new non-AWS error ErrNoSpaceLeftOnDevice (HTTP 507 Insufficient Storage) to s3err. Update all call sites in the posix backend that could return ENOSPC and return the new error when the underlying filesystem has no space remaining.

Fixes #2093

### DIFF
--- a/backend/meta/sidecar.go
+++ b/backend/meta/sidecar.go
@@ -20,6 +20,9 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"syscall"
+
+	"github.com/versity/versitygw/s3err"
 )
 
 // SideCar is a metadata storer that uses sidecar files to store metadata.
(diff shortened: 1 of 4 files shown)


niksis02 (commented):
I wonder how we should handle `syscall.ENOSPC` for xattr and sidecar? Technically setting extended attributes(less possible) and createing sidecar file/directory might return the no space error. Shouldn't we explicitly handle these cases in xattr and sidecar packages ?

benmcclelland:
> I wonder how we should handle `syscall.ENOSPC` for xattr and sidecar? Technically setting extended attributes(less possible) and createing sidecar file/directory might return the no space error. Shouldn't we explicitly handle these cases in xattr and sidecar packages ?

Yes, that's correct, I need to check those as well. Forgot about those.

niksis02 (approved):

PR #2099, opened 2026-04-28
s3api: add custom route and middleware options

miretskiy:
Add options for embedders to register Fiber routes and middleware before the S3 route table is initialized.

`WithRoute` registers a top-level route with explicit method and path matching. `WithMiddleware` registers prefix middleware that can handle a request or call ctx.Next() to continue into the S3 stack.

Add coverage for route registration order when a top-level route and catch-all middleware are both configured.

### DIFF
--- a/s3api/server.go
+++ b/s3api/server.go
@@ -55,7 +55,20 @@ type S3ApiServer struct {
 	maxRequests      int
 	webuiMountPrefix string
 	webuiSrvCfg      *webui.ServerConfig
+	routes           []routeMount
+	middlewares      []middlewareMount
 	socketPerm       os.FileMode
+}
+
+type routeMount struct {
+	method   string
+	path     string
+	handlers []fiber.Handler
+}
+
+type middlewareMount struct {
+	prefix  string
+	handler fiber.Handler
 }
 
 func New(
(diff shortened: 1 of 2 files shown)


versity-github:

s3api/server.go
```diff
@@ -130,6 +143,34 @@ func New(
...
+		if len(route.handlers) == 0 {
+			return nil, fmt.Errorf("invalid route for %s %s: no handlers", route.method, route.path)
+		}
+		for i, handler := range route.handlers {
+			if handler == nil {
+				return nil, fmt.Errorf("invalid route for %s %s: nil handler at index %d", route.method, route.path, i)
+			}
+		}
+		app.Add(route.method, route.path, route.handlers...)
+	}
```
niksis02, line 155: Requests hitting custom routes are never rate-limited. As rate limit is configurable with the gateway flags, it makes sense to patch the custom routes after the rate limiter middleware.

s3api/server.go
```diff
@@ -130,6 +143,34 @@ func New(
 		webui.MountOn(app, server.webuiMountPrefix, server.webuiSrvCfg)
 	}
 
+	for _, route := range server.routes {
+		if route.method == "" {
+			return nil, fmt.Errorf("invalid route for path %q: empty method", route.path)
+		}
```
niksis02: It makes sense to add full method validation: not only empty check.

s3api/server_test.go
niksis02, line 1: The test only covers the happy path (priority ordering). None of the error branches are exercised by tests: nil handler, empty method, invalid path, etc...

It would be nice to have the error cases tested.

niksis02 (changes requested):

miretskiy (commented):
Thank you for your review; Comments addressed.

niksis02 (approved):

PR #2100, opened 2026-04-29
fix: POSIX support for prefix parts containing spaces

### LINKED ISSUE #2098: [Bug] - Spaces in prefixes are incorrectly decoded with the POSIX backend
Note: I can work on a PR to try and manage it properly, once I figure out the extend to whcih ths applies. I have only seen it in two instances, but I would not be surprised if this happened on all POSIX operations.

**Describe the bug**

When browsing a prefix containing a space, no objects are shown. The corollary is that creating an object under that prefix will create a new prefix with the space encoded as `%20`.

The prefix, when parsed from `ctx.Query("prefix")`, is not URL decoded, and therefore includes a literal `%20`, which remains when performing filesystem operations.

**To Reproduce**

 * Create a directory on the file system called `Hello World` and a file within it.
 * Browse that prefix in the Web UI.
 * See that the file does not appear.

Also:

 * When within the prefix, in the Web UI
 * Upload a file
 * Look at the filesystem (or the Web UI), there are now two directories, `Hello World` and `Hello%20World`.

You can also notice that the name of the prefix part is not properly encoded in the breadcrumbs.

**Expected behavior**

Those space-containing prefixes to be properly managed.

**Server Version**

Both:

Reproduced in both v1.4.0 and v1.4.1.
On both macOS and Linux.

apognu:
Closes #2098.

This PR fixes invalid behavior when a prefix part contains spaces with the POSIX backend.

There was a mismatch between the encoding of the prefix in the query param (spaces encoded as `%20`) and lookups on the filesystem. It also fixes prefix display in the explorer breadcrumbs.

---

This PR is not done, there is still a weird behavior with those prefixes: "root" prefixes with spaces (but only those right under a bucket root) always contains a dummy object named after the prefix.

image

I am not very familiar with the codebase yet and might need some help pinpointing where this is coming from.

### DIFF
--- a/s3api/controllers/bucket-get.go
+++ b/s3api/controllers/bucket-get.go
@@ -15,6 +15,7 @@
 package controllers
 
 import (
+	"net/url"
 	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/service/s3"
@@ -299,7 +300,7 @@ func (c S3ApiController) GetBucketPolicyStatus(ctx *fiber.Ctx) (*Response, error
 func (c S3ApiController) ListObjectVersions(ctx *fiber.Ctx) (*Response, error) {
 	// url values
 	bucket := ctx.Params("bucket")
-	prefix := ctx.Query("prefix")
+	prefix, _ := url.QueryUnescape(ctx.Query("prefix"))
 	delimiter := ctx.Query("delimiter")
 	maxkeysStr := ctx.Query("max-keys")
 	keyMarker := ctx.Query("key-marker")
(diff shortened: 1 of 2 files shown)


versity-github:

benmcclelland:
I'll take a look at this and see if I can find why its not rendering the top level directories correctly there.

apognu:
As I commented in #2098, the dummy/marker file being rendered in the UI was a filter I had to adjust.

But it doesn't explain why it's also returned through the S3 API.

Remaining available if I need to add or change something.

benmcclelland:
Sorry it took so long to get back to this. I wasnt able to recreate any of the cli issues, but I did see that the webui was not correctly decoding the paths. I opened a new PR #2139 to fix that issue. See if that one works for you.

PR #1739, opened 2026-01-08
feat: implements admin CreateBucket endpoint/cli command

niksis02:
Closes #1731

Implements the admin `CreateBucket` (`PATCH /:bucket/create`) endpoint and CLI command, which create a new bucket with the provided owner access key ID. The endpoint internally calls the S3 `CreateBucket` API, storing the new owner information in the request context under the `bucket-owner` key. This value is then retrieved by the S3 API layer and the backends.

The endpoint uses the custom `x-vgw-owner` HTTP header to pass the bucket owner access key ID.

The admin CLI command mirrors `aws s3api create-bucket` and supports all flags implemented by the gateway (for example, `--create-bucket-configuration`, `--acl`, `--object-ownership`, etc.).

PR #2106, opened 2026-05-01
fix: apply CORS middleware to admin CreateBucket route

### LINKED ISSUE #2105: [Bug] - admin PATCH /:bucket/create response missing CORS headers
**Describe the bug**
The admin API route `PATCH /:bucket/create` returns success/error responses without any CORS headers (`Access-Control-Allow-Origin`, `Vary`, `Access-Control-Expose-Headers`), even when `--cors-allow-origin` is configured. All other admin PATCH routes (`/create-user`, `/delete-user`, `/update-user`, `/list-users`, `/change-bucket-owner`, `/list-buckets`) return CORS headers correctly.

The OPTIONS preflight handler for the same path is configured with `ApplyDefaultCORSPreflight` + `ApplyDefaultCORS`, so preflight succeeds — but the actual PATCH response from the browser is then blocked by the same-origin policy because `Access-Control-Allow-Origin` is absent.

Result: WebUI bucket creation against an admin endpoint on a different origin fails with `ERR_FAILED` despite the server returning `201 Created`.

**To Reproduce**
1. Run versitygw with `--cors-allow-origin "https://example.com"` and `--admin-port :7071`.
2. Send `PATCH /list-buckets` and `PATCH /:bucket/create` with header `Origin: https://example.com` (using sigv4):
   ```
   awscurl --service s3 -X PATCH -H "Origin: https://example.com" \
     -i http://127.0.0.1:7071/list-buckets
   awscurl --service s3 -X PATCH -H "Origin: https://example.com" \
     -H "X-Vgw-Owner: admin" \
     -i http://127.0.0.1:7071/mybucket/create
   ```
3. The first response includes `Access-Control-Allow-Origin: https://example.com`. The second response has no CORS headers at all.

**Expected behavior**
`PATCH /:bucket/create` should return the same CORS headers as the other admin PATCH routes when `--cors-allow-origin` is set.

**Server Version**
main @ 437d4685 (also affects every release since #1739).

**Additional context**
Root cause: `s3api/admin-router.go:109-113` is the only admin PATCH route whose handler chain omits `middlewares.ApplyDefaultCORS(corsAllowOrigin)`. The OPTIONS handler at lines 114-117 has it, so preflight works; only the actual response is broken. Introduced in #1739 when the endpoint was added.

A PR with the one-line fix is on the way.

---
*Investigation assisted by Claude (Anthropic).*

dyld-w:
## Summary

Adds the missing `middlewares.ApplyDefaultCORS(corsAllowOrigin)` to the `PATCH /:bucket/create` admin route handler chain. Every other admin PATCH route in `s3api/admin-router.go` already applies it; this one was overlooked when the endpoint was added in #1739.

The OPTIONS preflight handler for the same path already sets CORS headers, so preflight succeeds — but the actual `PATCH` response was returned without `Access-Control-Allow-Origin`, causing browsers to block it with `ERR_FAILED` even though the gateway returned `201 Created`. This breaks bucket creation from the WebUI when it is served from a different origin than the admin API.

Fixes #2105.

## Test plan

Manual reproduction with `--cors-allow-origin "https://example.com"` on `:7071`:

Before:
```
$ awscurl --service s3 -X PATCH -H "Origin: https://example.com" \
    -H "X-Vgw-Owner: admin" -i http://127.0.0.1:7071/mybucket/create
# response has no Access-Control-Allow-Origin / Vary / Access-Control-Expose-Headers
```

After:
```
$ awscurl --service s3 -X PATCH -H "Origin: https://example.com" \
    -H "X-Vgw-Owner: admin" -i http://127.0.0.1:7071/mybucket/create
# Access-Control-Allow-Origin: https://example.com
# Vary: Origin, Access-Control-Request-Headers, Access-Control-Request-Method
# Access-Control-Expose-Headers: ETag
```

Sibling routes (e.g. `/list-buckets`) behave identically before and after, confirming no regression.

- [x] Verified buggy build returns no CORS headers on `/:bucket/create`
- [x] Verified fixed build returns the same CORS headers as sibling admin PATCH routes
- [x] `go build ./s3api/...` clean

---
*Investigation and fix assisted by Claude (Anthropic).*

### DIFF
--- a/s3api/admin-router.go
+++ b/s3api/admin-router.go
@@ -110,6 +110,7 @@ func (ar *S3AdminRouter) Init(app *fiber.App, be backend.Backend, iam auth.IAMSe
 		controllers.ProcessHandlers(ctrl.CreateBucket, metrics.ActionAdminListBuckets, services,
 			middlewares.VerifyV4Signature(root, iam, region, false, true, false),
 			middlewares.IsAdmin(metrics.ActionAdminCreateBucket),
+			middlewares.ApplyDefaultCORS(corsAllowOrigin),
 		))
 	app.Options("/:bucket/create",
 		middlewares.ApplyDefaultCORSPreflight(corsAllowOrigin),


versity-github:

benmcclelland (approved):

PR #2114, opened 2026-05-06
fix: reject SigV2 requests

niksis02:
The gateway currently supports only Signature Version 4 (SigV4) authorization. Deprecated AWS SigV2 requests are now rejected with an AWS-specific `InvalidRequest` error for both Authorization-header requests and query-string requests(presigned URLs).

This also fixes SigV4 Authorization-header handling for date headers. SigV4 accepts two date headers: `Date` and `X-Amz-Date`. `X-Amz-Date` takes precedence, but when it is missing, `Date` should be used. The gateway now uses the `Date` header with lower precedence when `X-Amz-Date` is not present. No SDK integration test was added for this case because the SDK always sets `X-Amz-Date`, and this behavior is not configurable.

### DIFF
--- a/s3api/utils/presign-auth-reader.go
+++ b/s3api/utils/presign-auth-reader.go
@@ -254,11 +254,21 @@ func IsPresignedURLAuth(ctx *fiber.Ctx) bool {
 	signedHeaders := ctx.Query("X-Amz-SignedHeaders")
 	expires := ctx.Query("X-Amz-Expires")
 
-	return !isEmpty(algo, creds, signature, signedHeaders, expires)
+	return !allEmpty(algo, creds, signature, signedHeaders, expires) || IsPresignedURLAuthV2(ctx)
 }
 
-// isEmpty checks if all the given strings are empty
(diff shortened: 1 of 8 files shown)


s3api/utils/presign-auth-reader.go
```diff
@@ -254,7 +254,17 @@ func IsPresignedURLAuth(ctx *fiber.Ctx) bool {
...
+}
+
+// IsPresignedURLAuthV2 determines if the request is
+// query-string signed with aws v2 signer
+func IsPresignedURLAuthV2(ctx *fiber.Ctx) bool {
+	expires := ctx.Query("Expires")
+	access := ctx.Query("AWSAccessKeyId")
+	signature := ctx.Query("Signature")
+
+	return !isEmpty(expires, access, signature)
```
benmcclelland: isEmpty returns true only when all args are empty, so !isEmpty(...) is true when any one is non-empty. would a better check be:
```
return expires != "" && access != "" && signature != ""
```

benmcclelland (approved):

PR #2122, opened 2026-05-08
Bug: %2F used as bucket/key separator

alexgalie:
Fix for issue #2024 - some clients like AWS .NET SDK send bucket/key separator as "%2F", which should be valid as the documentation states "The value must be URL-encoded." (https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html)


versity-github:

auth/access-control.go
```diff
@@ -37,8 +38,14 @@ func VerifyObjectCopyAccess(ctx context.Context, be backend.Backend, copySource
 	if err := VerifyAccess(ctx, be, opts); err != nil {
 		return err
 	}
-	// Verify source bucket access
-	srcBucket, srcObject, found := strings.Cut(copySource, "/")
+	// Verify source bucket access.
+	// URL-decode the copy source before splitting so that clients which send
+	// the bucket/key separator as "%2F" correctly.
+	decodedSrc, err := url.QueryUnescape(strings.TrimPrefix(copySource, "/"))
```
benmcclelland: This fix looks correct, but I'm noticing that the other part of this fix stops at the versionid. @niksis02 Do we need to remove version here as well?
```go
src := strings.TrimPrefix(copySource, "/")
i := strings.LastIndex(src, "?versionId=")
if i != -1 {
    src = src[:i]
}
decodedSrc, err := url.QueryUnescape(src)
```

and actually, I think the call sites in object-put already call TrimPrefix("/"), so that part is probably not needed here.

alexgalie: Removed the prefix trimming. 

For a bit more context, the other one is a very narrow edge case where a key would contain the string "?versionId=" itself.

niksis02 (approved):
@benmcclelland Now it looks good to me !

auth/access-control.go
benmcclelland: fair enough, we can address the version issue in a new PR since that is an existing bug anyways.

benmcclelland:
> @benmcclelland Now it looks good to me !

I think we still have an issue that we can fix in a separate PR: If the caller passes "src-bucket/src-key?versionId=abc123", then srcObject becomes "src-key?versionId=abc123", which is passed to the ACL checker as the object key. So I think we need to remove the "?versionId..." suffix if it exists here.

benmcclelland (approved):

PR #2137, opened 2026-05-12
test: HeadObject query tests

lrm25:


tests/drivers/put_object/put_object_rest.sh
```diff
@@ -71,6 +71,35 @@ setup_bucket_and_add_files() {
   return 0
 }
 
+setup_bucket_and_add_file_v2() {
```
anodelman: since you are calling setup_bucket_v3, you might want to rename this function to setup_bucket_and_file_v3 to maintain consistent versioning across your test suite helpers.

tests/drivers/string.sh
```diff
@@ -0,0 +1,52 @@
...
+      } 2>"$TEST_FILE_FOLDER/$pipe_error_file"
+    )"
+    status=$?
+
+    if [ $status -ne 0 ]; then
+      log 2 "error creating chunk: $(cat "$TEST_FILE_FOLDER/$pipe_error_file")"
+      return 1
+    fi
+
+    string="${string}${chunk}"
```
anodelman, line 48: you append to string="${string}${chunk}". However, string is not cleared at the start. It will just keep growing, potentially returning a string much longer than your requested $size if this is called multiple times.

anodelman (changes requested):

anodelman (approved):

PR #2140, opened 2026-05-14
test: ListObjects util code cleanup, multipart updates

lrm25:


tests/drivers/list_objects/list_objects_s3api.sh
```diff
@@ -0,0 +1,124 @@
...
+    return 1
+  fi
+}
+
+list_check_objects_v2() {
+  if ! check_param_count "list_check_objects_v2" "bucket, expected key one, expected size one, expected key two, expected size two" 5 $#; then
+    return 1
+  fi
+  if ! response=$(list_objects_v2 "$1" 2>&1); then
+    log 2 "error listing objects (s3api, v1): $response"
```
anodelman: Should be v2 here

tests/drivers/list_objects/list_objects_s3api.sh
```diff
@@ -0,0 +1,124 @@
...
+    return 1
+  fi
+  if ! size_two=$(echo "$1" | jq '.Contents[1].Size' 2>&1); then
+    log 2 "error obtaining size two: $size_two"
+    return 1
+  fi
+  if [[ $size_two -ne "$5" ]]; then
+    log 2 "Object two size mismatch ($size_two, $5)"
+    return 1
+  fi
```
anodelman, line 82: Need default return 0 here

tests/drivers/list_objects/list_objects_s3api.sh
```diff
@@ -0,0 +1,124 @@
...
+  return 0
+}
+
+check_listed_objects() {
+  if ! check_param_count_gt "JSON data, expected key one, expected size one, expected key two, expected size two" 5 $#; then
+    return 1
+  fi
+  # shellcheck disable=SC2154
+  if ! key_one=$(echo "$1" | jq -r '.Contents[0].Key' 2>&1); then
```
anodelman: Amazon S3 (and compliant gateways) returns object listings sorted lexicographically (alphabetically) by Key. It does not return them in the order they were uploaded - so using an assumed ordering here could backfire and have the two keys reversed.

anodelman (changes requested):

tests/drivers/list_objects/list_objects_rest.sh
```diff
@@ -420,3 +421,214 @@ list_objects_delimiter() {
...
+  fi
+  if [ "$result" != "400" ]; then
+    log 2 "expected result code of '400' was '$result'"
+    return 1
+  fi
+  if ! check_xml_element "$TEST_FILE_FOLDER/$file_name" "InvalidArgument" "Error" "Code"; then
+    log 2 "invalid error code"
+    return 1
+  fi
+}
```
anodelman, line 603: nit - missing return 0

anodelman (changes requested):

anodelman (approved):

PR #2147, opened 2026-05-20
fix: reject invalid PostObject keys

niksis02:
Validate multipart PostObject key fields with the existing object name rules so path traversal and degenerate names return BadRequest. This prevents crafted object keys from escaping the gateway root.

### DIFF
--- a/s3api/middlewares/object-post-auth.go
+++ b/s3api/middlewares/object-post-auth.go
@@ -86,6 +86,16 @@ func AuthorizePostObject(root RootUserConfig, iam auth.IAMService, region string
 
 		fields := result.Fields
 
+		if fields["key"] == "" {
+			debuglogger.Logf("missing object key")
+			return s3err.PostAuth.MissingField("key")
+		}
+
+		if !utils.IsObjectNameValid(fields["key"]) {
+			debuglogger.Logf("invalid POST object key: %q", fields["key"])
+			return s3err.GetAPIError(s3err.ErrBadRequest)
+		}
+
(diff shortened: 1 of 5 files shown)


s3api/middlewares/object-post-auth.go
```diff
@@ -86,6 +86,11 @@ func AuthorizePostObject(root RootUserConfig, iam auth.IAMService, region string
 
 		fields := result.Fields
 
+		if !utils.IsObjectNameValid(fields["key"]) {
```
benmcclelland, line 94: need to check for presence in the map, since absent keys map to "" here. previously they would have fallen through and returned the more specific missing field error. maybe somethign like this instead:
```go
if key := fields["key"]; key != "" && !utils.IsObjectNameValid(key) {
    debuglogger.Logf("invalid POST object key: %q", key)
    return s3err.GetAPIError(s3err.ErrBadRequest)
}
```
if you are ok with leaving the missing key error handling to the controller.

tests/integration/PostObject.go
```diff
@@ -309,6 +309,45 @@ func PostObject_access_denied(s *S3Conf) error {
...
+	return actionHandler(s, testName, func(s3client *s3.Client, bucket string) error {
+		for _, obj := range []string{
+			".",
+			"..",
+			"./",
+			"/.",
+			"//",
+			"../",
+			"/..",
+			"/..",
```
benmcclelland: duplicate test case for "/.."

benmcclelland (approved):

PR #2153, opened 2026-05-22
fix: normalize object keys during bucket policy evaluation

niksis02:
Object key validation allowed internal parent-directory segments such as `public/../private.txt`. Bucket policy and auth checks evaluated the raw key, so a policy allowing bucket/public/* could match the request while posix backend later resolved the key with `filepath.Join` and accessed `bucket/private.txt`.

Add backend-specific object key normalization to close that mismatch. The Backend interface now exposes `NormalizeObjectKey` so authorization can evaluate resources using the same key shape a backend will use for storage access.

Backends that do not collapse object paths, including Azure and the S3 proxy, inherit `BackendUnsupported.NormalizeObjectKey`. That implementation returns the input key unchanged, avoiding unnecessary normalization and keeping policy evaluation unpolluted for object stores where ../ is part of the key name.

posix/scoutfs normalize keys with filepath.Join so policy resources and request keys are compared after internal dot segments are collapsed.

Bucket policy evaluation now normalizes both the incoming object key and object resource patterns from the policy before matching. Object lock governance bypass policy checks use the same backend normalizer as well, so retention and legal hold authorization cannot diverge from backend path resolution.

### DIFF
--- a/auth/bucket_policy.go
+++ b/auth/bucket_policy.go
@@ -60,6 +60,8 @@ type BucketPolicy struct {
 	Statement []BucketPolicyItem `json:"Statement"`
 }
 
+type objectKeyNormalizer func(bucket, object string) string
+
 func (bp *BucketPolicy) UnmarshalJSON(data []byte) error {
 	var tmp struct {
 		Version   *PolicyVersion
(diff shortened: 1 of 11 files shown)


backend/posix/posix.go
```diff
@@ -6598,6 +6598,10 @@ func (p *Posix) ListBucketsAndOwners(ctx context.Context) (buckets []s3response.
 	return buckets, nil
 }
 
+func (p *Posix) NormalizeObjectKey(key string) string {
+	return filepath.Join(".", key)
+}
```
benmcclelland: I think we need to pass the bucket in here to get the real normalized path. But maybe the return string should still prune the bucket name from the path? The Join here is rooted in the top level dir, not the bucket directory. So I think this might cause some possible issues resolving the actual paths?

auth/bucket_policy.go
```diff
@@ -60,6 +60,8 @@ type BucketPolicy struct {
 	Statement []BucketPolicyItem `json:"Statement"`
 }
 
+type ObjectKeyNormalizer func(string) string
```
benmcclelland: does ObjectKeyNormalizer need to be exported? looks like this is just an internal detail?

benmcclelland (approved):

PR #2157, opened 2026-05-26
test: ListMultipartUpload query tests

lrm25:


tests/drivers/list_multipart_uploads/list_multipart_uploads_rest.sh
```diff
@@ -0,0 +1,177 @@
...
+  return 0
+}
+
+list_multipart_uploads_check_next_values() {
+  if ! check_param_count_v2 "bucket, expected next key, expected upload ID" 3 $#; then
+    return 1
+  fi
+  if ! result=$(send_rest_go_command_callback "200" "check_multipart_upload_next_key_upload_id" "-bucketName" "$1" "-query" "uploads" \
+      "--" "$2" "$3" 2>&1); then
+    log 2 "error listing multipart upload and checking owner data: $result"
```
anodelman: This matches a previous log message and should be custom for this function (probably just a cut/paste issue).  Maybe something "error listing multipart upload and checking next values..."

tests/drivers/list_multipart_uploads/list_multipart_uploads_s3api.sh
```diff
@@ -0,0 +1,48 @@
...
+  if [ "$key" != "$2" ]; then
+    log 2 "expected key of '$2', was '$key'"
+    return 1
+  fi
+  if [ "$upload_id" != "$3" ]; then
+    log 2 "expected upload ID of '$3', was '$upload_id'"
+    return 1
+  fi
+
+  next_token="$(echo "$upload_json" | jq -r '.NextToken')"
```
anodelman, line 45: If the current page happens to be the last page in the dataset, the S3 gateway will omit the token parameter from its payload. When a key is missing, jq -r returns the literal string null with a successful exit code 0.

```
if [[ -z "$next_token" ]]; then
    log 5 "Terminal boundary reached. No further pagination markers present."
    echo "null"
  else
    echo "$next_token"
  fi
```

tests/drivers/list.sh
```diff
@@ -0,0 +1,88 @@
...
+  if ! check_xml_element "$1" "$3" "$6" "Delimiter"; then
+    log 2 "error checking delimiter"
+    return 1
+  fi
+  if ! check_element_count "$1" "$4" "$6" "CommonPrefixes" "Prefix"; then
+    log 2 "Prefix count mismatch"
+    return 1
+  fi
+  if ! check_element_count "$1" "$5" "$6" "$7" "Key"; then
+    log 2 "Prefix count mismatch"
```
anodelman: Doesn't look like the right log message here.

tests/test_rest_list_multipart_uploads.sh
```diff
@@ -0,0 +1,232 @@
...
+  assert_success
+  read -r bucket_name test_file <<< "$output"
+
+  run create_multipart_upload_rest "$bucket_name" "$test_file" "" "parse_upload_id"
+  assert_success
+  upload_id_one=$output
+
+  run send_rest_go_command_expect_error "400" "InvalidArgument" "Conflicting query string parameters" "-query" "uploads&uploadId=$upload_id" "-bucketName" "$bucket_name"
+  assert_success
+  upload_id=$output
```
anodelman: This variable is not used in the test?  Probably just a leftover.

tests/test_rest_list_multipart_uploads.sh
```diff
@@ -0,0 +1,232 @@
...
+  fi
+  run setup_bucket_and_file_v3 "$BUCKET_ONE_NAME"
+  assert_success
+  read -r bucket_name test_file <<< "$output"
+
+  run create_multipart_upload_rest "$bucket_name" "$test_file" "" "parse_upload_id"
+  assert_success
+  upload_id_one=$output
+
+  run send_rest_go_command_expect_error "400" "InvalidArgument" "Conflicting query string parameters" "-query" "uploads&uploadId=$upload_id" "-bucketName" "$bucket_name"
```
anodelman: Should this be using upload_id_one as defined above?

anodelman (changes requested):
In general, in your helper functions you should look towards doing local variable definitions.  Each call to a helper function is leaking variables into the main scope and could cause some interference if there's a name collision.

tests/drivers/list_multipart_uploads/list_multipart_uploads_s3api.sh
lrm25: The test (last one in `test_s3api_multipart.sh`) checks for `null` on the last line.

tests/test_rest_list_multipart_uploads.sh
lrm25: Test actually still works properly with any `uploadId` here, even an empty one, but yes it should.

anodelman (approved):

PR #2160, opened 2026-05-30
fix: connection early termination resulting in internal error

benmcclelland:
When the connection terminates before all bytes read, we were getting an io.ErrUnexpectedEOF that was not being handled as a standard io.EOF resulting in an internal error being raised. Translate io.ErrUnexpectedEOF to io.EOF so that we return the normal errors for unexpected content. Add a log message so that its clear the error is due to the connection being terminated before all data sent and not the fault of the gateway.

### DIFF
--- a/s3api/utils/csum-reader.go
+++ b/s3api/utils/csum-reader.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/cespare/xxhash/v2"
+	"github.com/versity/versitygw/debuglogger"
 	"github.com/versity/versitygw/s3err"
 	"github.com/zeebo/xxh3"
 )
@@ -127,6 +128,13 @@ func NewHashReader(r io.Reader, expectedSum string, ht HashType) (*HashReader, e
 // Read allows *HashReader to be used as an io.Reader
 func (hr *HashReader) Read(p []byte) (int, error) {
 	n, readerr := hr.r.Read(p)
+	// Treat ErrUnexpectedEOF as EOF so a truncated body triggers checksum
+	// validation (which will fail on partial data) rather than leaking a
+	// raw Go error as an internal server error.
+	if readerr == io.ErrUnexpectedEOF {
+		debuglogger.Logf("client connection terminated early")
+		readerr = io.EOF
+	}
 	_, err := hr.hash.Write(p[:n])
 	if err != nil {
 		return n, err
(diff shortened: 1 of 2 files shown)


s3api/utils/csum-reader.go
```diff
@@ -127,6 +128,13 @@ func NewHashReader(r io.Reader, expectedSum string, ht HashType) (*HashReader, e
 // Read allows *HashReader to be used as an io.Reader
 func (hr *HashReader) Read(p []byte) (int, error) {
 	n, readerr := hr.r.Read(p)
+	// Treat ErrUnexpectedEOF as EOF so a truncated body triggers checksum
+	// validation (which will fail on partial data) rather than leaking a
+	// raw Go error as an internal server error.
+	if readerr == io.ErrUnexpectedEOF {
+		debuglogger.Infof("client connection terminated early")
```
niksis02: nit: I think `Infof` is meant to provide additional information in complex success paths ? This is more like an unexpected behavior from client side. Maybe it would be better to use `logf` and print out with `[DEBUG]` prefix ?

s3api/utils/signed-chunk-reader.go
```diff
@@ -124,6 +124,14 @@ func NewSignedChunkReader(r io.Reader, authdata AuthData, canonicalString, secre
 // Read satisfies the io.Reader for this type
 func (cr *ChunkReader) Read(p []byte) (int, error) {
 	n, err := cr.r.Read(p)
+	// Treat ErrUnexpectedEOF as EOF so a connection that closes before
+	// all Content-Length bytes arrive follows the normal EOF path and
+	// returns a proper S3 error (e.g. ErrContentLengthMismatch or
+	// SignatureDoesNotMatch) instead of leaking an internal Go error.
+	if err == io.ErrUnexpectedEOF {
+		debuglogger.Infof("client connection terminated early")
```
niksis02: The same here

s3api/utils/csum-reader.go
benmcclelland: I'm a bit conflicted here. I dont want to print debug logs unless they are asked for. But we just hit this case with a real client and the only way we were able to identify that it was a client issue and not a server issue was with a log message indicating it was a connection terminated. But in general I dont like outputting logging if its not requested. Not sure what is best here.

niksis02: Anyways infof is printed only when debug log is enabled. So in terms of `printing-only-when-asked`, logf and infof are the same.

benmcclelland: actually, yeah, lets just do debug logging for now. and I can think about if we need something else here later.

niksis02 (approved):

PR #2161, opened 2026-05-30
feat: add group-based role mapping for FreeIPA IAM

benmcclelland:
FreeIPA users were always assigned RoleUser regardless of their group membership. This adds --ipa-admin-group-cn so that members of a specified FreeIPA group are automatically granted the admin role, and  --ipa-enable-userplus to elevate the default role to userplus for all IPA users. The user_show RPC is updated to request all attributes so that memberof_group is reliably returned.

### DIFF
--- a/auth/iam_ipa.go
+++ b/auth/iam_ipa.go
@@ -49,21 +49,25 @@ type IpaIAMService struct {
 	vaultName       string
 	username        string
 	password        string
+	adminGroupCN    string
+	enableUserPlus  bool
 	kraTransportKey *rsa.PublicKey
 	rootAcc         Account
 }
 
 var _ IAMService = &IpaIAMService{}
 
-func NewIpaIAMService(rootAcc Account, host, vaultName, username, password string, isInsecure bool) (*IpaIAMService, error) {
+func NewIpaIAMService(rootAcc Account, host, vaultName, username, password, adminGroupCN string, enableUserPlus, isInsecure bool) (*IpaIAMService, error) {
 	ipa := IpaIAMService{
-		id:        0,
-		version:   IpaVersion,
-		host:      host,
-		vaultName: vaultName,
-		username:  username,
-		password:  password,
-		rootAcc:   rootAcc,
+		id:             0,
+		version:        IpaVersion,
+		host:           host,
+		vaultName:      vaultName,
+		username:       username,
+		password:       password,
+		adminGroupCN:   adminGroupCN,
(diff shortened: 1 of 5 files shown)


auth/iam_ipa.go
```diff
@@ -124,15 +128,16 @@ func (ipa *IpaIAMService) GetUserAccount(access string) (Account, error) {
...
 	}
 
 	userResult := struct {
-		Gidnumber []string
-		Uidnumber []string
-		PidNumber []string
+		Gidnumber     []string `json:"gidnumber"`
+		Uidnumber     []string `json:"uidnumber"`
+		PidNumber     []string `json:"pidnumber"`
+		MemberOfGroup []string `json:"memberof_group"`
```
niksis02, line 140: Should nested FreeIPA group membership count for admin users? Right now we only read/check `memberof_group`, so a user who reaches `ipa-admin-group-cn` through a nested group would stay `user` or `userplus`. If nested groups should count, we should also read/check `memberofindirect_group`. Otherwise, the `direct-membership-only` behavior should be documented for `--ipa-admin-group-cn`.

niksis02 (commented):
I'm thinking we should make FreeIPA role management a bit more flexible. With the current implementation, users can effectively be mapped as either `user`/`admin` or `userplus`/`admin`, depending on whether `ipa-enable-userplus` is enabled. There doesn't seem to be a way to configure the gateway so all three roles, `user`, `userplus`, and `admin`, are managed through FreeIPA group membership.

What if we replace `ipa-enable-userplus` with something like `ipa-userplus-group-cn`, so `userplus` can be assigned based on FreeIPA group membership as well?

```
admin group -> admin
userplus group -> userplus
otherwise -> user
```

benmcclelland:
> I'm thinking we should make FreeIPA role management a bit more flexible. With the current implementation, users can effectively be mapped as either `user`/`admin` or `userplus`/`admin`, depending on whether `ipa-enable-userplus` is enabled. There doesn't seem to be a way to configure the gateway so all three roles, `user`, `userplus`, and `admin`, are managed through FreeIPA group membership.
> 
> What if we replace `ipa-enable-userplus` with something like `ipa-userplus-group-cn`, so `userplus` can be assigned based on FreeIPA group membership as well?
> 
> ```
> admin group -> admin
> userplus group -> userplus
> otherwise -> user
> ```

This is a good idea. Let me work with the current users to see if they have any feedback on these ideas. We can push this PR until a bit later to wait for feedback.

PR #2167, opened 2026-06-05
s3api: add WithOnListen option for server readiness notification

miretskiy:
WithOnListen registers a callback invoked once the server is bound and ready to accept connections. It wraps fiber's existing OnListen hook, which fires immediately before the first connection is served.

This allows callers to detect readiness deterministically rather than relying on a fixed sleep or polling a health endpoint.

Example use in tests:

    ready := make(chan struct{})
    srv, _ := s3api.New(..., s3api.WithOnListen(func() { close(ready) }))
    go srv.ServeMultiPort([]string{addr})
    <-ready // blocks until the server is accepting connections

### DIFF
--- a/s3api/server.go
+++ b/s3api/server.go
@@ -58,6 +58,7 @@ type S3ApiServer struct {
 	routes           []routeMount
 	middlewares      []middlewareMount
 	socketPerm       os.FileMode
+	onListen         func()
 }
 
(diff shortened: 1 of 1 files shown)


versity-github:

benmcclelland (approved):

PR #2168, opened 2026-06-05
test: jailing work (base)

lrm25:


tests/env.sh
```diff
@@ -330,23 +314,101 @@ check_user_vars() {
...
-delete_temp_log_if_exists() {
-  if [ -e "$TEST_LOG_FILE.tmp" ]; then
-    if ! error=$(rm "$TEST_LOG_FILE.tmp" 2>&1); then
+init_command_log() {
+  if ! delete_command_log; then
+    exit 1
+  fi
+  if ! echo "******** $(date +"%Y-%m-%d %H:%M:%S") $BATS_TEST_NAME COMMANDS ********" >> "$COMMAND_LOG"; then
+    echo "fatal error:  unable to write to file '$COMMAND_LOG'"
+    exit 1
```
anodelman: I would think that this should be a 'return 1', otherwise it quits the whole program?

tests/env.sh
```diff
@@ -330,23 +314,101 @@ check_user_vars() {
...
+  fi
+  if [ -f "$TEST_LOG_FILE.$TEST_ID" ] && [ "$BATS_TEST_COMPLETED" != "1" ]; then
+    cat "$TEST_LOG_FILE.$TEST_ID"
+  fi
+  main_log_cleanup
+}
+
+teardown_time_log() {
+  end_time=$(date +%s)
+  total_time=$((end_time - start_time))
```
anodelman: Should be defined as local vars.

tests/env.sh
```diff
@@ -330,23 +314,101 @@ check_user_vars() {
...
+  if [[ -f "$COMMAND_LOG" ]]; then
+    teardown_command_log
+  fi
+  if [ -f "$VERSITY_LOG_FILE.$TEST_ID.1" ]; then
+    teardown_versity_log 1
+  fi
+  if [ -f "$VERSITY_LOG_FILE.$TEST_ID.2" ]; then
+    teardown_versity_log 2
+  fi
+  if [ -f "$TEST_LOG_FILE.$TEST_ID" ] && [ "$BATS_TEST_COMPLETED" != "1" ]; then
```
anodelman: For safety, should use "${TEST_LOG_FILE}.${TEST_ID}", not just here but in other occurrences.

tests/versity.sh
```diff
@@ -15,52 +15,74 @@
...
+
+  for ((check_num=1; check_num<=10; check_num++)); do
+    sleep 1
+    if check_result=$(kill -0 "$1" 2>&1); then
+      return 0
+    fi
+  done
+
+  log 1 "versitygw failed to start: $check_result"
+  if [ "$2" != "" ]; then
```
anodelman: Should use [[ -n "$2" ]]

tests/versity.sh
```diff
@@ -15,52 +15,74 @@
...
+
+verify_process_started() {
+  if ! check_param_count_ge_le "pid, log file (if any)" 1 2 $#; then
+    return 1
+  fi
+  local check_result
+
+  for ((check_num=1; check_num<=10; check_num++)); do
+    sleep 1
+    if check_result=$(kill -0 "$1" 2>&1); then
```
anodelman: there is a risk here if there's a zombie process - if the gateway crashes instantly during startup due to a bad configuration flag or a blocked port, it can enter a temporary "Zombie" or "Defunct" state. Because a zombie process remains in the OS process table until its parent reads its exit status, kill -0 will continue to return a successful exit code (0). Your function will prematurely declare return 0 (Process Started), causing all subsequent S3 test assertions to fail with network connection dropouts.

You'll want some secondary checks:
```
# ensure the process isn't dead/zombie ('Z') or defunct
      # 'ps -o state=' returns a single character state indicator (e.g., S, R, D, Z)
      local proc_state
      proc_state=$(ps -p "$target_pid" -o state= 2>/dev/null | tr -d ' ')
      
      if [[ "$proc_state" != "Z" && -n "$proc_state" ]]; then
        log 5 "Process validation confirmed: PID $target_pid is active in state '$proc_state'."
        return 0
      fi
```

tests/versity.sh
```diff
@@ -151,37 +187,42 @@ run_versity_app() {
...
-    # shellcheck disable=SC2154
-    if ! stop_single_process "$versitygw_pid_2"; then
-      log 2 "error stopping versity process two"
+check_versity_process_status() {
+  local status_one="" status_two=""
+
+  status_one="none"
+  status_two="none"
+  if [ "$RUN_VERSITYGW" == "true" ]; then
+    if [[ -n "$VERSITYGW_PID_1" ]] && kill -0 "$VERSITYGW_PID_1" >/dev/null; then
```
anodelman: This also has the same zombie process pitfall as the other function.

anodelman (changes requested):

tests/env.sh
```diff
@@ -327,26 +339,109 @@ check_user_vars() {
...
+    return 1
+  fi
+  return 0
+}
+
+teardown_versity_log() {
+  if ! check_param_count_v2 "versitygw process ID" 1 $#; then
+    return 1
+  fi
+  echo "**********************************************************************************" >> "$VERSITY_LOG_FILE.$TEST_ID.$1"
```
anodelman: Minor fix to use proper braces, ${VERSITY_LOG_FILE}.${TEST_ID}.$1

tests/versity.sh
```diff
@@ -14,53 +14,84 @@
...
+    if [ "$process_running" == "true" ] && list_buckets_rest "" "" >/dev/null; then
+      return 0
+    fi
+  done
+
+  if [ "$process_running" == "false" ]; then
+    log 1 "versitygw failed to properly initialize: $check_result"
+  else
+    log 1 "process running, but not communicating properly"
+  fi
```
anodelman, line 95: Small update for zombies:

```
if [[ -n "$proc_state" && "$proc_state" != "Z" ]]; then
    log 1 "Process Health Failure: Gateway process is running (State: $proc_state), but  failed to respond"
  else
    log 1 "Process Initialization Failure: Gateway failed to start or crashed out completely: $check_result"
  fi
```

anodelman (changes requested):

tests/setup.sh
```diff
@@ -64,9 +64,24 @@ static_user_versitygw_setup() {
...
-  setup_test_log_file
+  source_config_file
+
+  if ! setup_test_log_file; then
+    log 1 "error creating test log file"
+    return 1
+  fi
+
+  if [ -n "$TEST_LOG_FILE" ]; then
+    printf "\n%s\n\n" "**** $TEST_ID ****" >> "$TEST_LOG_FILE.$TEST_ID"
```
anodelman: Oh no!  Needs more braces!  (sorry)

anodelman (dismissed):
Final nit, I'll r+ so you can just push that up and get things merged.

anodelman (approved):

PR #2170, opened 2026-06-08
fix: prevent panic in ParseCopySource on empty input

SebTardif:
## Summary

`ParseCopySource` indexes into `copySourceHeader[0]` without checking
for an empty string, causing an index-out-of-range panic. All three backend
callers (azure, posix x2) pass the `x-amz-copy-source` header value
directly, so a malformed or missing header can propagate an empty string
and crash the gateway.

## Fix

Add an empty-string guard at the top of `ParseCopySource` that returns
an `InvalidArgCopySourceBucket` error, consistent with the existing
error returned when the source path has no bucket/object separator.

## Testing

- Added a table-driven test case that reproduces the panic without the
  fix and verifies the correct error with the fix.
- All 15 `TestParseCopySource` cases pass (14 existing + 1 new).
- `go vet ./backend/` and `go build ./backend/...` clean.

### DIFF
--- a/backend/common.go
+++ b/backend/common.go
@@ -230,6 +230,10 @@ func ParseCopySourceRange(size int64, acceptRange string) (int64, int64, error) 
 // ParseCopySource parses x-amz-copy-source header and returns source bucket,
 // source object, versionId, error respectively
 func ParseCopySource(copySourceHeader string) (string, string, string, error) {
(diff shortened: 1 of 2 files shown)


versity-github:

benmcclelland (approved):

PR #2176, opened 2026-06-10
feat(helm): Make it possible to specify deployment strategy

catdog2:
This adds support to specify [Deployment Strategy](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy) inside the helm chart.

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -24,6 +24,10 @@ spec:
   selector:
     matchLabels:
       {{- include "versitygw.selectorLabels" . | nindent 6 }}
+  {{- with .Values.strategy }}
+  strategy:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   template:
     metadata:
       annotations:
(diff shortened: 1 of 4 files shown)


versity-github:

chart/templates/deployment.yaml
```diff
@@ -24,6 +24,9 @@ spec:
   selector:
     matchLabels:
       {{- include "versitygw.selectorLabels" . | nindent 6 }}
+  {{- with .Values.strategy }}
+  strategy: {{ . | toYaml | nindent 4 }}
+  {{- end }}
```
benmcclelland, line 30: This would be better to match the style of the other complex values in the file:
```yaml
{{- with .Values.strategy }}
strategy:
  {{- toYaml . | nindent 4 }}
{{- end }}
```

benmcclelland (changes requested):
Changing this needs a bump to version 0.3.2 in the Chart.yaml.
optional: A one-liner in the README.md noting that strategy: {type: Recreate} is useful for RWO volumes would give this better discoverability

benmcclelland (approved):

PR #2189, opened 2026-06-15
test: ListObjectVersions query tests

lrm25:


tests/drivers/list_object_versions/list_object_versions_rest.sh
```diff
@@ -184,3 +184,99 @@ parse_latest_version_id() {
   echo "$version_id"
   return 0
 }
+
+check_page_with_two_different_keys() {
+  if ! check_param_count_ge_le "data file, version ID marker, version ID, next version ID marker (optional)" 3 4 $#; then
+    return 1
+  fi
+  if ! check_version_page_order "$1" "KeyMarker" "Key" "NextKeyMarker" "$2" "$3" "$4"; then
+    log 2 "error checking pages for same key, different version IDs"
```
anodelman: Wrong error message, since function is looking at two different keys.

tests/drivers/objects_and_versions.sh
```diff
@@ -0,0 +1,142 @@
...
+
+list_objects_check_key() {
+  if ! check_param_count_v2 "bucket name, key, encoding type, additional query, main element, object element" 6 $#; then
+    return 1
+  fi
+  local query_params=() query
+  if [ "$3" != "" ]; then
+    query="encoding-type=$3"
+    if [ "$4" != "" ]; then
+      query+="&$4"
```
anodelman, line 92: Is there a concern here if both $3 and $4 are empty?  I believe that the query would then fail.  Maybe it should error out if there is no query information provided.

tests/test_rest_list_object_versions.sh
```diff
@@ -67,3 +68,119 @@ source ./tests/util/util_time.sh
...
+  fi
+  run setup_bucket_and_files_v3 "$BUCKET_ONE_NAME" "2"
+  assert_success
+  read -r bucket_name file_one file_two <<< "$output"
+
+  run put_object_rest "$TEST_FILE_FOLDER/$file_one" "$bucket_name" "$file_one"
+  assert_success
+
+  run put_object_rest "$TEST_FILE_FOLDER/$file_two" "$bucket_name" "$file_two"
+  assert_success
```
anodelman: Your helper function setup_bucket_and_files_v3 handles creating and pushing objects to the REST gateway dynamically. Manually running put_object_rest immediately after wastes time and creates duplicate, redundant version records on your backend storage pools.

tests/test_rest_list_object_versions.sh
```diff
@@ -67,3 +68,119 @@ source ./tests/util/util_time.sh
...
+    "-bucketName" "$bucket_name" "-query" "versions"
+  assert_success
+  mapfile -t modified_times_and_version_ids <<< "$output"
+
+  log 5 "version IDs: ${modified_times_and_version_ids[*]}"
+  local version_ids=()
+  while IFS=$'\n' read -r modified_time_and_version_id; do
+    read -r _ version_id <<< "$modified_time_and_version_id"
+    version_ids+=("$version_id")
+  done < <(printf '%s\n' "${modified_times_and_version_ids[@]}" | sort -r)
```
anodelman: S3 LastModified timestamps use ISO 8601 formatting (e.g., 2026-06-16T10:30:00.000Z). Sorting this array purely by text string using sort -r assumes that a newer string payload indicates a newer S3 transaction. If your underlying network layer or system clock experiences a slight latency shift, raw text string sorting can easily parse historical timestamps incorrectly.

Since S3 natively streams <Version> elements pre-sorted from newest to oldest, you can skip this entire fragile sort pipe entirely.

tests/test_xml.sh
```diff
@@ -92,3 +92,30 @@ source ./tests/setup_unit.sh
...
+  assert_success
+
+  run check_for_empty_element "$TEST_FILE_FOLDER/$file_three" "Outer" "Inner"
+  assert_failure 1
+
+  run check_for_empty_element "$TEST_FILE_FOLDER/$file_four" "Outer" "Inner"
+  assert_success
+
+  run check_for_empty_element "$TEST_FILE_FOLDER/$file_four" "MoreOuter" "Outer" "Inner"
+  assert_failure 1
```
anodelman: As to these last two assertions, this design is contradictory:
File Four contains raw text: <Outer></Outer>. The <Inner> node does not exist at all.

Your fourth assertion expects this condition to return success. This means your underlying helper check_for_empty_element treats a completely non-existent element as an empty element.

However, in your fifth assertion, when checking for an invalid nested tree block that also doesn't exist (MoreOuter/Outer/Inner), it expects a failure 1.

If the intent of check_for_empty_element is to return success whenever an element doesn't contain textual content, it should logically pass for both missing-node scenarios. If its true intent is to enforce that a node must physically exist in the DOM tree but contain no text data, then Assertion 4 is a false positive and should be an explicit failure.

tests/drivers/objects_and_versions.sh
```diff
@@ -0,0 +1,142 @@
...
+  bucket_name="$response"
+
+  file_name="a+ b.txt"
+  expected_encoding="a%2B+b.txt"
+  if ! create_test_file "$file_name"; then
+    log 2 "error creating test file"
+    return 1
+  fi
+
+  payload_file="$TEST_FILE_FOLDER/$file_name"
```
anodelman, line 66: Needs to be local varaible.

anodelman (changes requested):

tests/test_rest_list_object_versions.sh
lrm25: Was also worried about XMLlint, but I checked and that returns them in the same order too, so yeah, I can remove it

tests/test_xml.sh
lrm25: How about breaking them up into something like `check_for_empty_element` and `check_for_nonexistent_element_inside_parent`?

tests/drivers/objects_and_versions.sh
lrm25: It was supposed to not add any query parameters in that case, but yes, it would have passed the `-query` parameter without a value in that case, so it was broken.

tests/test_rest_list_object_versions.sh
lrm25: `setup_bucket_and_*add*_files_v3` does that, `setup_bucket_and_files_v3` does not, but replacing it with `setup_bucket_and_add_files_v3` is better anyway

tests/test_xml.sh
lrm25: The rationale is that sometimes versitygw returns the element with an empty string whereas S3 doesn't return the element at all, and I'm worried that simply checking for the nonexistent node is too lenient (e.g. it would pass if the data was actually literally blank or wrong entirely due to a previous error).

anodelman: Yes, i think that separating assert_xml_element_empty() (element path must physically exist in the DOM tree, but its inner string content must be completely empty)  and assert_xml_element_absent() (parent node exists, but the specific target sub-element must be completely missing from the output stream)

lrm25: I changed it to `check_for_empty_or_nonexistent_element`.  Is that good, or you think two separate functions is good?

tests/drivers/list_object_versions/list_object_versions_rest.sh
```diff
@@ -184,3 +184,92 @@ parse_latest_version_id() {
...
+  fi
+  if ! check_version_page_order "$1" "VersionIdMarker" "VersionId" "NextVersionIdMarker" "$2" "$3" "$4"; then
+    log 2 "error checking pages for same key, different version IDs"
+    return 1
+  fi
+  return 0
+}
+
+check_version_page_order() {
+  if ! check_param_count_ge_le "data file, previous marker name, value name, next marker name, prev value, value, next value" 6 7 $#; then
```
anodelman, line 235: documentation string describes 7 distinct parameters ($1 through $7), but your parameter verification boundary checks allow a floor count of 6 (6 7 $#). If a test calls this function with only 6 parameters, your trailing logic checks $7 (the next value) inside an if [ -z "$7" ] block. While technically allowed by the condition, evaluating unpassed, uninitialized positional indices inside generic validation engines can cause unexpected behavior depending on the shell environment.

tests/drivers/objects_and_versions.sh
```diff
@@ -0,0 +1,144 @@
...
```
anodelman, line 66: Is this resolved because you want this payload_file to be global?  Is this intentional.

tests/drivers/xml.sh
```diff
@@ -179,13 +183,39 @@ check_xml_element_inside_string() {
...
+  fi
+  xpath="$response"
+
+  result=$(echo "$1" | xmllint --xpath "boolean($xpath)" - 2>&1)
+  if [ "$result" == "false" ]; then
+    log 2 "element matching '$xpath' doesn't exist"
+    return 1
+  fi
+
+  if ! response=$(echo "$1" | xmllint --xpath "${xpath}/text()" - 2>/dev/null); then
```
anodelman: When an XML node exists but is explicitly empty (e.g., <Prefix></Prefix> or <NextVersionIdMarker/>), appending /text() forces xmllint to throw a non-zero exit status code because a text leaf node does not physically exist in that DOM sub-tree.

This function intercepts this non-zero exit status as a command execution failure, skips printing the true response variable, and blindly outputs a blank string (echo ""). While this returns the correct value for a blank node, it completely hides genuine command failures. If xmllint crashes because of system resource limits or a malformed binary flag, the function will silently swallow the error and falsely claim success (return 0).

anodelman (changes requested):

tests/drivers/objects_and_versions.sh
lrm25: It should be defined as local earlier in the function.

tests/drivers/list_object_versions/list_object_versions_rest.sh
lrm25: How about `if [ $# -lt 7 ]`?

tests/test_xml.sh
lrm25: Also, I changed it to return 0 in any condition where the element doesn't exist or is empty, and added another check to see if the parent element exists when needed.

tests/drivers/list_object_versions/list_object_versions_rest.sh
anodelman: Sure, that would work.

tests/drivers/xml.sh
```diff
@@ -79,23 +78,26 @@ get_inner_xpath_string_for_element() {
...
   if ! get_xml_data "$1" "$1.xml"; then
     log 2 "error getting XML data"
-    return 1
+    return 2
   fi
+  log 5 "middle elements: ${*:2:$#-2}"
   if grep -q '<[^/ ?>].*>' "$1.xml"; then
     if xmllint --xpath "${xpath}[not(normalize-space())]" "$1.xml" 1>/dev/null 2>&1; then
       return 0
+    elif get_element "$1.xml" "${@:2:$#-2}" && ! get_element "$1.xml" "${@:2}"; then
```
anodelman: the syntax ${@:position:length} takes a starting position and an explicit slice length, not an ending position.

If your test calls this function with 4 arguments (e.g., file.xml "Outer" "Inner"), $# equals 4.

Your slice statement evaluates to ${@:2:2}, which extracts exactly 2 arguments starting from index 2: "Outer" and "Inner".

Therefore, get_element "$1.xml" "${@:2:$#-2}" is checking if the entire path including the child exists, completely defeating your parent-presence verification check!

To dynamically isolate only the parent nodes (dropping the last argument element), you must calculate the exact length offset cleanly beforehand using basic shell arithmetic.

tests/drivers/xml.sh
```diff
@@ -79,23 +78,26 @@ get_inner_xpath_string_for_element() {
...
 }
 
-check_for_empty_element() {
+check_for_empty_or_nonexistent_element() {
   if ! check_param_count_gt "data file, XML tree" 2 $#; then
     return 1
   fi
 
   # shellcheck disable=SC2068
   if ! xpath=$(build_xpath_string_for_element ${@:2} 2>&1); then
```
anodelman: dropping double quotes here forces the shell to run word splitting on any argument values containing natural spaces or special characters. If an object prefix contains spaces, your arguments will fragment before they ever reach build_xpath_string_for_element.  use "${@:2}"

tests/drivers/xml.sh
```diff
@@ -79,23 +78,26 @@ get_inner_xpath_string_for_element() {
...
-    return 1
+    return 2
   fi
   if ! get_xml_data "$1" "$1.xml"; then
     log 2 "error getting XML data"
-    return 1
+    return 2
   fi
+  log 5 "middle elements: ${*:2:$#-2}"
   if grep -q '<[^/ ?>].*>' "$1.xml"; then
```
anodelman: If your XML data stream contains inline values that look like tags or contains an XML declaration header line like <?xml version="1.0" encoding="UTF-8"?>, this pattern can evaluate unpredictably. Since you are already leveraging xmllint natively for your inner operations, you can completely bypass this fragile string scanning condition.

tests/drivers/xml.sh
anodelman, line 105: Example fixed function

```
check_for_empty_or_nonexistent_element() {
  if ! check_param_count_gt "data file, XML tree node selectors" 2 "$#"; then
    return 1
  fi

  local data_file="$1"
  local xml_workspace="$data_file.xml"
  local xpath response parent_element_exists target_element_exists
  
  # Calculate exact slice dimensions safely before evaluating positional parameters
  # Total positional count minus the data file parameter ($1) and the target leaf node parameter
  local -i parent_nodes_length=$(( $# - 2 ))

  if ! response=$(build_xpath_string_for_element "${@:2}" 2>&1); then
    log 2 "error building XPath search string: $response"
    return 2
  fi
  xpath="$response"

  if ! get_xml_data "$data_file" "$xml_workspace"; then
    log 2 "error retrieving XML data framework from storage target"
    return 2
  fi

  log 5 "Isolated parent nodes elements slice count: ${parent_nodes_length}"

  if xmllint --xpath "${xpath}[not(normalize-space())]" "$xml_workspace" >/dev/null 2>&1; then
    rm -f "$xml_workspace"
    return 0
  fi

  if get_element "$xml_workspace" "${@:2:$parent_nodes_length}" >/dev/null 2>&1; then
    if ! get_element "$xml_workspace" "${@:2}" >/dev/null 2>&1; then
      rm -f "$xml_workspace"
      return 0
    fi
  fi

  rm -f "$xml_workspace"
  return 1
}
```

tests/test_xml.sh
```diff
@@ -92,3 +92,81 @@ source ./tests/setup_unit.sh
...
+  run get_file_names 4
+  assert_success
+  read -r file_one file_two file_three file_four <<< "$output"
+
+  printf '<Outer><Inner>text</Inner></Outer>' > "$TEST_FILE_FOLDER/$file_one"
+  printf '<Outer><Inner/></Outer>' > "$TEST_FILE_FOLDER/$file_two"
+  printf '<Outer><Inner></Inner></Outer>' > "$TEST_FILE_FOLDER/$file_three"
+  printf '<Outer></Outer>' > "$TEST_FILE_FOLDER/$file_four"
+
+  run check_xml_element "$TEST_FILE_FOLDER/$file_one" "tex" "Outer" "Inner"
```
anodelman, line 156: Add 2>&1 to the end of these check_xml_element calls to ensure that you are capturing the error logging that you are partially marching against.

anodelman (changes requested):

tests/drivers/xml.sh
lrm25: I previously decided to remove that "parent element exist, child does not" check to begin with, but accidentally left it in, so I'll just get rid of it.

tests/test_xml.sh
lrm25: It should automatically route to stdout unless using `--separate-stderr`.

tests/drivers/xml.sh
```diff
@@ -79,24 +78,29 @@ get_inner_xpath_string_for_element() {
...
 }
 
-check_for_empty_element() {
+check_for_empty_or_nonexistent_element() {
   if ! check_param_count_gt "data file, XML tree" 2 $#; then
     return 1
   fi
+  local response
 
   # shellcheck disable=SC2068
```
anodelman: Should no longer need this shellcheck since you added the necessary quotes.

tests/drivers/xml.sh
```diff
@@ -79,24 +78,29 @@ get_inner_xpath_string_for_element() {
...
+    if [[ "$response" == *"XPath set is empty"* ]]; then
       return 0
     fi
+    log 2 "error checking for empty or nonexistent param: $response"
+    return 2
+  elif xmllint --xpath "${xpath}[not(normalize-space())]" "$1.xml" 1>/dev/null 2>&1; then
+    return 0
   fi
   return 1
 }
```
anodelman, line 105: This will leak $.xml files, there's no cleanup of the temp file.

anodelman (changes requested):

tests/drivers/xml.sh
lrm25: Right now, given that the test files should be placed in `TEST_FILE_FOLDER` and cleared after tests during normal operation, this shouldn't be a major issue.  But I was thinking about changing the code in the future to rely more on passing the XML data more via strings rather than creating unnecessary files.

anodelman (approved):

PR #2199, opened 2026-06-17
feat: webui restore original intended page after login

benmcclelland:
Unauthenticated users who open a protected page are now returned to that same page after they sign in, instead of always landing on a default screen.

### DIFF
--- a/webui/web/index.html
+++ b/webui/web/index.html
@@ -621,14 +621,8 @@ under the License.
         const rememberKey = document.getElementById('remember-access-key').checked;
         saveRecentGateway(s3Endpoint, region, accessKey, rememberKey);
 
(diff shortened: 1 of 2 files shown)


benmcclelland:
This is an alternate approach to #2079 to try to satisfy the codeql gods.

apognu:
From what I see in the code and my local testing, it works to get to a top-level page, but does not record, for the explorer, the current bucket/prefix.

As a note, my initial usecase was to be able to link to a bucket or a specific prefix in a bucket.

benmcclelland:
oh, right, i was thinking it was just the top level for some reason. yeah, i dont think it will allow just any user provided path here 😞

PR #2202, opened 2026-06-22
test: README updates, add quick comparison

lrm25:


tests/drivers/head_object/head_object_rest.sh
```diff
@@ -286,3 +286,98 @@ get_delete_marker_and_verify_405() {
...
+  remote_key_size="${response_lines[3]}"
+
+  if [ "$remote_checksum" == "" ] || [ "$checksum_type" == "COMPOSITE" ]; then
+    log 5 "skipping calcuation, no checksum or COMPOSITE type"
+    return 2
+  fi
+
+  if [ "$local_file_size" -ne "$remote_key_size" ]; then
+    log 2 "file size mismatch ('$local_file_size' locally, '$remote_key_size' remotely)"
+    return 3
```
anodelman: wrong return here?  should be 1 for mismatch?

tests/drivers/head_object/head_object_rest.sh
```diff
@@ -286,3 +286,98 @@ get_delete_marker_and_verify_405() {
...
+      checksum="$value"
+    elif [[ "$lowercase_key" == "content-length" ]]; then
+      file_size="$value"
+    fi
+  done <<< "$(grep -aE '^.+: .+$' "$1")"
+
+  echo "$checksum_algorithm"
+  echo "$checksum"
+  echo "$checksum_type"
+  echo "$file_size"
```
anodelman: Looking this over, it's possible for these variables to end up being unassigned - which would then throw off the return values.  Maybe defaults here?

anodelman (changes requested):

tests/drivers/head_object/head_object_rest.sh
```diff
@@ -286,3 +286,98 @@ get_delete_marker_and_verify_405() {
...
```
anodelman: Safer to switch to 
```
echo "${checksum_algorithm:-none} ${checksum:-none} ${checksum_type:-none} ${file_size:-0}"
```

Ensures that empty variable values don't affect value passing.
Then you can use

```
mapfile -t response_lines <<< "$response"
```

tests/drivers/head_object/head_object_rest.sh
```diff
@@ -286,3 +286,98 @@ get_delete_marker_and_verify_405() {
...
+    log 2 "error getting checksum and file size: $response"
+    return 3
+  fi
+  log 5 "RESPONSE: $response"
+  mapfile -t response_lines <<< "$response"
+
+  checksum_algorithm="${response_lines[0]}"
+  remote_checksum="${response_lines[1]}"
+  checksum_type="${response_lines[2]}"
+  remote_key_size="${response_lines[3]}"
```
anodelman: switch to

```
read -r checksum_algorithm remote_checksum checksum_type remote_key_size <<< "$response"
```

anodelman (changes requested):

tests/drivers/head_object/head_object_rest.sh
lrm25: Doing this, just also setting the default file size to `none` since it could actually be zero.

PR #2206, opened 2026-06-23
Add possibility to enable sidecar directory to the chart

white-gecko:
This should be a fix for the situation described in #2204 .

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -213,6 +213,12 @@ spec:
               mountPath: /mnt/iam
               subPath: iam
               readOnly: false
+            {{- if .Values.sidecar.enabled }}
+            - name: data
+              mountPath: /mnt/metadata
+              subPath: metadata
+              readOnly: false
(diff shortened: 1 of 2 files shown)


versity-github:

chart/templates/deployment.yaml
benmcclelland, line 64: I think if sidecar is enabled, we also need an option here:
```
            - name: VGW_BACKEND_ARGS
              value: {{ ternary (printf "%s --sidecar %s" .Values.gateway.backend.args .Values.sidecar.path) .Values.gateway.backend.args (and .Values.sidecar.enabled (eq .Values.gateway.backend.type "posix")) | quote }}
```

I'm not sure if this is the best way to do this. We can also set the sidecar path with env VGW_META_SIDECAR.

benmcclelland (changes requested):
also need to bump version in chart.yaml for these changes
```
version: 0.3.4
```

benmcclelland:
this was fixed in #2227

PR #2212, opened 2026-06-26
test: add go executable, install base dependencies script

lrm25:


tests/commands/command.sh
```diff
@@ -52,4 +52,30 @@ send_command() {
...
+}
+
+generate_go_command() {
+  if ! check_param_count_gt "command params" 1 $#; then
+    return 1
+  fi
+  local response
+
+  if [ -n "$GENERATE_COMMAND_EXECUTABLE" ]; then
+    if [ ! -f "$GENERATE_COMMAND_EXECUTABLE" ] && ! response=$(go build -o "$GENERATE_COMMAND_EXECUTABLE" ./tests/rest_scripts/generateCommand.go 2>&1); then
```
anodelman: what happens right after compilation completes successfully: The function skips executing the command for the current test run - doesn't make it to the second if to do go run.  is that intended?

Also, should separate the build output from the $response - since you are using response specifically for generateCommand output and you wouldn't want that to get muddied.

tests/install_base_dependencies.sh
```diff
@@ -0,0 +1,237 @@
...
+run_as_root() {
+  if is_root; then
+    "$@"
+  elif command -v sudo >/dev/null 2>&1; then
+    sudo "$@"
+  else
+    echo "error: root privileges required and sudo not found" >&2
+    return 1
+  fi
+  return 0
```
anodelman: Don't overwrite command output this way, so remove last 'return 0' so that you get command result as return value.

tests/install_base_dependencies.sh
```diff
@@ -0,0 +1,237 @@
...
+
+install_linux_libraries() {
+  local pkg_manager
+  pkg_manager="$(detect_linux_package_manager)"
+
+  echo "Package manager: $pkg_manager"
+
+  case "$pkg_manager" in
+    apt)
+      run_as_root apt-get update
```
anodelman: Should do `run_as_root apt-get update || return 1` to short circuit execution if there is mirror slowness.  Same for other apt-get interactions

anodelman (changes requested):

tests/commands/command.sh
lrm25: If the executable is created, it should run that rather than `go run`.  However, I changed it to try `go run` as well if there's a failure building the executable.

tests/install_base_dependencies.sh
lrm25: added code to return 1 for previous errors

anodelman: Still masking those return codes that the command actually generated, instead of just returning what they do.  Just allow them to run and return.

tests/install_base_dependencies.sh
```diff
@@ -0,0 +1,241 @@
...
+
+install_linux_libraries() {
+  local pkg_manager
+  pkg_manager="$(detect_linux_package_manager)"
+
+  echo "Package manager: $pkg_manager"
+
+  case "$pkg_manager" in
+    apt)
+      run_as_root apt-get update || return 1
```
anodelman: This also needs DEBIAN_FRONTEND=noninteractive in case of repository/mirror changes.

anodelman (changes requested):

tests/install_base_dependencies.sh
lrm25: How about `|| return $?`.  Or just better to remove the `return 0`?

lrm25: NM, it's a minor semantic difference.

anodelman (approved):

PR #2217, opened 2026-06-30
azure: return quoted, consistent multipart part ETags

edespino:
## Summary

The Azure backend returns raw, unquoted block IDs as S3 multipart part ETags. S3 ETags are quoted strings, so a spec-compliant client that parses the quoted form cannot extract the part ETag, then sends an empty/invalid ETag in `CompleteMultipartUpload`, which the gateway rejects with `400 InvalidPart`. This change makes the Azure backend return spec-compliant quoted part ETags consistently across all multipart surfaces, and makes `CompleteMultipartUpload` accept both quoted and raw client ETags.

## Problem

`UploadPart` and `ListParts` returned the Azure block ID directly as the ETag, e.g.:

```
ETag: AQAAAA==
```

A client that expects the S3-standard quoted form (`ETag: "AQAAAA=="`) extracts an empty value, so the subsequent `CompleteMultipartUpload` carries empty part ETags and the gateway returns `400`. `UploadPartCopy` returned an empty `CopyPartResult` with no ETag at all.

This is inconsistent with both the S3 contract and versitygw's own posix backend, whose `GenerateEtag` already returns a quoted ETag.

## Fix

All changes are in `backend/azure/azure.go`:

- Add a small `quoteETag` helper that wraps a block ID in double quotes via `fmt.Sprintf("%q", ...)`. Azure block IDs are standard base64, whose alphabet contains no characters that `%q` would escape, so this is a plain quote wrap.
- `UploadPart`: return the quoted ETag from both return sites. The raw block ID is still passed to `StageBlock`.
- `ListParts`: quote the ETag for both regular and zero-byte parts.
- `UploadPartCopy`: populate the previously-empty `CopyPartResult` with the quoted ETag and `LastModified`, matching the posix backend's field set. The raw block ID is still passed to `StageBlockFromURL`.
- `CompleteMultipartUpload`: normalize the client-supplied ETag with `backend.TrimEtag` before comparing it against the raw Azure block ID, so both quoted and raw client ETags are accepted.

All Azure multipart part-ETag surfaces now emit the same quoted form, and Complete accepts it. Non-Azure backends are untouched.

## Backward compatibility

`CompleteMultipartUpload` accepts both quoted and raw ETags (`backend.TrimEtag` is a no-op on an already-raw value), so clients that round-trip the previously-returned raw ETag continue to work.

## Tests

Adds `UploadPart_etag_quoting_consistency` to the integration suite (runs in the Azure full-flow). It uploads multipart parts and asserts that the ETags returned by `UploadPart` are quoted, that `ListParts` reports the same quoted ETags, and that `CompleteMultipartUpload` accepts them. `UploadPartCopy` remains skipped under Azurite (its `StageBlockFromURL` returns HTTP 500); a comment notes that the real-Azure path now returns a populated, quoted `CopyPartResult`.

## Validation

Verified end to end against a live Azure Blob deployment with a real S3 client performing large (multipart) object writes. A workload that previously aborted at `CompleteMultipartUpload` now completes cleanly, and a full read/write workload passes with consistent ETags across `UploadPart`, `ListParts`, and `CompleteMultipartUpload`.


versity-github:

benmcclelland (approved):

edespino:
Thank you @benmcclelland and VersityGW

PR #2218, opened 2026-07-01
feat: add configurable default ETag for files without metadata

benmcclelland:
When versitygw serves files that were placed on the filesystem outside of the gateway, those files have no stored etag attribute and are served with an empty ETag. This causes problems for S3 clients that rely on ETags being non empty.

The new --default-etag flag (VGW_DEFAULT_ETAG) lets operators specify a fallback ETag value returned for any object that lacks a stored etag attribute, making pre-existing filesystem data behave more predictably under S3 workloads without requiring a metadata migration.

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -91,6 +91,10 @@ type Posix struct {
 	// frontend handlers are already validating bucket names
 	validateBucketName bool
 
+	// defaultEtag is returned for objects that do not have an etag attribute
+	// stored (e.g. files placed on the filesystem outside of versitygw).
+	defaultEtag string
+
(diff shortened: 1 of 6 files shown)


crabique:
Would it make sense to add a safe to use example here? It's clear that it should be a string, but not much other than that.

niksis02 (dismissed):

benmcclelland:
> Would it make sense to add a safe to use example here? It's clear that it should be a string, but not much other than that.

Actually, yes, thats a good point. Some clients assume md5 sum unless it has a "-" in the name (assuming multipart etag). Might be good to document recommendations for this.

PR #2220, opened 2026-07-01
fix: include policy-accessible buckets in ListBuckets

### LINKED ISSUE #2219: S3 ListBuckets does not return buckets accessible via bucket policy (only owned buckets)
## Description

When a user has read-only access to a bucket via a bucket policy, VersityGW's S3 `ListBuckets` API (`GET /`) returns **0 buckets**, even though the user can successfully call `ListObjects` on that bucket directly.

This breaks the bucket discovery UX: users with policy-based access cannot see the bucket in their bucket list (WebUI Explorer or any S3 client that relies on `ListBuckets` for discovery). They can only access the bucket if they already know its name and type it in manually.

## Version

- VersityGW **1.6.0**
- Build: `0aa0f0d4fc7602baa61db3428ad54f866275479c` (2026-06-26)
- Backend: posix
- IAM: internal (`--iam-dir`)

### Setup
1. Create user `owner-x` (role: `user`) — owner of bucket `my-bucket`
2. Create user `readonly-y` (role: `user`)
3. Create bucket `my-bucket` with owner `owner-x`
4. Apply bucket policy granting `readonly-y` read access:

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "ReadOnly",
      "Effect": "Allow",
      "Principal": ["readonly-y"],
      "Action": ["s3:GetObject", "s3:ListBucket", "s3:GetBucketLocation"],
      "Resource": [
        "arn:aws:s3:::my-bucket",
        "arn:aws:s3:::my-bucket/*"
      ]
    },
    {
      "Sid": "FullAccess",
      "Effect": "Allow",
      "Principal": ["owner-x"],
      "Action": "s3:*",
      "Resource": [
        "arn:aws:s3:::my-bucket",
        "arn:aws:s3:::my-bucket/*"
      ]
    }
  ]
}
```

### Test as `readonly-y`

```bash
# ListBuckets → returns 0 buckets (expected: my-bucket should appear)
GET /  →  <ListAllMyBucketsResult><Buckets></Buckets></ListAllMyBucketsResult>

# ListObjects on the bucket directly → works (28 objects returned)
GET /my-bucket?list-type=2  →  200 OK with objects
```

### Test as `owner-x`
```bash
GET /  →  returns my-bucket  ✓
```

### Test as root
```bash
GET /  →  returns all buckets  ✓
```

## Expected vs Actual

**Expected:** `ListBuckets` should return buckets the requesting principal has access to (either via ownership or via bucket policy). This matches AWS S3 behavior, where `ListBuckets` returns all buckets and per-bucket access is controlled by IAM/bucket policy.

**Actual:** `ListBuckets` returns only buckets the principal **owns**. Bucket policies grant object-level access but do not affect bucket visibility in `ListBuckets`.

## WebUI Impact

The VersityGW WebUI Explorer uses `listBucketsS3()` (a presigned `GET /`) to populate the bucket list. As a result, read-only users see **"No buckets available"** even when they have valid bucket policy access to one or more buckets.

**Workaround:** Users can manually type the bucket name into the "Enter bucket name…" field and click **Open** to access it — but the bucket never appears in their bucket list.

## Additional Notes

- Admin/root accounts (`role: admin`) see all buckets via the admin API (`PATCH /list-buckets`).
- The admin API `PATCH /list-buckets` correctly returns all buckets regardless of ownership.
- The behavior appears intentional in `ListBucketsS3` (returns only owned buckets), but it conflicts with the AWS S3 contract where `ListBuckets` returns all buckets the caller is allowed to know about.
- Related: there is no way to grant a non-owner user bucket visibility without either changing ownership (`change-bucket-owner`) or promoting them to admin.

## Suggested Fix

`ListBuckets` should evaluate bucket policies and return any bucket where the requesting principal has at least `s3:ListBucket` permission, in addition to buckets they own.

omertahaoztop:
## Summary

`ListBuckets` (S3 `GET /`) only returned buckets the requesting account owns. Buckets accessible via a bucket policy granting `s3:ListBucket` were invisible to non-owner users, breaking the WebUI Explorer and any S3 client relying on `ListBuckets` for discovery.

Fixes #2219.

## Changes

- **New shared helper** `auth.PolicyGrantsListBucket(policy, access, bucket) bool` — returns true if the bucket policy grants the account `s3:ListBucket` on the bucket. Fail-closed: empty/unparseable/denied policies return false.
- **posix** `ListBuckets` — after the ownership check fails, lazily fetches the bucket policy (`policykey` attribute) and includes the bucket if the policy grants `s3:ListBucket`.
- **s3proxy** `ListBuckets` — same pattern via `getMetaBucketObjData(metaPrefixPolicy)`.
- **azure** `ListBuckets` — same pattern via `getContainerMetaData(keyPolicy)`.

## Design notes

- **Lazy fetch**: the policy attribute is only read when `acl.Owner != account`, so the common case (user lists their own buckets) has no extra I/O.
- **Fail-closed on policy errors**: a corrupted/unreadable policy attribute on one bucket is silently skipped rather than aborting the entire `ListBuckets` response. ACL parse errors still abort (preserving prior behavior). This asymmetry is intentional — the policy read is new, and one bad bucket should not break listing for everyone.
- **No-ACL buckets**: preserved prior behavior (skipped). The policy fallback only triggers when an ACL exists but the owner differs.
- **Admin/root path**: unchanged — admins see all buckets.
- **Deny overrides Allow**: handled by the existing `isAllowed` logic in `VerifyBucketPolicy`.
- **Public principal** (`"Principal": "*"`): a bucket with a public `s3:ListBucket` policy will now appear in every user's `ListBuckets` output. This is consistent with the issue's stated intent and AWS S3 behavior.

## Testing

- **Unit tests** (`auth/bucket_policy_listbucket_test.go`): 10 table-driven cases for `PolicyGrantsListBucket` — nil/empty policy, grants to account, grants to different account, GetObject-only, `s3:*` wildcard, public principal, explicit Deny overrides Allow, resource mismatch, malformed JSON.
- **Integration test** (`tests/integration/ListBuckets.go`): `ListBuckets_with_bucket_policy` — creates a user, verifies the bucket is NOT visible before the policy, applies a bucket policy granting `s3:ListBucket`, verifies the bucket IS visible after. Registered in `group-tests.go`.
- All 7 existing `ListBuckets` integration tests pass with no regressions (verified locally against a posix backend).
- `go build`, `go vet`, and `go test ./auth/...` all pass.

### DIFF
--- a/auth/bucket_policy.go
+++ b/auth/bucket_policy.go
@@ -250,6 +250,16 @@ func ValidatePolicyDocument(policyBin []byte, bucket string, iam IAMService) err
 	return nil
 }
 
+// PolicyGrantsListBucket reports whether the bucket policy grants the account
+// s3:ListBucket permission on the bucket. Returns false for an empty or
+// unparseable policy, or when access is denied (fail-closed).
+func PolicyGrantsListBucket(policy []byte, access, bucket string) bool {
+	if len(policy) == 0 {
+		return false
+	}
+	return VerifyBucketPolicy(policy, access, bucket, "", nil, ListBucketAction) == nil
+}
+
 func VerifyBucketPolicy(policy []byte, access, bucket, object string, normalizeObjectKey objectKeyNormalizer, actions ...Action) error {
 	if len(actions) == 0 {
 		return s3err.GetAPIError(s3err.ErrAccessDenied)
--- a/auth/bucket_policy_listbucket_test.go
+++ b/auth/bucket_policy_listbucket_test.go
@@ -1,0 +1,133 @@
+// Copyright 2023 Versity Software
+// This file is licensed under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPolicyGrantsListBucket(t *testing.T) {
+	const bucket = "my-bucket"
+	const account = "alice"
+
+	// policyStmt builds a single-statement bucket policy JSON document.
+	policyStmt := func(effect, principal, action, resource string) []byte {
+		return []byte(fmt.Sprintf(`{
+			"Version": "2012-10-17",
+			"Statement": [{
+				"Effect": "%s",
+				"Principal": %s,
+				"Action": %s,
+				"Resource": %s
+			}]
+		}`, effect, principal, action, resource))
+	}
+
+	// multiStmt builds a two-statement policy (e.g. Allow + Deny).
+	multiStmt := func(stmts ...string) []byte {
+		stmtsJSON := ""
+		for i, s := range stmts {
+			if i > 0 {
+				stmtsJSON += ","
+			}
+			stmtsJSON += s
+		}
+		return []byte(fmt.Sprintf(`{
+			"Version": "2012-10-17",
+			"Statement": [%s]
+		}`, stmtsJSON))
+	}
+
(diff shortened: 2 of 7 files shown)


versity-github:

benmcclelland:
This is not correct behavior according to the S3 spec.  See the notes in #2219 for the WebUI workaround using bucket favorites.

PR #2221, opened 2026-07-01
feat: add AWS-compatible standalone IAM service

### LINKED ISSUE #1640: [Feature] - AWS compatible IAM system
This task introduces the initial scope of the AWS IAM–compatible system.
The goal is to establish a clear contract that mirrors AWS IAM behavior and naming conventions, serving as the foundation for later implementation layers.

### **Scope**

* Define IAM interface methods for basic user lifecycle operations:

  * `CreateUser`
  * `UpdateUser`
  * `DeleteUser`
  * `GetUser`
  * `ListUsers`
* Define data structures for IAM user entities:

  * User name, ARN, path, unique ID, and metadata (e.g., creation date)
* Define AWS-compatible error types (e.g., `EntityAlreadyExists`, `NoSuchEntity`, `LimitExceeded`)
* Document method signatures, expected inputs, and outputs aligned with AWS IAM APIs
* Ensure naming, structure, and behavior are consistent with AWS IAM specifications

### **Out of Scope:**

* IAM policies (inline or managed)
* OIDC authentication or external identity providers
* Access keys, login profiles, or credential management
* Persistence or API implementation details

**Additional context**
* [AWS IAM](https://docs.aws.amazon.com/singlesignon/latest/userguide/what-is.html)
* [AWS IAM: CreateUser](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateUser.html)
* [AWS IAM: UpdateUser](https://docs.aws.amazon.com/IAM/latest/APIReference/API_UpdateUser.html)
* [AWS IAM: DeleteUser](https://docs.aws.amazon.com/IAM/latest/APIReference/API_DeleteUser.html)
* [AWS IAM: GetUser](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetUser.html)
* [AWS IAM: ListUsers](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListUsers.html)
* [AWS IAM: Common Errors](https://docs.aws.amazon.com/IAM/latest/APIReference/CommonErrors.html)

niksis02:
Closes #1640

Add a standalone AWS IAM Query API implementation for managing IAM users through standard AWS SDKs and the AWS CLI.

Server usage

Start the IAM server with internal file-backed storage:

    mkdir -p /tmp/versitygw-iam
    ./versitygw --port 127.0.0.1:7070 --access user --secret pass iam --dir /tmp/versitygw-iam

Start the IAM server with Vault KV v2 storage using AppRole:

    VGW_IAM_VAULT_ROLE_SECRET=<role-secret> ./versitygw --port 127.0.0.1:7070 --access user --secret pass iam --vault-endpoint-url http://127.0.0.1:8200 --vault-auth-method approle --vault-role-id <role-id> --vault-mount-path kv --vault-secret-storage-path iam

Vault authentication also supports root tokens, separate authentication and secret-storage namespaces, custom mount paths, server certificate validation, and mutual TLS client certificates.

Configure the AWS CLI credentials used by the IAM server:

    export AWS_ACCESS_KEY_ID=user
    export AWS_SECRET_ACCESS_KEY=pass
    export AWS_DEFAULT_REGION=us-east-1

Implemented IAM actions

CreateUser creates an IAM user with an AWS-compatible ARN, generated AIDA user ID, creation timestamp, optional path, and tags. It validates usernames, paths, tag limits, reserved tag prefixes, duplicate tag keys, and existing users.

    aws --endpoint-url http://127.0.0.1:7070 iam create-user --user-name bob

    aws --endpoint-url http://127.0.0.1:7070 iam create-user --user-name bob --path /engineering/ --tags Key=team,Value=storage

GetUser returns a stored user or the root identity when requested without a username through the IAM Query API.

    aws --endpoint-url http://127.0.0.1:7070 iam get-user --user-name bob

ListUsers returns users in deterministic username order and supports path filtering, marker-based pagination, and MaxItems limits.

    aws --endpoint-url http://127.0.0.1:7070 iam list-users

    aws --endpoint-url http://127.0.0.1:7070 iam list-users --path-prefix /engineering/ --max-items 100

UpdateUser updates the username and/or path, recalculates the user ARN, and rejects conflicts with existing users.

    aws --endpoint-url http://127.0.0.1:7070 iam update-user --user-name bob --new-user-name robert --new-path /platform/

DeleteUser permanently removes an IAM user and returns AWS-compatible errors for missing users.

    aws --endpoint-url http://127.0.0.1:7070 iam delete-user --user-name robert

IAM protocol and authentication

- Support the AWS IAM Query protocol version 2010-05-08 over GET and POST form requests.
- Return AWS-compatible XML responses, error documents, status codes, request IDs, user metadata, and pagination fields.
- Authenticate root credentials with AWS Signature Version 4 for the IAM service in us-east-1.
- Support both Authorization-header and query-string SigV4 authentication.
- Validate credential scope, signed headers, timestamps, clock skew, content length, signatures, and unsupported signature or session-token modes.
- Add IAM-specific validation and error mapping for malformed requests, invalid actions, duplicate entities, missing users, throttling, and internal failures.

Storage implementations

- Add an internal JSON-backed store using iam.json and iam.json.backup with atomic temporary-file replacement, concurrent access protection, stable ordering, pagination, and persistence across restarts.
- Add a Vault KV v2 store with one secret per user, CAS-based duplicate protection, permanent deletion, AppRole reauthentication, namespace support, configurable authentication and KV mounts, root-token authentication, and TLS/mTLS configuration.
- Introduce a common Storer interface and require exactly one storage backend to be configured.

Server and embedding support

- Register the new `versitygw iam` command with environment-variable and CLI configuration for both storage backends.
- Add `embedgw.RunIAMAPI` and `IAMConfig` for embedding the IAM service in Go applications.

Gateway-level internal packages

- Add `internal/iamstore` as a reusable generic file-backed IAM persistence engine and migrate the existing gateway internal IAM service to it.
- Add `internal/sigv4auth` for shared SigV4 header and presigned-query parsing, canonical request generation, signature verification, and structured authentication errors.
- Refactor the S3 authentication paths to use the shared SigV4 implementation while preserving S3-specific error responses.
- Add `internal/httpctx` for shared Fiber context keys and AWS-style request ID handling.
- Add `internal/routekit` for shared query, form, and header route matchers.
- Add `internal/netutil` for reusable certificate storage, hostname-aware listeners, multi-address serving, TLS listeners, and UNIX socket handling.
- Update the custom SigV4 signer to honor an explicitly supplied signed-header list so unrelated headers do not alter IAM signatures.

Testing and CI

- Add AWS IAM SDK-based integration coverage for all supported user actions, header authentication, query authentication, validation, errors, filtering, and pagination.
- Split standalone IAM tests into `versitygw test iam` and retain existing gateway IAM tests under `versitygw test gw-iam`.
- Add unit coverage for controllers, authentication, routing, storage, embedding, listeners, request matching, persistence, and signing behavior.
- Add `runiamtests.sh` to exercise internal storage over HTTP and HTTPS plus Vault storage through AppRole.
- Add a dedicated IAM functional-test workflow with a Vault service and merged runtime coverage reporting.
- Include the IAM test runner in shellcheck and add the AWS IAM SDK dependency.


iamapi/storage/vault.go
```diff
@@ -0,0 +1,435 @@
...
+		}
+	} else {
+		// Delete all versions then re-create so CAS=0 succeeds.
+		if err := s.deleteByPath(input.UserName); err != nil {
+			return nil, err
+		}
+		if _, err := s.CreateUser(ctx, *user); err != nil {
+			return nil, err
+		}
+	}
```
benmcclelland: failures in here can leave user in an inconsistent state.
Rename flow creates new user, then deletes old user. If delete fails, both entries remain.
Same-name update flow deletes first, then recreates. If create fails, the user is lost.

just flagging this for now. the complexity of a fix might not be worth it in the initial pass due to lack of transactional support in vault.

I think to do this in vault:
Same-key update
Use KV v2 write with CAS set to the current version (or CAS=0 for create-only semantics).
That gives optimistic concurrency and idempotent retry behavior for one key.

Rename (old key -> new key)
Create-new + delete-old cannot be one atomic commit in KV v2.
You need application-level compensation/saga logic:
Write target key with CAS constraints.
Mark operation state (journal/marker record).
Delete source key.
On retry/recovery, consult marker and complete or roll back deterministically.

iamapi/server.go
```diff
@@ -0,0 +1,207 @@
...
+
+func WithSocketPerm(perm os.FileMode) Option {
+	return func(s *IAMApiServer) { s.socketPerm = perm }
+}
+
+func WithOnListen(fn func()) Option {
+	return func(s *IAMApiServer) { s.onListen = fn }
+}
+
+func WithRootUserCreds(root RootCredentials) Option {
```
benmcclelland: do we want this to be an optional setup, or should we always require RootCredentials be provided to New()?

iamapi/internal/iammiddleware/auth.go
```diff
@@ -0,0 +1,258 @@
...
+
+func parseContentLength(contentLengthStr string) (int64, error) {
+	if contentLengthStr == "" {
+		return 0, nil
+	}
+
+	contentLength, err := strconv.ParseInt(contentLengthStr, 10, 64)
+	if err != nil {
+		return 0, iamerr.GetAPIError(iamerr.ErrInvalidContentLength)
+	}
```
benmcclelland, line 308: should we also return this error for negative contentLength values?

niksis02: Actually the code execution never reaches here when invalid Content-Length is provided, because the fasthttp parser throws the error to the global error handler, and it returns a raw 400 status.

cmd/versitygw/main.go
```diff
@@ -783,6 +804,60 @@ func initFlags() []cli.Flag {
...
+			Usage:       "access key this gateway signs its own calls to the standalone IAM service with (defaults to --access/root)",
+			EnvVars:     []string{"VGW_IAM_STANDALONE_ACCESS"},
+			Destination: &standaloneIAMAccess,
+		},
+		&cli.StringFlag{
+			Name:        "iam-standalone-secret",
+			Usage:       "secret key this gateway signs its own calls to the standalone IAM service with (defaults to --secret/root)",
+			EnvVars:     []string{"VGW_IAM_STANDALONE_SECRET"},
+			Destination: &standaloneIAMSecret,
+		},
```
benmcclelland, line 824: iam.go:198 constructs the private API with only RootUserAccess / RootUserSecret, and server.go:104 applies VerifyRootOnlySigV4.
Therefore any dedicated value differing from the root access key is rejected before the private endpoint handles the request.
should we just get rid of these for now, or do you think its important to have the alternative static credentials?

embedgw/iam.go
```diff
@@ -0,0 +1,497 @@
...
+func newPrivateAPI(store storage.Storer, cfg *IAMConfig) (*private.PrivateAPI, netutil.TLSOptions, error) {
+	allSet := cfg.PrivateCertFile != "" && cfg.PrivateKeyFile != "" && cfg.PrivateClientCAFile != ""
+	noneSet := cfg.PrivateCertFile == "" && cfg.PrivateKeyFile == "" && cfg.PrivateClientCAFile == ""
+	if !allSet && !noneSet {
+		return nil, netutil.TLSOptions{}, fmt.Errorf("--private-cert, --private-cert-key, and --private-client-ca must all be set together, or all left empty for a unix-socket-only private listener")
+	}
+
+	var tlsOpts netutil.TLSOptions
+	if allSet {
+		cs := netutil.NewCertStorage()
```
benmcclelland: this creates a separate CertStorage for the private listener but retains only its GetCertificate callback in privateTLSOpts.
after key rotation, new gateway-to-IAM TLS connections continue to receive the old certificate until the IAM service restarts.
Return/store the private CertStorage alongside privateTLSOpts, then reload both stores in the SIGHUP branch.

iamapi/server.go
```diff
@@ -0,0 +1,218 @@
...
+		var ln net.Listener
+		var err error
+
+		if s.CertStorage != nil {
+			ln, err = netutil.NewMultiAddrTLSListener(fiber.NetworkTCP, portSpec, s.CertStorage.GetCertificate, netutil.ListenerOptions{SocketPerm: s.socketPerm})
+		} else {
+			ln, err = netutil.NewMultiAddrListener(fiber.NetworkTCP, portSpec, netutil.ListenerOptions{SocketPerm: s.socketPerm})
+		}
+		if err != nil {
+			return fmt.Errorf("failed to bind iam listener %s: %w", portSpec, err)
```
benmcclelland, line 192: minor since we are probably exiting shortly after this, but we may want to cleanup any previously successful listeners cleanly here

iamapi/private/listener.go
```diff
@@ -0,0 +1,68 @@
...
+	for _, addr := range addrs {
+		var ln net.Listener
+		var err error
+		if netutil.IsUnixSocketPath(addr) && tlsOpts.GetCertificate == nil {
+			ln, err = netutil.NewMultiAddrListener(fiber.NetworkTCP, addr, netutil.ListenerOptions{SocketPerm: p.socketPerm})
+		} else {
+			ln, err = netutil.NewMultiAddrTLSListenerWithOptions(fiber.NetworkTCP, addr, tlsOpts, netutil.ListenerOptions{SocketPerm: p.socketPerm})
+		}
+		if err != nil {
+			return fmt.Errorf("failed to bind private iam listener %s: %w", addr, err)
```
benmcclelland, line 58: minor since we are probably exiting shortly after this, but we may want to cleanup any previously successful listeners cleanly here

auth/iam_standalone.go
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	access := cfg.Access
+	if access == "" {
+		access = rootAcc.Access
+	}
+	secret := cfg.Secret
+	if secret == "" {
+		secret = rootAcc.Secret
+	}
```
benmcclelland: Access and secret default independently. Supplying only --iam-standalone-access combines that access key with the root secret
may need a check that either both or neither are supplied

iamapi/private/server.go
```diff
@@ -0,0 +1,124 @@
...
+// WithPrivateQuiet suppresses per-request summary logging, mirroring
+// iamapi.WithQuiet for the public API. Callers should gate both on the same
+// flag so the two log streams turn on and off together.
+func WithPrivateQuiet() PrivateAPIOption {
+	return func(p *PrivateAPI) { p.quiet = true }
+}
+
+// New constructs the private endpoint set. root is the identity these
+// endpoints authenticate every request against.
+func New(store storage.Storer, root iammiddleware.RootCredentials, opts ...PrivateAPIOption) (*PrivateAPI, error) {
```
benmcclelland, line 125: Treat the private API as a versioned internal contract. It is intentionally used by independently deployed gateway and IAM processes, but has no protocol version or compatibility handshake. The current response-shape checks fail closed, which is good; a small version endpoint/header would make incompatible rolling upgrades diagnosable rather than merely failing requests.

benmcclelland (approved):

PR #2222, opened 2026-07-02
test: REST - ListParts queries

lrm25:


tests/commands/create_presigned_url.sh
```diff
@@ -1,27 +1,31 @@
...
     return 1
   fi
   if [[ $presign_result -ne 0 ]]; then
-    log 2 "error generating presigned url: $presigned_url"
+    log 2 "error generating presigned url: $response"
     return 1
   fi
-  export presigned_url
+  if [ "$1" == 'mc' ]; then
+    presigned_url="${response#*Share: }"
```
anodelman: Strip the prefix, but also remove anything trailing or newlines.
```

    presigned_url=$(echo "$response" | grep "Share: " | sed 's/.*Share: //')
    presigned_url="${presigned_url//$'\r'/}"
```

tests/drivers/get_object_lock_config/get_object_lock_config_rest.sh
```diff
@@ -15,21 +15,51 @@
...
+  local response response_code response_data enabled_value
+
+  if ! response=$(get_object_lock_configuration_rest "$1" 2>&1); then
+    log 2 "error getting object lock config: $response"
+    return 1
+  fi
+  read -r response_code response_data <<< "$response"
+
+  if ! response=$(xmllint --xpath '//*[local-name()="ObjectLockEnabled"]/text()' "$response_data" 2>&1); then
+    log 2 "error getting object lock config enabled value: $response"
```
anodelman, line 55: if you pass a variable name directly as the trailing argument to xmllint, it expects that variable to hold a physical file path string (e.g., /tmp/response.xml).

hardened xmlint call:

```
if ! response=$(xmllint --xpath '//*[local-name()="ObjectLockEnabled"]/text()' - <<< "$response_data" 2>&1); then
    log 2 "error getting object lock config enabled value from XML payload: $response"
    return 1
  fi
```

tests/drivers/list_parts/list_parts_rest.sh
```diff
@@ -129,3 +129,76 @@ upload_check_part() {
...
+
+  if ! response=$(get_element "$1" "ListPartsResult" 2>&1); then
+    log 2 "error getting response: $response"
+    return 1
+  fi
+  list_parts_result="$response"
+  shift
+
+  local xml_array=("Bucket" "Key" "UploadId" "MaxParts" "PartNumberMarker" "NextPartNumberMarker")
+  for xml_name in "${xml_array[@]}"; do
```
anodelman: Using the shift doesn't just clear out the metadata arguments; it strips away your trailing payload arguments - before you ever reach the part calculation block, those position variables are completely empty, causing a shell integer breakdown.

validate elements using explicit local targets rather than shifting positional variables


```
  if ! check_xml_element_inside_string "$list_parts_result" "$expected_bucket" "Bucket" >/dev/null 2>&1 ||
     ! check_xml_element_inside_string "$list_parts_result" "$expected_key" "Key" >/dev/null 2>&1 ||
     ! check_xml_element_inside_string "$list_parts_result" "$expected_upload_id" "UploadId" >/dev/null 2>&1 ||
     ! check_xml_element_inside_string "$list_parts_result" "$expected_max_parts" "MaxParts" >/dev/null 2>&1 ||
     ! check_xml_element_inside_string "$list_parts_result" "$part_number_marker" "PartNumberMarker" >/dev/null 2>&1 ||
     ! check_xml_element_inside_string "$list_parts_result" "$next_part_number_marker" "NextPartNumberMarker" >/dev/null 2>&1; then
    log 2 "error: XML root metadata header fields verification failed"
    return 1
  fi
```

tests/drivers/list_parts/list_parts_rest.sh
```diff
@@ -129,3 +129,76 @@ upload_check_part() {
...
+  fi
+  if [ ${#parts[@]} -ne "$part_count" ]; then
+    log 2 "part count mismatch, expected '$part_count, actual '${#parts[@]}'"
+    return 1
+  fi
+  if [ $((part_count*2)) -ne "$#" ]; then
+    log 2 "part count and etag/size mismatch, expected $((part_count*2)) fields, actual is $#"
+    return 1
+  fi
+  for ((i=0; i<part_count; i++)); do
```
anodelman, line 185: Again, avoid the destructive shift - in this case it doens't cause downstream harm but is still a risky way to process the data.

```
local -i pair_arg_index=8
  local -i i
  for (( i = 0; i < part_count; i++ )); do
    local part_number=$(( part_number_marker + 1 + i ))
    local expected_etag="${!pair_arg_index}"
    pair_arg_index+=1
    local expected_size="${!pair_arg_index}"
    pair_arg_index+=1

    if ! check_xml_element_inside_string "${parts[$i]}" "$part_number" "PartNumber" >/dev/null 2>&1; then
      log 2 "error checking PartNumber mismatch at segment positional offset: $part_number"
      return 1
    fi
    if ! check_xml_element_inside_string "${parts[$i]}" "$expected_etag" "ETag" >/dev/null 2>&1; then
      log 2 "error checking ETag for part sequence $part_number (Expected: $expected_etag)"
      return 1
    fi
    if ! check_xml_element_inside_string "${parts[$i]}" "$expected_size" "Size" >/dev/null 2>&1; then
      log 2 "error checking Size for part sequence $part_number (Expected: $expected_size)"
      return 1
    fi
  done
```

tests/drivers/xml_string.sh
```diff
@@ -0,0 +1,154 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+build_xpath_string() {
+  if ! check_param_count_gt "XML tree" 1 $#; then
```
anodelman, line 18: xpath not declared locally, and doesn't have a default value.

tests/drivers/xml_string.sh
```diff
@@ -0,0 +1,154 @@
...
+  fi
+
+  local response xpath
+  if ! response=$(build_xpath_string_for_element "${@:2}" 2>&1); then
+    log 2 "error building XPath search string: $response"
+    return 1
+  fi
+  xpath="$response"
+
+  if ! response=$(echo "$1" | xmllint --xpath "${xpath}" - 2>&1); then
```
anodelman: replace naked echo pipe with a safe herestring to immunize against flag injections

```
  if ! response=$(xmllint --xpath "${xpath}" - <<< "$xml_content" 2>&1); then
    if [[ "$response" == *"XPath set is empty"* ]]; then
      echo ""
      return 0
    fi
    log 2 "error getting element node matching path: $response"
    return 1
  fi
```

tests/drivers/xml_string.sh
```diff
@@ -0,0 +1,154 @@
...
+  local response xpath result
+
+  if ! response=$(build_xpath_string_for_element "${@:2}" 2>&1); then
+    log 2 "error building XPath search string: $response"
+    return 1
+  fi
+  xpath="$response"
+
+  result=$(echo "$1" | xmllint --xpath "boolean($xpath)" - 2>&1)
+  if [ "$result" == "false" ]; then
```
anodelman, line 142: secure boolean existence assertion check using an injection-safe herestring

```
  element_exists=$(xmllint --xpath "boolean(${xpath})" - <<< "$xml_content" 2>&1)
  if [[ "$element_exists" == "false" ]]; then
    log 2 "element matching path selector target string '${xpath}' does not exist inside document"
    return 1
  fi
```

extract text content safely without option-flag hijacking bugs

```
  if ! response=$(xmllint --xpath "${xpath}/text()" - <<< "$xml_content" 2>&1); then
    if [[ "$response" == *"XPath set is empty"* ]]; then
      echo ""
      return 0
    fi
    log 2 "error extracting structural element leaf text value: $response"
    return 1
  fi
```

tests/test_rest_list_parts.sh
```diff
@@ -0,0 +1,77 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+load ./bats-support/load
+load ./bats-assert/load
+
+source ./tests/drivers/create_bucket/create_bucket_rest.sh
+source ./tests/setup.sh
+
+@test "REST - ListParts - invalid part number marker" {
```
anodelman, line 23: Ensure variables have base definitions

```
  local bucket_name file_name upload_id response file_item
  local -a parts=()
```

tests/test_rest_list_parts.sh
```diff
@@ -0,0 +1,77 @@
...
+@test "REST - ListParts - part number marker" {
+  run setup_bucket_and_large_file_v3 "$BUCKET_ONE_NAME"
+  assert_success
+  read -r bucket_name file_name <<< "$output"
+
+  run create_multipart_upload_rest "$bucket_name" "$file_name" "" "parse_upload_id"
+  assert_success
+  upload_id=$output
+
+  local parts=() sizes=() etags=()
```
anodelman:
```
  local bucket_name file_name upload_id chunk_item i
  local -a parts=()
  local -a sizes=()
  local -a etags=()
```

anodelman (changes requested):

tests/drivers/get_object_lock_config/get_object_lock_config_rest.sh
lrm25: Actually already is a file, I'll change the variable name

anodelman (approved):

PR #2225, opened 2026-07-03
fix: copy-source parsing mismatch that could bypass path validation

benmcclelland:
The copy-source header was being interpreted differently by request validation and the backend parser. An encoded ?versionId= separator could be decoded at one layer but treated as part of the object path at another, which opened a path traversal risk on filesystem-backed copy operations.

Align backend copy-source parsing with validation by decoding the header before splitting the versionId suffix, so both layers derive the same bucket, object, and versionId values. This closes the traversal path and adds regression coverage for encoded copy-source inputs.

Reported by 5ud0 / Tarmo Technologies.

### DIFF
--- a/backend/common.go
+++ b/backend/common.go
@@ -238,28 +238,20 @@ func ParseCopySource(copySourceHeader string) (string, string, string, error) {
 		copySourceHeader = copySourceHeader[1:]
 	}
 
-	// Split the raw header on the versionId query parameter before any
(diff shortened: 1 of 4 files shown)


backend/common.go
niksis02, line 1: It would be nice to have this case in integration tests.

I would add these cases in `Versioning_CopyObject_invalid_versionId` and `Versioning_UploadPartCopy_invalid_versionId`

niksis02 (approved):

PR #2226, opened 2026-07-03
fix: LDAP access key filter injection

benmcclelland:
LDAP-backed IAM lookups built a search filter using the request access key before signature verification, allowing filter metacharacters to alter query semantics. This made account lookup behavior observable and could be abused as a pre-authentication oracle in affected LDAP deployments.

This change escapes access key values with ldap.EscapeFilter when constructing LDAP search filters and adds a regression test that verifies metacharacters are encoded rather than interpreted as filter operators.

Reported by GitHub user KrisKennawayDD.

### DIFF
--- a/auth/iam_ldap.go
+++ b/auth/iam_ldap.go
@@ -137,7 +137,7 @@ func (ld *LdapIAMService) CreateAccount(account Account) error {
 	if ld.rootAcc.Access == account.Access {
 		return ErrUserExists
 	}
-	userEntry := ldap.NewAddRequest(fmt.Sprintf("%v=%v,%v", ld.accessAtr, account.Access, ld.queryBase), nil)
+	userEntry := ldap.NewAddRequest(ld.buildUserDN(account.Access), nil)
 	userEntry.Attribute("objectClass", ld.objClasses)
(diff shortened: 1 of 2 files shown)


auth/iam_ldap.go
niksis02: This issue has been aging forever in my to-do list, but since we’re fixing the filter escaping, it also makes sense to add DN escaping here.

Without escaping, DN metacharacters such as `,`, `+`, and `=` in the access key could alter the DN structure. This may produce a malformed DN or cause create/update/delete operations to target an unintended LDAP entry. It’s not as risky as the filter injection because these operations require admin authorization, but it’s still worth addressing.

e.g
```go
userEntry := ldap.NewAddRequest(fmt.Sprintf("%v=%v,%v", ld.accessAtr, ldap.EscapeDN(account.Access), ld.queryBase), nil)
```

niksis02: Same here about `ldap.EscapeDN(access)`

niksis02: And here: `ldap.EscapeDN(access)`

niksis02 (changes requested):

niksis02 (approved):

PR #2234, opened 2026-07-09
feat: Add 'sessionAffinity' and 'sessionAffinityConfig'

Mart-Kuc:
This PR adds support for `sessionAffinity` and `sessionAffinityConfig` for Kubernetes Services.

By default, when an application is running with multiple instances (HA), client requests are distributed across the pods using a round-robin strategy.

However, some applications require session persistence. For example, if a client uploads or creates data on one pod and then sends a follow-up request, that request may be routed to a different pod where the data is not available yet, resulting in unexpected behavior.

Enabling `sessionAffinity` ensures that requests from the same client are consistently routed to the same pod.

This feature is also useful for scenarios such as multipart file uploads, where all requests belonging to the same upload session should be handled by the same application instance.

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: versitygw
 description: A Helm chart for deploying the Versity S3 Gateway on Kubernetes
 type: application
-version: 0.3.4
+version: 0.3.5
 sources:
   - https://github.com/versity/versitygw
 icon: https://raw.githubusercontent.com/versity/versitygw/main/webui/web/assets/images/Versity-logo-blue-horizontal.png
--- a/chart/templates/service.yaml
+++ b/chart/templates/service.yaml
@@ -6,6 +6,13 @@ metadata:
     {{- include "versitygw.labels" . | nindent 4 }}
 spec:
   type: {{ .Values.service.type }}
+  {{- if .Values.service.sessionAffinity }}
+  sessionAffinity: {{ .Values.service.sessionAffinity }}
+  {{- end }}
+  {{- with .Values.service.sessionAffinityConfig }}
+  sessionAffinityConfig:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   ports:
     - port: {{ .Values.gateway.port }}
       targetPort: s3-api
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -56,6 +56,15 @@ securityContext:
 # To make the server available outside the cluster, the type should be changed to `NodePort` or `LoadBalancer`.
 service:
   type: ClusterIP
+  # Session affinity for the Service. Supported values: "None" (default) or "ClientIP".
+  # Use "ClientIP" to route requests from the same client to the same pod.
+  sessionAffinity: ""
+  # Configuration for the session affinity (only used when sessionAffinity is "ClientIP").
+  # Example:
+  #   sessionAffinityConfig:
+  #     clientIP:
+  #       timeoutSeconds: 10800
+  sessionAffinityConfig: {}
 
 # --- Ingress ---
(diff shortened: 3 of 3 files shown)


versity-github:

benmcclelland:
Can you describe the scenario where data is not consistent across pods? The assumption from versitygw is that the backend storage system is immediately consistent across the cluster.

daleckystepan:
I hit the same issue. Assume 2 replicas of VersityGW, mounted the same PVC backed by NFS share. Under the standard situation Kubernetes Service requests are load-balanced to replicas using round robin. Write which is immediately followed by read caused the issue - data were not there (on the other replica). We are using Velero for backups which connects to S3 provided by VersityGW.

sessionAffinity fixed it and we have not hit the issue since changed manually.

Mart-Kuc:
We encountered an issue when running two instances of versityGW. A client uploaded data to one instance, but when it requested the same data afterwards, the request was routed to the other instance, where the data was not available.

However, after reviewing our configuration, we discovered a misconfiguration. We were running two VersityGW instances while using a RWO for the PVC. Maybe The question could be why the application continued to run instead of failing (as this configuration is not valid).

We have now updated the PVC to use **RWX**. We'll monitor the system and see whether the same issue still occurs

benmcclelland:
> We encountered an issue when running two instances of versityGW. A client uploaded data to one instance, but when it requested the same data afterwards, the request was routed to the other instance, where the data was not available.
> 
> However, after reviewing our configuration, we discovered a misconfiguration. We were running two VersityGW instances while using a RWO for the PVC. Maybe The question could be why the application continued to run instead of failing (as this configuration is not valid).
> 
> We have now updated the PVC to use **RWX**. We'll monitor the system and see whether the same issue still occurs

Ah yes, we probably want to prevent the invalid config if possible here. I can allow the session affinity too, but I'm concerned that it will hide real issues like this.

benmcclelland:
> I hit the same issue. Assume 2 replicas of VersityGW, mounted the same PVC backed by NFS share. Under the standard situation Kubernetes Service requests are load-balanced to replicas using round robin. Write which is immediately followed by read caused the issue - data were not there (on the other replica). We are using Velero for backups which connects to S3 provided by VersityGW.
> 
> sessionAffinity fixed it and we have not hit the issue since changed manually.

As far as I know, a read on any instance immediately following a write on NFS should be fully consistent without the need to wait for any background sync. The close() on nfs should flush all data to the server before it returns. This blocks the object put response such that the data should be flushed to the server by the time the s3 client gets the successful put response.

daleckystepan:
Not sure if this kopia error might be related or not. I noticed this once I reverted back the `sessionAffinity: None`. This is produced by Velero/kopia maintenance job pointed to S3 provided by VersityGW. RWX is already fixed.

```
Error (time="2026-07-26T21:41:39Z" level=info msg="Start to open repo for maintenance, allow index write on load" logSource="pkg/repository/udmrepo/kopialib/lib_repo.go:211" time="2026-07-26T21:41:40Z" level=info msg="Succeeded to open repo for maintenance" logSource="pkg/repository/udmrepo/kopialib/lib_repo.go:218" Repo maintenance error: failed to prune repo: error to prune backup repo: error to maintain repo: error to run maintenance under mode auto: error checking for clock skew: clock skew detected: local clock is out of sync with repository timestamp by more than allowed 5m0s (local: 2026-07-26 21:41:40.774304975 +0000 UTC repository: 2026-07-26 13:37:38 +0000 UTC skew: 8h4m2.774304975s). Refusing to run maintenance.)
```

daleckystepan:
I have verified that `sessionAffinity: ClientIP` fixed the issue in the comment above. I also need to admit that I do not fully understand the root cause.

benmcclelland:
closing this for now since we really can only support fully consistent backends, but I'm fine with revisiting this if needed for some other reason.

Issue #2235, opened 2026-07-09
[Bug] -  s3proxy backend panics with nil pointer dereference on ListMultipartUploads and ListParts


jamalveli:
The s3proxy backend panics with runtime error: invalid memory address or nil pointer dereference on essentially every call to ListMultipartUploads and ListParts, when it forwards the request to a real S3 backend and rebuilds the response.

Both S3Proxy.ListMultipartUploads and S3Proxy.ListParts in backend/s3proxy/s3.go unconditionally dereference several optional *string/*int32/*bool fields from the AWS SDK response (KeyMarker, UploadIdMarker, NextKeyMarker, NextUploadIdMarker, Delimiter, Prefix, MaxUploads/MaxParts, IsTruncated, PartNumberMarker/NextPartNumberMarker, Initiator, Owner) with no nil check. Per the AWS SDK's own doc comments, several of these are only populated conditionally:

- Delimiter: "If you don't specify a delimiter in your request, this element is absent from the response."
- NextKeyMarker/NextUploadIdMarker/NextPartNumberMarker: "When a list is truncated, this element specifies..." — absent when IsTruncated=false.
- Prefix: "When a prefix is provided in the request, this field contains the specified prefix." — absent otherwise.

For ListParts specifically, this is a guaranteed panic on every completed call: the pagination loop in any client only terminates once it receives a non-truncated page, and that exact page is the one with NextPartNumberMarker=nil.

For ListMultipartUploads, this fires on the very common case of checking whether an upload already exists for a given key (no delimiter, and the result isn't truncated since there are far fewer than 1,000 matching uploads).

To Reproduce

1. Run versitygw with the s3 (s3proxy) backend pointed at a real S3 bucket:
./versitygw --access=<root> --secret=<root-secret> s3 --access=<s3-key> --secret=<s3-secret> --endpoint=https://s3.amazonaws.com --region=<region>
2. Using any S3 SDK client configured against the gateway, call ListMultipartUploads with a fresh, never-before-used key prefix (no existing multipart upload):
GET /{bucket}?uploads&prefix=<fresh-key>
3. Observe a 500 InternalError response, and [PANIC]: runtime error: invalid memory address or nil pointer dereference logged server-side.
4. Separately: CreateMultipartUpload → UploadPart → ListParts on that upload — ListParts panics the same way once the list is non-truncated (i.e., on essentially any single-page result).

Expected behavior

[...]

PR #2237, opened 2026-07-10
fix: prevent nil pointer panic in s3proxy ListMultipartUploads and ListParts

jamalveli:
## Problem

`S3Proxy.ListMultipartUploads` and `S3Proxy.ListParts` in `backend/s3proxy/s3.go`
unconditionally dereference several optional `*string`/`*int32`/`*bool` fields
from the AWS SDK response (`KeyMarker`, `UploadIdMarker`, `NextKeyMarker`,
`NextUploadIdMarker`, `Delimiter`, `Prefix`, `MaxUploads`/`MaxParts`,
`IsTruncated`, `PartNumberMarker`/`NextPartNumberMarker`, `Initiator`, `Owner`)
with no nil check.

The AWS SDK leaves these nil whenever the corresponding XML element is absent
from the response - which happens whenever the result isn't truncated (the
terminal page of any pagination loop) or an optional request parameter wasn't
set. This is a guaranteed panic on every completed `ListParts` pagination, and
on the common `ListMultipartUploads` case of a non-truncated result (i.e.
checking whether an upload already exists for a key).

This is the same bug class already fixed in sibling `s3proxy` methods
(`GetObjectAttributes`, `GetBucketVersioning`, `GetBucketOwnershipControls`),
just missed here.

## Fix

Replace the unconditional dereferences with nil-safe `aws.To*` accessors, and
default the `PartNumberMarker`/`NextPartNumberMarker` string-to-int parse to
`0` when the field is absent (matching the semantics of "no marker"). Added
two small helpers, `s3ProxyOwner`/`s3ProxyInitiator`, since the SDK's
`*types.Owner`/`*types.Initiator` pointers can themselves be nil, not just
their inner fields.

## Verification

Built and ran locally against a real S3 bucket (both on `v1.4.1` and current
`main` - the affected code is identical between the two):

- Before: `ListMultipartUploads` and `ListParts` both panicked
  (`500 InternalError`) on every call.
- After: `ListMultipartUploads` → `CreateMultipartUpload` → `UploadPart` →
  `ListParts` → `AbortMultipartUpload` all complete successfully with zero
  panics.

`go build`, `go vet`, and `gofmt -l` all clean on the changed file.

https://github.com/versity/versitygw/issues/2235

### DIFF
--- a/backend/s3proxy/s3.go
+++ b/backend/s3proxy/s3.go
@@ -23,6 +23,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/aws/aws-sdk-go-v2/aws"
 	v4 "github.com/aws/aws-sdk-go-v2/aws/signer/v4"
 	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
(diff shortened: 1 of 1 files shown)


versity-github:

jamalveli:
Hi @benmcclelland  
I've prepared this PR to address. Whenever you have a moment, I'd appreciate your feedback or a review! Thanks.

benmcclelland:
> Hi @benmcclelland I've prepared this PR to address. Whenever you have a moment, I'd appreciate your feedback or a review! Thanks.

looks good! thanks for the PR

benmcclelland (approved):

PR #2253, opened 2026-07-16
feat: add helm chart admin and webui ingress support

### LINKED ISSUE #2251: The helm chart is missing an ingress for the WebUI and Admin API (while it provides a HTTPRoute for it)
Dear maintainers,

just for the sake of completeness, even though Ingress is being "deprecated" by GatewayAPI:

The helm chart allows to expose the Admin API and the WebUI using a HTTPRoute, but does not allow the same using the ingress.

Compare the Ingress definition with the one for the HTTP route. The latter has 3 if-conditions, while the former has only one.

https://github.com/versity/versitygw/blob/main/chart/templates/ingress.yaml

vs.
https://github.com/versity/versitygw/blob/main/chart/templates/httproute.yaml

It would be great (and should not be rocket science) to have feature parity until everyone is on GatewayAPI.

Thanks in advance!

Kind Regards,
Johannes

### LINKED ISSUE #2252: The helm chart is missing a comment that `corsAllowOrigin` should have a `http://´ or `https://` prefix
Dear maintainers,

small thing that might be useful for new users of the helm chart.

The chart allows setting `gateway.corsAllowOrigin`. However, the chart's `values.yaml` file does not mention that this should have a `http://´ or `https://` prefix. This leads to the pod hanging in `CrashLoopBackOff´ and is annoying for new users, especially as other parameters accept the bare hostname without this prefix.

https://github.com/versity/versitygw/blob/main/chart/values.yaml#L167

Kind Regards,
Johannes

benmcclelland:
This change is needed to expose the admin API and WebUI through standard Kubernetes Ingress resources instead of only through service access or Gateway API routes. It adds configurable ingress blocks in values.yaml, and renders dedicated ingress objects for admin and webui when enabled.

Also update corsAllowOrigin comment.

Fixes #2251
Fixes #2252

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: versitygw
 description: A Helm chart for deploying the Versity S3 Gateway on Kubernetes
 type: application
-version: 0.3.4
+version: 0.3.5
 sources:
   - https://github.com/versity/versitygw
 icon: https://raw.githubusercontent.com/versity/versitygw/main/webui/web/assets/images/Versity-logo-blue-horizontal.png
--- a/chart/templates/ingress.yaml
+++ b/chart/templates/ingress.yaml
@@ -33,3 +33,75 @@ spec:
           {{- end }}
     {{- end }}
 {{- end }}
+{{- if and .Values.admin.enabled .Values.admin.ingress.enabled }}
+---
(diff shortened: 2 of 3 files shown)


niksis02 (approved):

kastl-ars:
That was quick! Thanks a lot!

PR #2256, opened 2026-07-17
test: bucket resetting code cleanup

lrm25:


tests/drivers/user.sh
```diff
@@ -119,61 +119,83 @@ setup_bucket_file_and_user_v2() {
...
 }
 
+reset_bucket() {
+  if ! check_param_count_ge_le "bucket, region (optional), endpoint (optional)" 1 3 $#; then
+    return 1
+  fi
+  local bucket="$1" region="$2" endpoint="$3"
+  local response
+
+  if [ "$region" == "" ] || [ "$endpoint" == "" ]; then
```
anodelman: Safer to use -z in case of "" for empty values.


```
  if [[ -z "$input_region" || -z "$input_endpoint" ]]; then
    if ! response=$(get_bucket_location_and_endpoint "$target_bucket" 2>&1); then
      log 2 "error getting bucket region and endpoint location context: $response"
      return 1
    fi
    
    read -r discovered_region discovered_endpoint <<< "$(echo "$response" | xargs)"
  fi
```

anodelman (changes requested):

anodelman (approved):

PR #2262, opened 2026-07-22
website: return 400 for plaintext HTTP sent to a TLS listener

### LINKED ISSUE #2261: [Bug] - bucket website - `http` call when `https` is enabled returns empty response (curl code 52)
**Describe the bug**
For versitygw bucket websites, `https` is the default.  However, when sending to `http` in this case, the response is a code 52 from curl or an empty response error in a browser.  Response should be something that better clarifies things to the user, such as a 400 or 403.

**To Reproduce**
1. Start versitygw with website capability, e.g.:  `./versitygw --access=AKIA**** --secret=******** --region=us-east-1 --cert /Users/lukemccrone/devel/versitygw/tester/cert.pem --key /Users/lukemccrone/devel/versitygw/tester/versitygw.pem --website :8071 --website-domain localhost posix --versioning-dir /tmp/versioning_tester /tmp/gw-tester`
2.  More to be added

**Expected behavior**

**Server Version**

**Additional context**
<!-- Describe s3 client and version if applicable.

JSap0914:
## Summary

Fixes #2261.

When versitygw is configured with TLS (for example a bucket-website endpoint started with `--cert`/`--key`), a client that sends **plaintext HTTP to the HTTPS port** currently gets an aborted connection with no response body — `curl` reports exit code 52 (`Empty reply from server`) and browsers show a generic empty-response error. The user is given no hint that they simply used `http://` instead of `https://`.

This mirrors a well-known case that Go's own `net/http` TLS server already handles: it detects the plaintext request and replies with `400 Bad Request: Client sent an HTTP request to an HTTPS server.` versitygw's fiber/fasthttp servers do not, so the misconfiguration is silent.

## Change

`NewMultiAddrTLSListener` (used by the website, S3 API, admin and web UI TLS endpoints) now wraps each raw listener with a tiny detector that peeks at the first bytes of a connection **before** the TLS layer consumes them:

- If they are the start of a plaintext HTTP request line (`GET /`, `POST `, `HEAD `, …), it writes a plain `HTTP/1.0 400 Bad Request` with `Client sent an HTTP request to an HTTPS server.` and closes the connection.
- Otherwise the bytes are replayed untouched to the TLS handshake. A real TLS `ClientHello` always begins with the `0x16` handshake record byte, which never matches an HTTP method prefix, so legitimate TLS clients are unaffected.

The change is confined to the listener wiring — no protocol or handler changes.

## Tests

Added `s3api/utils/tls_http_detect_test.go`:

- `TestMultiAddrTLSListenerRejectsPlaintextHTTP` — reproduces #2261: without the fix a plaintext `GET` yields an EOF/empty reply; with the fix the client reads back an `HTTP/... 400` response.
- `TestMultiAddrTLSListenerAllowsRealTLS` — regression guard: a genuine `tls.Dial` handshake still completes and round-trips data.

```
go test ./s3api/utils/     # ok
go vet ./s3api/utils/      # clean
gofmt -s -l ...            # clean
staticcheck ./s3api/utils/ # clean
```

Signed-off-by: JSup <hanjisang0914@gmail.com>

### DIFF
--- a/s3api/utils/multi_listener.go
+++ b/s3api/utils/multi_listener.go
@@ -15,9 +15,11 @@
 package utils
 
 import (
+	"bufio"
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"os"
 	"path/filepath"
@@ -371,7 +373,7 @@ func NewMultiAddrTLSListener(network, address string, getCertificateFunc func(*t
 				return nil, fmt.Errorf("failed to set permissions on socket %s: %w", address, err)
 			}
 		}
-		return NewMultiListener(tls.NewListener(ln, config)), nil
+		return NewMultiListener(tls.NewListener(newHTTPDetectListener(ln), config)), nil
 	}
 
 	addrs, err := resolveHostnameAddrs(address)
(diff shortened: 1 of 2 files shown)


versity-github:

s3api/utils/multi_listener.go
```diff
@@ -391,9 +393,82 @@ func NewMultiAddrTLSListener(network, address string, getCertificateFunc func(*t
...
+func newHTTPDetectListener(ln net.Listener) net.Listener {
+	return httpDetectListener{Listener: ln}
+}
+
+func (l httpDetectListener) Accept() (net.Conn, error) {
+	conn, err := l.Listener.Accept()
+	if err != nil {
+		return nil, err
+	}
+	return &httpDetectConn{Conn: conn, r: bufio.NewReader(conn)}, nil
```
niksis02, line 449: `bufio.NewReader` allocates a 4096-byte buffer just to inspect five bytes. It remains in the data path for the connection’s lifetime. At 50,000 keep-alive connections, that is roughly 195 MiB of additional buffering.

Consider using `bufio.NewReaderSize(conn, 5)` instead, which will internally allocate only 16 bytes.

s3api/utils/multi_listener.go
```diff
@@ -391,9 +393,82 @@ func NewMultiAddrTLSListener(network, address string, getCertificateFunc func(*t
...
+	checked bool
+	blocked bool
+}
+
+func (c *httpDetectConn) Read(p []byte) (int, error) {
+	if !c.checked {
+		c.checked = true
+		if peek, err := c.r.Peek(5); err == nil && looksLikePlaintextHTTP(peek) {
+			_, _ = io.WriteString(c.Conn, plaintextHTTPResponse)
+			c.blocked = true
```
niksis02, line 467: The `Date` header is required for all `4xx` responses.
Dictating from HTTP RFC:
>An origin server with a clock (as defined in [Section 5.6.7](https://www.rfc-editor.org/rfc/rfc9110.html#http.date)) MUST generate a Date header field in all [2xx (Successful)](https://www.rfc-editor.org/rfc/rfc9110.html#status.2xx), [3xx (Redirection)](https://www.rfc-editor.org/rfc/rfc9110.html#status.3xx), and [4xx (Client Error)](https://www.rfc-editor.org/rfc/rfc9110.html#status.4xx) responses, and MAY generate a Date header field in [1xx (Informational)](https://www.rfc-editor.org/rfc/rfc9110.html#status.1xx) and [5xx (Server Error)](https://www.rfc-editor.org/rfc/rfc9110.html#status.5xx) responses.

[RFC URL](https://www.rfc-editor.org/rfc/rfc9110.html#section-6.6.1)

s3api/utils/multi_listener.go
```diff
@@ -391,9 +393,82 @@ func NewMultiAddrTLSListener(network, address string, getCertificateFunc func(*t
...
+var errPlaintextHTTPOnTLS = errors.New("plaintext HTTP request received on TLS listener")
+
+// plaintextHTTPResponse is the reply sent to a client that speaks plaintext HTTP
+// to a TLS port. It mirrors the behaviour of net/http's TLS server, which turns
+// the same misconfiguration into a clear 400 rather than an aborted connection.
+const plaintextHTTPResponse = "HTTP/1.0 400 Bad Request\r\n" +
+	"Content-Type: text/plain; charset=utf-8\r\n" +
+	"Connection: close\r\n" +
+	"\r\n" +
+	"Client sent an HTTP request to an HTTPS server.\n"
```
niksis02, line 416: HTTP HEAD requests are not allowed to have response body, while the implementation sends the same `plaintextHTTPResponse` for all http requests, which is a HTTP protocolo violation.

[RFC](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.2)

s3api/utils/multi_listener.go
```diff
@@ -391,9 +393,82 @@ func NewMultiAddrTLSListener(network, address string, getCertificateFunc func(*t
...
+// start of a plaintext HTTP request line. A real TLS ClientHello begins with the
+// handshake record type byte (0x16), which never matches these method prefixes.
+func looksLikePlaintextHTTP(b []byte) bool {
+	if len(b) < 5 {
+		return false
+	}
+	switch string(b[:5]) {
+	case "GET /", "HEAD ", "POST ", "PUT /", "DELET", "OPTIO", "PATCH", "TRACE", "CONNE":
+		return true
+	}
```
niksis02, line 428: The http GET request detection doesn't seem to be all correct. HTTP/1.1 servers must accept absolute-form request targets(e.g. `GET http://example/...`), although direct clients normally use `/` origin-form.

[RFC](https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2)

niksis02 (changes requested):
It would be nice to have an integration test covering this case. I guess the best place to put the test is in `tests/integration/server.go`

benmcclelland:
we will address this later, closing due to staleness for now

PR #2267, opened 2026-07-28
test: first website tests

lrm25:


tests/drivers/cloudfront.sh
```diff
@@ -0,0 +1,309 @@
...
+    return 1
+  fi
+  local username="$1"
+  local response cloudtrail_data distribution_ids
+  local -a distribution_id_array=()
+
+  if ! response=$(send_command env AWS_IGNORE_CONFIGURED_ENDPOINT_URLS=true aws cloudtrail lookup-events \
+      --region us-east-1 --lookup-attributes AttributeKey=Username,AttributeValue="${username}" \
+      --query 'Events[].CloudTrailEvent' --output text 2>&1); then
+    log 2 "error getting distribution ids: $cloudtrail_data"
```
anodelman: cloudtrail_data hasn't been set yet to be part of our log output

tests/drivers/cloudfront.sh
```diff
@@ -0,0 +1,309 @@
...
+    return 1
+  fi
+  cloudtrail_data="$response"
+  distribution_ids=$(jq -r 'select(.eventSource == "cloudfront.amazonaws.com")
```
anodelman: Since you are processing line by line you need to handle potential empty line or malformed json:
```
  if [[ -n "$cloudtrail_data" ]]; then
    distribution_ids=$(jq -r -s '
      .[] | select(. != null) | try fromjson catch null
      | select(. != null and .eventSource == "cloudfront.amazonaws.com")
      | select(.eventName == "CreateDistribution" or .eventName == "CreateDistributionWithTags")
      | .responseElements.distribution.id
    ' <<< "$cloudtrail_data" 2>&1)
  fi
```

tests/rest_scripts/command/putBucketWebsiteCommand.go
```diff
@@ -0,0 +1,95 @@
...
+	HostName                    string = "hostName"
+	Protocol                    string = "protocol"
+	ConditionString             string = "condition"
+	HttpErrorCodeReturnedEquals string = "httpErrorCodeReturnedEquals"
+	KeyPrefixEquals             string = "keyPrefixEquals"
+	RedirectString              string = "redirect"
+	HttpRedirectCode            string = "httpRedirectCode"
+	ReplaceKeyPrefixWith        string = "replaceKeyPrefixWith"
+	ReplaceKeyWith              string = "replaceKeyWith"
+	RoutingRuleString           string = "routingRule"
```
anodelman: From the next item in this list, I think that this should be routingRules.

tests/rest_scripts/command/putBucketWebsiteCommand.go
```diff
@@ -0,0 +1,95 @@
...
+	RoutingRulesString          string = "routingRule"
+)
+
+type ErrorDocument struct {
```
anodelman, line 9: Adding json guarantees JSON payloads unmarshal into Go structs regardless of casing conventions - add this to all the xml: strings.

```
type ErrorDocument struct {
	Key string `json:"key,omitempty" xml:"Key,omitempty"`
}
```

tests/rest_scripts/command/putBucketWebsiteCommand.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func NewPutBucketWebsiteCommand(command *S3RequestBuilder, websiteString string) (*S3RequestBuilder, error) {
+	command.Config.Method = "PUT"
+	command.Config.Query = "website"
+	xmlData, err := convertWebsiteJsonToXml(websiteString)
+	if err != nil {
+		return nil, fmt.Errorf("error assembling website configuration: %w", err)
+	}
+	command.Config.Payload = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" + string(xmlData)
```
anodelman: Can simply this:

```
command.Config.Payload = xml.Header + string(xmlData)
```

anodelman (changes requested):

tests/drivers/cloudfront.sh
lrm25: This is used for testing directly against S3 for comparison purposes, not against versitygw.

anodelman: was referring to the logged error message with an unset var.

anodelman (approved):

tests/drivers/cloudfront.sh
lrm25: Oh, OK, got it.

PR #2268, opened 2026-07-29
feat: add best-effort O_DIRECT support for posix put/get-object put-part

benmcclelland:
This change introduces an opt-in O_DIRECT mode for POSIX object data paths while keeping behavior safe and predictable across different filesystems and kernel constraints. We now use direct I/O when available and beneficial, but preserve correctness by falling back to buffered I/O when runtime read behavior indicates alignment or capability mismatches.

The implementation keeps fast paths available for full-object reads and descriptor-to-descriptor copy operations so kernel-level optimizations can still be used where possible. At the same time, it avoids global assumptions from single runtime failures and performs fallback at the stream level so requests can continue successfully without broad feature disablement.

### DIFF
--- a/backend/posix/data_io_linux.go
+++ b/backend/posix/data_io_linux.go
@@ -1,0 +1,192 @@
+// Copyright 2026 Versity Software
+// This file is licensed under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//go:build linux
+
+package posix
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strconv"
+	"sync"
+	"syscall"
+
+	"github.com/versity/versitygw/backend"
+)
+
+// openDataRead opens object data for reading and applies O_DIRECT when
+// requested. O_DIRECT is best-effort and falls back to buffered I/O when
+// unsupported by the filesystem.
+func openDataRead(name string, useODirect bool) (*os.File, error) {
+	if !useODirect {
+		return os.Open(name)
+	}
+
+	f, err := os.OpenFile(name, syscall.O_RDONLY|syscall.O_DIRECT, 0)
+	if err != nil {
+		if isODirectUnsupportedOpenErr(err) {
+			warnODirectUnsupportedOnce("openDataRead", err)
+			return os.Open(name)
+		}
+
+		return nil, err
+	}
+
+	return f, nil
+}
+
+func buildGetObjectBody(f *os.File, path string, startOffset, length, objSize int64, useODirect bool, readBufferSize int) (io.ReadCloser, error) {
+	if startOffset == 0 && length == objSize {
+		if useODirect {
+			return newODirectReadFallbackFile(path, f, true), nil
+		}
+		return f, nil
+	}
+
+	if !useODirect {
+		rdr := io.NewSectionReader(f, startOffset, length)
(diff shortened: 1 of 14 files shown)


backend/posix/data_io_linux.go
```diff
@@ -0,0 +1,189 @@
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return r.f.Close()
+}
+
+func (r *odirectReadFallbackFile) switchToBufferedAtOffset(offset int64) error {
+	bf, openErr := os.Open(r.path)
+	if openErr != nil {
+		return openErr
+	}
```
niksis02: This reopens the object by pathname after an O_DIRECT read failure. A concurrent put or delete between the original open and fallback can make the response use old metadata/length but new object data—or fail despite the original descriptor remaining valid. Reopen through `/proc/self/fd/<fd>` to preserve the original inode.

backend/posix/with_otmpfile.go
```diff
@@ -67,14 +69,34 @@ func (p *Posix) openTmpFile(dir, bucket, obj string, size int64, acct auth.Accou
 	// file descriptor into the namespace.
 	// Not all filesystems support this, so fallback to CreateTemp for when
 	// this is not supported.
-	fd, err := unix.Open(dir, unix.O_RDWR|unix.O_TMPFILE|unix.O_CLOEXEC, defaultFilePerm)
+	openFlags := unix.O_RDWR | unix.O_TMPFILE | unix.O_CLOEXEC
+	useODirect := false
+	if p.enableODirect {
+		openFlags |= unix.O_DIRECT
+		useODirect = true
+	}
```
niksis02: Enabling O_DIRECT here affects every openTmpFile caller, but the runtime fallback is implemented only in tmpfile.Write. Several existing paths—including multipart assembly—write directly to f.File(), bypassing that method and therefore bypassing the fallback. In particular, when disable-copy-file-range is enabled, io.Copy(f.File(), &onlyRead{pf}) performs ordinary writes to the raw O_DIRECT descriptor. O_DIRECT may require the buffer address, file offset, and write length to be aligned; because the final multipart part can have an arbitrary size, its final write can return EINVAL. That error is propagated as a failed CompleteMultipartUpload instead of reopening the staging file in buffered mode. Either restrict O_DIRECT to paths that write through tmpfile.Write, route these copies through the fallback-aware wrapper, or alternatively an equivalent fallback handling could be added to the raw descriptor-copy paths.

niksis02 (changes requested):

backend/posix/with_otmpfile.go
benmcclelland: nice catch, i have disabled odirect mode in the copy paths. the intent for now is just the put/get data paths will support odirect.

backend/posix/otmpfile_common.go
```diff
@@ -15,22 +15,104 @@
...
+			return n, err
+		}
+
+		n, err := io.Copy(tmp.f, r)
+		tmp.size -= n
+		return n, err
+	}
+
+	return io.CopyBuffer(writerOnly{tmp}, r, make([]byte, 32*1024))
+}
```
niksis02: Adding ReadFrom to tmpfile means io.CopyBuffer(f, rdr, buf) now invokes tmpfile.ReadFrom and ignores the supplied IOBufferSize buffer. When O_DIRECT is disabled or unavailable, this delegates directly to `os.File.ReadFrom`, bypassing `tmpfile.Write`. This bypasses the remaining-content check:
```go
	if int64(len(b)) > tmp.size {
		return 0, fmt.Errorf("write exceeds content length %v", tmp.size)
	}
```
So a backend reader that produces more bytes than the declared content length can write an oversized object and make tmp.size negative. It also allocates the configured 1 MiB pooled buffer in PutObject/UploadPart without ever using it. ReadFrom should preserve the size enforcement and intentionally use the configured buffer, or be removed so io.CopyBuffer continues writing through `tmpfile.Write`.

I'm not sure if this case will ever happen in real world, when the actual bytes deliverred will exceed the Content-Length(maybe it's validated at network layer by fiber/fasthttp). But as long as the size check is there, maybe it makes sense to preserve the size check here as well?

niksis02 (approved):

PR #2269, opened 2026-07-30
feat: add optional data-integrity ETag mode for posix/scoutfs

benmcclelland:
Introduce an opt-in mode that derives ETags from checksums instead of relying on MD5 assumptions, aligning with AWS-compatible scenarios where ETag should be treated as an opaque identifier rather than a content-MD5 signal.

This can significantly reduce CPU load on the server to disable MD5 checksums when we are already validating data with other checksums.

This change adds --data-integrity-etag and VGW_DATA_INTEGRITY_ETAG support for both posix and scoutfs backends. In this mode, PUT object ETags, multipart part ETags, and completed multipart object ETags are checksum-derived, while multipart completion preserves idempotency for non-MD5 part ETags by using deterministic claim token behavior and stable return semantics under concurrent completion.

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"hash"
 	"io"
 	"io/fs"
 	"net/http"
@@ -117,6 +118,16 @@ type Posix struct {
 	// ioBufferSize is the buffer size used by buffered copy/read paths.
 	ioBufferSize int
 	ioBufferPool sync.Pool
+
+	// dataIntegrityEtag, when true, replaces the standard MD5-based ETag with
+	// a checksum-derived ETag that embeds the algorithm name and value.
+	// For multipart uploads this applies to both part ETags and the final
+	// completed object ETag:
+	//   single PUT:          "ALGO-<checksum>"
+	//   multipart part:       "CRC64NVME-<part-checksum>"
+	//   multipart CRC64NVME: "CRC64NVME-<whole-file-checksum>"
+	//   multipart composite: "ALGO-<composite-checksum>-<part-count>"
+	dataIntegrityEtag bool
 }
 
 var _ backend.Backend = &Posix{}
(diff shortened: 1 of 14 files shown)


backend/posix/posix.go
```diff
@@ -1807,30 +1826,60 @@ func (p *Posix) CompleteMultipartUploadWithCopy(ctx context.Context, input *s3.C
...
+			return string(etagBytes)
+		}
+		return multipartClaimToken
+	}
 
 	err = os.Rename(uploadIDDir, uploadIDInProgress)
 	if errors.Is(err, fs.ErrNotExist) {
 		// Another call already claimed this slot and is still assembling the object.
 		if _, statErr := os.Stat(uploadIDInProgress); statErr == nil {
+			etag := multipartClaimToken
```
niksis02, line 1994: In data-integrity mode, multipartClaimToken is a SHA-256-derived synchronization token, while the eventual finalEtag is checksum-derived. A concurrent completion hitting this branch returns success with the claim token, so identical completion requests can return different ETags and the losing response will not match `HeadObject`.
e.g.
1. Request A starts CompleteMultipartUpload.
2. The server creates a SHA-256-based multipartClaimToken solely to name the in-progress directory.
3. Request A later computes the real object ETag, such as "CRC64NVME-<whole-object-checksum>".
4. While A is working, concurrent request B sees the in-progress directory and immediately returns success using `multipartClaimToken` as its `ETag`.

I'm not sure how much problematic this ETag mismatch might be in real world scenarios, but at least we diverge from s3 here.

niksis02 (changes requested):
1. UploadPartCopy still unconditionally calculates and returns an MD5 ETag. With `--data-integrity-etag`, copied parts therefore use MD5 while normally uploaded parts use CRC64NVME, and the intended MD5 CPU savings are lost for copy workloads. 

2. The directory-object path bypasses the new ETag behavior and still stores and returns emptyMD5. Since the option is documented as affecting PUT object ETags, a zero-length key ending in / should use the already-computed empty checksum to produce the same ALGO-<checksum> format.

backend/posix/posix.go
```diff
@@ -3474,6 +3667,14 @@ func (p *Posix) UploadPartCopy(ctx context.Context, upi *s3.UploadPartCopyInput)
...
 	}
 
+	if p.dataIntegrityEtag && crc64nvmeRdr == nil {
+		crc64nvmeRdr, err = utils.NewHashReader(tr, "", utils.HashTypeCRC64NVME)
+		if err != nil {
+			return s3response.CopyPartResult{}, fmt.Errorf("initialize etag crc64nvme reader: %w", err)
+		}
+		tr = crc64nvmeRdr
+	}
+
```
niksis02: This should be initialized before io.Copy, so it calculates the crc64nvme correctly, otherwise for non-empty copies it will always result to `"CRC64NVME-AAAAAAAAAAA="`

niksis02 (changes requested):

niksis02 (dismissed):

niksis02 (approved):

PR #2272, opened 2026-07-31
feat: add STS web identity federation, IAM policy Condition support, and access control enforcement

niksis02:
Implements the `AssumeRoleWithWebIdentity` and `GetCallerIdentity` STS actions, letting callers exchange an external OIDC token for temporary credentials scoped to an IAM role. Token handling covers JWT claim parsing, issuer/audience resolution (including `azp` override semantics), JWKS fetching and caching with `singleflight`-deduplicated refresh, and rate-limited forced refresh on unrecognized `kid` values. OIDC provider thumbprint fetching now performs a real TLS handshake verified against the system trust store and the provider hostname (previously `InsecureSkipVerify`), since the observed certificate is persisted as a long-lived trust anchor rather than used once and discarded; all discovery-document and JWKS fetches go through an SSRF-safe HTTP client with bounded redirects and response size.

Adds policy `Condition` block evaluation, supporting `String`, `Numeric`, `Date`, `Bool`, `BinaryEquals`, and `IpAddress` operators along with their `IfExists`/`Not` variants and `ForAllValues`/`ForAnyValues` set qualifiers, plus policy variable substitution (e.g. `${aws:username}`) in supported operators. Adds identity-based inline policy evaluation and a new IAM authorization middleware that authorizes each request against action, resource, and condition context together, applying the session-policy-intersects-role-policy semantics for assumed-role sessions.

Adds a new debug logger `--log-level` flag (`silent`/`debug`/`unsafe`), along with a tree-based XML masker that redacts secrets and tokens at the property level in logged request/response bodies instead of skipping the whole body. The old `--debug/VGW_DEBUG` flag is kept as a deprecated alias for `--log-level=debug`, printing a console warning that points users at `--log-level` for finer-grained control.

Fixes a Vault storage bug where CAS (check-and-set) writes always read the current document version as 0 because `kvVersion` asserted metadata as `float64` while the Vault client actually returns `json.Number`, causing every write past the first to be rejected as a concurrent modification. Also adds a constant-time `SecureCompare` for signature/token comparisons in sigv4 auth.

Adds an integration test suite (`iam_access_control.go`) covering IAM access control across user, role, and session identities.


iamapi/internal/iamutil/webidentity.go
```diff
@@ -0,0 +1,887 @@
...
+				}
+				for _, ip := range ips {
+					if isDisallowedFetchTarget(ip) {
+						return nil, fmt.Errorf("refusing to dial disallowed address %q for host %q", ip, host)
+					}
+				}
+				return dialer.DialContext(ctx, network, net.JoinHostPort(ips[0].String(), port))
+			},
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true, // verified ourselves via VerifyConnection below
```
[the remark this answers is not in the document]
niksis02: This is a false positive. `InsecureSkipVerify: true` is required here to disable Go's default trust-store-only verification so `VerifyConnection` (line 824) can run instead — per Go's `crypto/tls` semantics, if `InsecureSkipVerify` is `false`, normal verification happens before VerifyConnection and aborts the handshake on failure, so a custom pinned-CA acceptance path would never be reachable. This implements AWS's documented OIDC thumbprint-pinning fallback (accept if the cert chains to the pinned CA thumbprint OR to a system-trusted CA — see the doc comment on `verifyOIDCConnection`). The callback fully replicates standard hostname+chain verification (`cs.PeerCertificates[0].Verify(...)`), so nothing that default crypto/tls verification would reject is being let through. This is the standard Go pattern for certificate pinning, not a disabled TLS check.

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/versity/versitygw
 
 go 1.25.0
 
+toolchain go1.26.5
```
benmcclelland, line 5: is there a reason we need to pin the toolchain here? i would prefer not to have to do this if possible

cmd/versitygw/main.go
```diff
@@ -49,7 +50,7 @@ var (
 	adminLogFile                           string
 	healthPath                             string
 	virtualDomain                          string
-	debug                                  bool
+	logLevel                               string
```
benmcclelland, line 53: I'm ok with adding a log level for debugging, but I think we should keep the old debug option to not break old behavior. it would be best if we still had to have `--debug` to enable debugging and then a new optional `--log-level` to change the type of debug output with reasonable defaults so that previous behavior still works.

benmcclelland: alternatively maybe we could leave in a deprecated but functional --debug option to set loglevel to debug?

kfox1111:
Would be awesome to get this tested along with https://github.com/spiffe/aws-spiffe-workload-helper. workloads automatically getting a jwt from spire, then accessing buckets from versitygw's.

go.mod
niksis02: There was a vulnurability found in `golang.org/x/text v0.38.0`, that's why I upgraded it to `v0.39.0` which requires `go1.26.5`

niksis02:
> Would be awesome to get this tested along with https://github.com/spiffe/aws-spiffe-workload-helper. workloads automatically getting a jwt from spire, then accessing buckets from versitygw's.

For E2E success case testing, we use github actions as OIDC provider: https://github.com/versity/versitygw/pull/2278

go.mod
benmcclelland: I dont think this is needed. the go.mod for golang.org/x/text only requires 1.25.0: https://cs.opensource.google/go/x/text/+/master:go.mod;l=3
Lets just remove the toolchain here. we always build with latest compiler versions anyways. I dont want to pin any versions here.

benmcclelland: actually, i think this is ok. I thought this pinned the version, but looks like it still allows for newer. my misunderstanding.

benmcclelland (approved):

PR #2275, opened 2026-08-03
feat: add new vgwrdma service for cuObject compatible server

benmcclelland:


cumiddleware/cuobj.go
```diff
@@ -0,0 +1,193 @@
...
+
+		return ctx.Next()
+	}
+
+	// Combined token scheme (real cuObject client SDK). The descriptor
+	// passed to the RDMA backend is the raw token string; the remote base
+	// address is parsed from the token's first field, and the transfer
+	// size comes from the standard Content-Length header (RDMA replaces
+	// only the HTTP body, not the usual Content-Length semantics).
+	rctx.SetUserValue(localKeyRDMADescr, token)
```
niksis02, line 150: [`It would be nice to have fixed before merging`]

NVIDIA’s server API treats the RDMA descriptor as opaque and expects the request size and remote-buffer start address as separate control-plane values. This path instead derives the address from the descriptor’s first field and the size from HTTP Content-Length, but neither assumption is established as a client-facing wire contract here. Because an RDMA control request does not carry its payload in the HTTP body, its HTTP content length may be zero, causing PUT to fail without a transfer size. Please define and document a header schema that carries the opaque descriptor, remote address, and transfer size independently.
e.g.
`<opaque-cuObject-descriptor>:<remote-pointer>:<request-size>` with the final two fields encoded as hex(mino-cpp).

cumiddleware/cuobj.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		return fiber.NewError(fiber.StatusBadRequest, HeaderRDMAToken+": "+err.Error())
+	}
+	rctx.SetUserValue(localKeyRDMARemoteStart, remoteStart)
+
+	if size := rctx.Request.Header.ContentLength(); size > 0 {
+		rctx.SetUserValue(localKeyRDMASize, int64(size))
+	}
+
+	return ctx.Next()
```
niksis02, line 164: [`It would be nice to have fixed before merging`]

Successful offloads never emit x-amz-rdma-reply or a transferred-byte count. NVIDIA’s workflow requires an RDMA reply tag so the client can distinguish a completed offload from an ordinary HTTP response. Please emit the appropriate reply only after the RDMA operation succeeds and report the actual number of transferred bytes, particularly for ranged GETs. 

[NVIDIA cuObject workflow](https://docs.nvidia.com/gpudirect-storage/cuobject/index.html#rdma-enabled-get-and-put-workflow).

cubackend/cuserver.go
```diff
@@ -0,0 +1,321 @@
...
+
+// CuServer is a versitygw Backend that embeds another backend and overrides
+// PutObject/GetObject for RDMA-accelerated transfers via cuObjServer.
+// All other S3 operations (list, head, delete, multipart, etc.) pass
+// through to the embedded backend unchanged.
+type CuServer struct {
+	backend.Backend
+	rdmaSrv *rdma.Server
+	pool    *bufferpool.Pool
+}
```
niksis02, line 50: UploadPart cannot safely pass through unchanged when an RDMA descriptor is present. With RDMA offload, the object data travels out-of-band rather than in the HTTP request body, but the embedded backend still expects that body to contain the part. Please implement an RDMA-aware UploadPart path or explicitly decline offload so the client can retry using the normal HTTP data path. NVIDIA lists UPLOAD_PART as a supported cuObject operation in its [supported-operation table](https://docs.nvidia.com/gpudirect-storage/cuobject/index.html#supported-s3-operations).

rdma/rdma.go
```diff
@@ -0,0 +1,295 @@
...
+// StartSession initiates the RDMA listening session.
+// StartSession must not be called concurrently with CloseSession or Close.
+func (s *Server) StartSession() error {
+	s.mu.Lock()
+	alreadyOpen := s.sessionOpen
+	s.mu.Unlock()
+	if alreadyOpen {
+		return nil
+	}
+	rc := C.cuobj_server_start_session(s.csrv)
```
niksis02, line 134: [`It would be nice to have fixed before merging`]

NVIDIA documents that the endpoint constructors already start the RDMA session and that construction can still return an object when session startup failed; callers must check isConnected(). Here sessionOpen starts false, so the method may start an already-running session again. Conversely, when dlsym cannot find the method, the C++ wrapper returns success without checking connectivity. Please use IsConnected() as the constructor readiness check and avoid the duplicate start.

[cuObjServer constructor contract](https://docs.nvidia.com/gpudirect-storage/cuobject/cuObjServer-api/index.html).

Makefile
```diff
@@ -38,6 +71,90 @@ build: $(BIN)
...
+		cuwrapper/rdma_host_client_wrapper.cpp
+	$(AR) rcs $(HOSTCLIENT_WRAPPER_LIB) cuwrapper/rdma_host_client_wrapper.o
+	rm -f cuwrapper/rdma_host_client_wrapper.o
+
+.PHONY: vgwrdma
+vgwrdma: $(VGWRDMA_WRAPPER_LIB)
+	CGO_ENABLED=1 \
+	CGO_CFLAGS="$(VGWRDMA_RDMA_CGO_CFLAGS)" \
+	CGO_LDFLAGS="$(VGWRDMA_RDMA_CGO_LDFLAGS)" \
+		$(GOBUILD) $(LDFLAGS) -o $(VGWRDMA_BIN) $(VGWRDMA_CMD)
```
niksis02: [`It would be nice to have fixed before merging`]

Could we add a clean Linux compile/link check for vgwrdma, cuobjtest-gpu, and cuobjtest-host before merging? The current green Go checks do not final-link these no-test native packages, while the wrapper archives are generated, gitignored, and dependent on libraries available only in the builder images. At present CI does not prove that any newly shipped native binary builds.

cuobjclient/transport.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return err
+	}
+	defer out.Body.Close()
+	_, err = io.Copy(io.Discard, out.Body)
+	return err
+}
+
+func withRDMAHeaders(base *s3lib.Client, descr string, size int64, remoteStart uint64) *s3lib.Client {
+	opts := base.Options()
```
niksis02, line 129: [`Not so urgent, could be fixed later`]

A client created through config.LoadDefaultConfig normally carries WhenSupported request and response checksum policies. The SDK will checksum the empty HTTP PUT body, which the gateway then compares against the actual RDMA payload, and GET validation can compare the object checksum against the empty HTTP response body. Configure these control requests appropriately and handle any desired data checksum out-of-band. AWS documents the default behavior in its

cuobjclient/transport.go
```diff
@@ -0,0 +1,87 @@
...
+type rdmaTransport struct {
+	base        http.RoundTripper
+	descr       string
+	size        int64
+	remoteStart uint64
+}
+
+func (t *rdmaTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	r := req.Clone(req.Context())
+	r.Header.Set(cumiddleware.HeaderRDMADescr, t.descr)
```
niksis02: [`Not so urgent, could be fixed later`]

These legacy RDMA descriptor, address, and size headers are inserted in RoundTrip, after SigV4 signing, so they are absent from SignedHeaders. On the test tool’s default plain-HTTP endpoint, an intermediary can modify the RDMA controls without invalidating authentication. Add them through SDK middleware before signing.

cuobjclient/session_linux.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	if err := s.memset(0); err != nil {
+		return err
+	}
+	token, err := s.getToken(cuObjOpGet)
+	if err != nil {
+		return err
+	}
+	defer s.putToken(token)
+	if err := doGet(base, bucket, key, int64(s.size), C.GoString(token), s.remoteStart); err != nil {
```
niksis02, line 137: [`Not so urgent, could be fixed later`]

doGet returns only an error, so the session cannot distinguish a full transfer from an object smaller than the fixed session size. A shorter object succeeds and the entire zero-padded GPU buffer is copied to dst. Return and validate the transferred-byte count; the host implementation has the same issue.

cuobjclient/transport.go
```diff
@@ -0,0 +1,87 @@
...
+		return err
+	}
+	defer out.Body.Close()
+	_, err = io.Copy(io.Discard, out.Body)
+	return err
+}
+
+func withRDMAHeaders(base *s3lib.Client, descr string, size int64, remoteStart uint64) *s3lib.Client {
+	opts := base.Options()
+	opts.HTTPClient = &http.Client{
```
niksis02: [`Not so urgent, could be fixed later`]

Replacing the caller’s HTTP client with http.DefaultTransport discards custom TLS roots, proxies, dial behavior, timeouts, and test transports. An endpoint that works through the base S3 client can therefore fail only on RDMA operations. Please wrap or preserve the configured HTTP client.

cubackend/cuserver.go
```diff
@@ -0,0 +1,321 @@
...
+			chunk := rdmaPutChunkSize(len(local), remaining)
+			if chunk <= 0 {
+				wrapped := fmt.Errorf("cuserver: invalid RDMA PUT buffer size %d", len(local))
+				_ = pw.CloseWithError(wrapped)
+				producerErrCh <- wrapped
+				return
+			}
+
+			// RDMA READ current chunk from client into the start of local buffer.
+			if _, e := cs.rdmaSrv.HandlePut(key, buf, remoteStart+uint64(offset), chunk, descr, channelID); e != nil {
```
niksis02: [`Not so urgent, could be fixed later`]

Please require HandlePut to return exactly chunk bytes instead of discarding its count. A short positive result causes the full slice, including stale pooled bytes, to be written. The GET call at line 282 should likewise validate and accumulate its result.

rdma/bufferpool/pool.go
```diff
@@ -0,0 +1,97 @@
...
+	for i := range count {
+		ptr, err := server.AllocHostBuffer(bufSize)
+		if err != nil {
+			p.Close()
+			return nil, fmt.Errorf("bufferpool: alloc buffer %d: %w", i, err)
+		}
+
+		buf, err := server.RegisterBuffer(ptr, bufSize)
+		if err != nil {
+			p.Close()
```
niksis02, line 61: [`Not so urgent, could be fixed later`]

If registration fails, the newly allocated ptr has not yet been added to p.all, so p.Close() frees only earlier buffers and leaks this allocation. Free the current pointer before closing the partial pool.

cuwrapper/rdma_host_client_wrapper.cpp
```diff
@@ -0,0 +1,415 @@
...
+        if (ibv_query_gid(c->ctx, c->port_num, i, &gid)) {
+            continue;
+        }
+        if (gid_is_zero(gid)) {
+            continue;
+        }
+        if (gid_is_link_local(gid)) {
+            continue;
+        }
+        if (gid_is_ipv4_mapped(gid)) {
```
niksis02, line 136: [`Not so urgent, could be fixed later`]

Auto-selection examines GID bytes but not GID type. mlx5 commonly exposes identical values in adjacent RoCEv1 and RoCEv2 entries, with v1 first, so this can select a protocol cuObject does not support. Query the GID type and prefer RoCEv2. NVIDIA documents the paired layout in its [GID-table documentation](https://networking-docs.nvidia.com/mlnxofedswum/24.10-5.1.6.1lts/rdma-over-converged-ethernet-roce).

cmd/vgwrdma/main.go
```diff
@@ -0,0 +1,1039 @@
...
+			Usage:       "maximum allowed source object size in bytes for CopyObject; objects larger than this are rejected",
+			EnvVars:     []string{"VGW_COPY_OBJECT_THRESHOLD"},
+			Value:       5 * 1024 * 1024 * 1024,
+			Destination: &gwcli.CopyObjectThreshold,
+		},
+		&cli.StringFlag{
+			Name:        "rdma-ip",
+			Usage:       "IP address for RDMA interface (required to enable cuObject RDMA backend)",
+			EnvVars:     []string{"VGW_RDMA_IP"},
+			Required:    true,
```
niksis02: [`Not so urgent, could be fixed later`]

Please scope the RDMA-IP requirement to commands that start a gateway. Currently both vgwrdma --version and vgwrdma admin --help print their output and then exit 1 with Required flag "rdma-ip" not set; admin, utils, help, and version operations do not require a server bind address.

cmd/vgwrdma/main.go
```diff
@@ -0,0 +1,1039 @@
...
+	var s3Opts []s3api.Option
+	if rdmaIP != "" {
+		rdma.ConfigureTelemetry(debug)
+
+		var tunables *rdma.RDMATunables
+		if rdmaTunablesSet {
+			t := rdma.DefaultRDMATunables()
+			t.DCKey = rdmaDCKey
+			t.NumDCIs = rdmaNumDCIs
+			t.CQDepth = uint32(rdmaCQDepth)
```
niksis02, line 943: [`Not so urgent, could be fixed later`]

Please validate the remaining numeric RDMA flags before narrowing or passing them. Negative rdma-num-dcis, rdma-cq-depth values above MaxUint32, and non-positive pool values are currently accepted, wrapped, or silently replaced with defaults.

cmd/cuobjtest/main.go
```diff
@@ -0,0 +1,490 @@
...
+	"github.com/versity/versitygw/cuobjclient"
+)
+
+var (
+	endpoint   = flag.String("endpoint", "http://localhost:7070", "cuserver S3 endpoint URL")
+	bucket     = flag.String("bucket", "cuobjtest", "S3 bucket (created if absent)")
+	key        = flag.String("key", "cuobjtest-object", "S3 object key prefix")
+	sizeStr    = flag.String("size", "64MiB", "Transfer size per iteration: e.g. 4MiB, 256MiB, 1GiB")
+	access     = flag.String("access", "admin", "S3 access key")
+	secret     = flag.String("secret", "", "S3 secret key")
```
niksis02: [`Not so urgent, could be fixed later`]

Please support an environment or AWS credential provider instead of requiring the secret in argv. -secret exposes the credential through shell history and process command lines for the duration of the benchmark.

cmd/cuobjtest/main.go
```diff
@@ -0,0 +1,490 @@
...
+		UsePathStyle: true,
+	})
+}
+
+func ensureBucket(c *s3lib.Client, bucket string) error {
+	_, err := c.HeadBucket(context.Background(), &s3lib.HeadBucketInput{Bucket: aws.String(bucket)})
+	if err == nil {
+		return nil
+	}
+	_, err = c.CreateBucket(context.Background(), &s3lib.CreateBucketInput{
```
niksis02, line 400: [`Not so urgent, could be fixed later`]

Only create the bucket after establishing that it is absent. A 403, bad credentials, timeout, or transient server error currently falls through to CreateBucket, masking the original error and requiring unnecessary bucket-management permission.

cmd/cuobjtest/main.go
```diff
@@ -0,0 +1,490 @@
...
+func s3GetObject(base *s3lib.Client, bucket, key string, dst []byte) error {
+	out, err := base.GetObject(context.Background(), &s3lib.GetObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(key),
+	})
+	if err != nil {
+		return err
+	}
+	defer out.Body.Close()
+	_, err = io.ReadFull(out.Body, dst)
```
niksis02, line 378: [`Not so urgent, could be fixed later`]

io.ReadFull returns success after filling dst even if the response contains additional bytes. In -std-s3 -get-only mode, a too-small -size silently benchmarks only a prefix. Validate ContentLength or read one additional byte and require EOF.

cmd/cuobjtest/main.go
```diff
@@ -0,0 +1,490 @@
...
+	putCount := len(stats.putDurs)
+	getCount := len(stats.getDurs)
+	putBytes := int64(putCount * size)
+	getBytes := int64(getCount * size)
+	totalBytes := putBytes + getBytes
+
+	fmt.Printf("OVERALL elapsed=%s", fmtDur(benchmarkDur))
+	if putCount > 0 {
+		fmt.Printf("  PUT agg=%s (%d ops, %.2f ops/s)",
+			fmt.Sprintf("%.3f GB/s", float64(putBytes)/benchmarkDur.Seconds()/1e9),
```
niksis02: [`Not so urgent, could be fixed later`]

PUT and GET “aggregate” rates both divide by the duration of the entire combined benchmark, including the opposite operation and buffer preparation. When both operations run, neither value is the corresponding operation’s throughput. Use the sum of the relevant operation durations or rename these as amortized wall-clock rates.

Makefile
```diff
@@ -38,6 +71,90 @@ build: $(BIN)
...
+.PHONY: cuobjtest-host
+cuobjtest-host: $(HOSTCLIENT_WRAPPER_LIB)
+	CGO_ENABLED=1 \
+		$(GOBUILD) -tags $(CUOBJTEST_HOST_TAG) $(LDFLAGS) -o $(CUOBJTEST_BIN) $(CUOBJTEST_CMD)
+
+.PHONY: vgwrdma-builder-image
+vgwrdma-builder-image:
+	docker build --platform $(VGWRDMA_BUILDER_PLATFORM) -f $(VGWRDMA_BUILDER_DOCKERFILE) -t $(VGWRDMA_BUILDER_IMAGE) .
+
+.PHONY: vgwrdma-docker
```
niksis02, line 122: [`Not so urgent, could be fixed later`]

Could we add an operator and build guide? The PR does not document the Linux/amd64+cgo restriction, NVIDIA/RDMA runtime requirements, matching DC key and NIC configuration, builder-only nature of the Dockerfiles, or the default allocation of four registered 1 GiB buffers.

cumiddleware/cuobj.go
```diff
@@ -0,0 +1,193 @@
...
+// the fasthttp request context so the backend can retrieve them via the
+// GetRDMA* helper functions.
+//
+// If neither the legacy descriptor header nor the combined RDMA token
+// header is present, the request passes through unchanged — the backend
+// will use the normal (non-RDMA) code path.
+//
+// If a descriptor is present but required fields are malformed, a 400 Bad
+// Request is returned immediately.
+func CuObjMiddleware(ctx fiber.Ctx) error {
```
niksis02, line 105: [`Not so urgent, could be fixed later`]

It would be nice to have protocol-level tests that do not require RDMA hardware. The fixtures should treat the NVIDIA descriptor as opaque and provide the remote address and transfer size through the documented control-plane representation. Tests should also cover zero-body RDMA control requests, reply headers, UploadPart, ranged GETs, and short native transfers.

Makefile
benmcclelland: yeah, i'll be updating a wiki page with all of this info

cumiddleware/cuobj.go
benmcclelland: I think the minio-cpp implementation is from an older nvidia protocol definition. It seems to no longer be compatible with the published protocol. this protocol definition comes from the cuObjectServer release docs https://github.com/versity/versitygw/blob/5d1b0db14c1e1f87591c8bb38c377a9453c8db44/cumiddleware/cuobj.go#L58-L77

cuobjclient/transport.go
```diff
@@ -0,0 +1,139 @@
...
+	smithymiddleware "github.com/aws/smithy-go/middleware"
+	smithyhttp "github.com/aws/smithy-go/transport/http"
+
+	s3lib "github.com/aws/aws-sdk-go-v2/service/s3"
+
+	"github.com/versity/versitygw/cumiddleware"
+)
+
+func doPut(base *s3lib.Client, bucket, key string, size int64, descr string, remoteStart uint64) error {
+	c := withRDMAHeaders(base, descr, size, remoteStart, nil, nil)
```
niksis02: doPut passes nil, nil for the response-header captures and returns success based only on the http status. Because this control request deliberately has an empty body and Content-Length: 0, an endpoint where RDMA offload is disabled or ignored can accept the request, store an empty object, and return 200; Upload then reports success. This is especially invisible in PUT-only mode. Please capture and validate `X-Amz-Rdma-Reply` and require `X-Amz-Rdma-Bytes-Transferred == size`, as `doGet` already does.

rdma/rdma.go
niksis02: Constructing cuObjServer starts a usable server, and  v1.2.0 starts the session internally. However, NewServer returns with sessionOpen == false, so StartSession invokes the dynamically resolved startRDMASession again whenever that symbol is visible. The IsConnected fallback only handles a nonzero result; it does not prevent the second call. Please check constructor state before invoking the method—or rely exclusively on the supported constructor lifecycle—and make the close path follow the same ownership model.

niksis02: This remains critical because it affects whether the gateway can reliably start and shut down with the supplied libcuobjserver-1.2.0.68 runtime.

cubackend/cuserver.go
```diff
@@ -0,0 +1,345 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Determine the actual content size
+	var size int64
+	if res.ContentLength != nil {
+		size = *res.ContentLength
+	}
+	if size <= 0 {
```
niksis02, line 254: The GET path uses the object’s ContentLength as the transfer size but never compares it with GetRDMASize(ctx) or the token’s maximum-region-size field. If the object is larger than the registered client buffer, the gateway begins RDMA writes using the larger object size and relies on the native library to reject an out-of-range operation, potentially after a partial transfer. Please validate the requested capacity against the object or range size before the first HandleGet call.

niksis02 (changes requested):

niksis02 (dismissed):

PR #2280, opened 2026-08-07
fix: buckets with versioning enabled don't have "Show versions" button in UI

buttilda:
The `Show versions` button only appears right after enabling versioning. Refreshing the UI causes it to disappear.
Screenshot 2026-08-07 at 16 42 36

After refreshing, the `Versioning` modal also always asks if you want to enable versioning even for buckets where it is already enabled.


versity-github:

benmcclelland (approved):

PR #2288, opened 2026-08-10
fix(posix): make conditional PUT evaluation and publication atomic per key

cmer:
## Problem

The POSIX backend evaluates `If-Match` / `If-None-Match: *` correctly for sequential PUTs, but precondition evaluation and object publication are not one atomic operation. Under concurrent conditional PUTs to the same bucket/key, multiple writers read the same old ETag, all pass the check, and all publish successfully. This breaks applications that use S3 conditional writes as a compare-and-swap primitive.

## Reproduction

Observed by Buzz's git object-store A3 conformance probe (https://github.com/block/buzz) against the POSIX backend on NFSv4.2: 32 concurrent PUTs with the same valid `If-Match` ETag (or `If-None-Match: *` on an absent key) should yield exactly 1×2xx + 31×412, but the gateway allowed anywhere from 2 to all 32 writers to succeed. The new barrier-synchronized regression tests reproduce this deterministically (`got 32 / 22 / 10 winners` on unpatched code), as does the included external SigV4 test against a live gateway — including with two gateway processes sharing one backend directory (`5×2xx + 27×412` before the fix).

## Root cause

`PutObjectWithPostFunc` and `CompleteMultipartUploadWithCopy` in `backend/posix/posix.go` use a check-then-act pattern: the ETag is read and `backend.EvaluateObjectPutPreconditions` evaluated before the request body is staged, while publication happens much later via `tmpfile.link()` (linkat/rename). Nothing serializes the check against publication, so the race window spans the entire body upload (or multipart assembly).

## Fix

Introduce a per-object publish lock (`lockObjectPublish`, `backend/posix/objlock.go`) that every object-publication path holds across authoritative precondition re-evaluation, metadata stores, and the final link:

- Exclusion is an advisory file lock (`flock` on unix, `LockFileEx` on windows) on `bucket/.sgwtmp/objlock/<shard>`, where shard is the first byte of `sha256(key)` (256 shards per bucket). The gateway is stateless and multiple instances may share one filesystem, so a process-local mutex alone is not sufficient. A process-local striped mutex is taken alongside so in-process contention never thrashes the filesystem lock. Lock files are empty, bounded, created on demand, and never unlinked (avoiding the unlink/recreate flock race, which cannot be detected reliably on NFS due to attribute caching).
- Request bodies are staged to the temp file before the lock is taken; the lock covers only the short commit phase. The kernel releases advisory locks on close or process death, so failure, cancellation, panic, or crash cannot leave a stale lock.
- The pre-staging precondition check is kept as an advisory fast-fail; the authoritative check repeats under the lock. Participating paths: ordinary PUT (conditional and unconditional), directory-object PUT, CopyObject (via PutObject), scoutfs (via `PutObjectWithPostFunc` / `CompleteMultipartUploadWithCopy`), and multipart completion.
- Conditional writes on versioned buckets defer the version snapshot until the precondition is confirmed under the lock, so losing writers no longer create spurious version snapshots.
- The `postprocess` hook now runs before any path-based metadata is written, so a failing hook no longer leaves a stale sidecar ETag for a never-published object.

## Concurrency guarantees

- For one bucket/key, state read, condition evaluation, and publication are atomic: exactly one winner per conditional race, within a process, across processes sharing a filesystem, and across NFSv4 clients with server-backed locking.
- Unrelated keys remain concurrently writable (shard collisions only briefly serialize the commit phase).
- If the filesystem does not support advisory locking (e.g. NFS mounted with `-o nolock`), the gateway degrades to process-local exclusion and logs a warning once instead of failing PUTs.

## Tests performed

- New regression tests (`backend/posix/posix_conditional_put_test.go`), written RED-first and proven failing on unpatched code: 32-way `If-None-Match: *` and `If-Match` races for ordinary PUT and multipart completion, sequential semantics (stale `If-Match` → 412, `If-Match` on missing key → NoSuchKey preserved), unconditional-vs-conditional race, 32 distinct keys concurrently writable, and cleanup after 412/injected publication failure — all for both xattr and sidecar metadata modes, barrier-synchronized, no sleeps.
- `go test -race -count=3` on the new tests; `go test -race -count=2 ./backend/posix/`; full `go test ./...` — all pass.
- `gofmt -s -l` clean; `staticcheck` (CI version) clean; `go build ./cmd/versitygw` OK.
- External SigV4 verification (`tests/conditionalrace/`, env-guarded, skipped unless `VERSITY_ENDPOINT`/`VERSITY_ACCESS_KEY`/`VERSITY_SECRET_KEY`/`VERSITY_BUCKET` are set): replicates the Buzz A3 probe against a live gateway — sequential round-trip with ETag consistency, both 32-way races for three rounds, plus basic ops including a 20 MiB multipart upload with SHA-256 comparison. Passes against one gateway and against two gateway processes sharing one root (`VERSITY_ENDPOINT2`), which exercises the cross-process file locking.

## Compatibility / portability

- unix (`flock`) and windows (`LockFileEx`) implementations; cross-compiles verified for linux, windows, and freebsd.
- NFS: Linux clients map flock to NFSv4 byte-range locks (NLM on NFSv3), giving cross-client exclusion; `nolock` / `local_lock=*` mounts reduce exclusion to a single client (documented in `objlock.go`, warned at runtime).
- S3 response behavior unchanged: failed `If-Match` and `If-None-Match: *` → 412; `If-Match` on a missing object retains the existing NoSuchKey behavior; successful PUT returns the committed object's ETag.
- Known remaining gap (pre-existing, out of scope): conditional DeleteObject retains its own check-then-act window.

### DIFF
--- a/backend/posix/objlock.go
+++ b/backend/posix/objlock.go
@@ -1,0 +1,148 @@
+// Copyright 2026 Versity Software
+// This file is licensed under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package posix
+
+import (
+	"context"
+	"crypto/sha256"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/versity/versitygw/backend"
+	"github.com/versity/versitygw/debuglogger"
+)
+
+// Object publish locking
+//
+// S3 conditional writes (If-Match / If-None-Match: *) require that reading the
+// current object state, evaluating the condition, and publishing the
+// replacement happen as one atomic step per bucket/key. The gateway is
+// stateless and multiple gateway processes may share the same backend
+// filesystem, so an in-process mutex alone is not sufficient: exclusion is
+// provided by an advisory lock (flock on unix, LockFileEx on windows) on a
+// shared lock file, combined with a process-local striped mutex so that
+// contention within one process is resolved cheaply and each process presents
+// at most one waiter to the filesystem lock.
+//
+// Lock identity: bucket/.sgwtmp/objlock/<shard>, where shard is the first
+// byte of sha256(object key) rendered as two hex characters. Hashing the key
+// gives a stable, traversal-safe, fixed-length name; sharding (256 slots per
+// bucket) keeps the number of lock files bounded while still letting writes
+// to unrelated keys proceed concurrently in the common case. Lock files are
+// empty, created on demand, and never unlinked: unlinking an flock file opens
+// a classic race where a waiter holds a lock on an unlinked inode while a new
+// file takes its place, which is unsafe to detect reliably on NFS due to
+// attribute caching.
+//
+// The lock is held only for the commit phase (condition re-check, metadata
+// stores, final link/rename) — request bodies are staged to a temp file
+// before the lock is taken. The OS releases advisory locks automatically when
+// the file handle is closed or the process exits, so failures, cancellation,
+// or crashes cannot leave a permanently stale lock.
+//
+// NFS notes: flock on Linux NFS clients is mapped to NFSv4 byte-range locks
+// (or NLM on NFSv3), giving cross-client exclusion. Mounting with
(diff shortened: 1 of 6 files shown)


versity-github:

benmcclelland:
This overall looks ok. I'm a little nervous allowing to fallback to in process locking if the filesystem doesnt support flock. Since this will not be able to enforce behavior if multiple gateways are deployed. But I guess the warning log might be good enough to convey that its best effort if the filesystem doesnt offer the needed functionality, and the fallback behavior isnt any worse that what it is now apparently. Going to re-review this again, but just wanted to comment on my current considerations for why this hasn't been immediately merged.

julian-corbet:
I need exactly this fix / behaviour, too. Can you guys please pull it? =)

cmer:
@benmcclelland like you pointed out, i don't think there's any downsides in merging this, vs the way it's currently working.

benmcclelland (approved):

PR #2289, opened 2026-08-11
feat: add configurable file permissions for new objects

streamvsriver:
Replace the hardcoded 0644 defaultFilePerm with a NewFilePerm option on the posix and scoutfs backends, exposed as the --file-perms flag and VGW_FILE_PERMS env var alongside the existing dir-perms option.

The mode passed to open() is masked by the process umask, so the O_TMPFILE path now chmods explicitly to match the CreateTemp fallback path and give new objects the configured mode regardless of umask.

### DIFF
--- a/cmd/internal/gwcli/posix.go
+++ b/cmd/internal/gwcli/posix.go
@@ -24,11 +24,17 @@ import (
 	"github.com/versity/versitygw/backend/posix"
 )
 
+// maxFilePerms is the highest accepted value for the file-perms option. The
+// posix backend only applies permission bits to new objects, so anything
+// above this would be silently dropped.
+const maxFilePerms = 0777
+
 var (
 	chownuid, chowngid   bool
 	bucketlinks          bool
 	versioningDir        string
 	dirPerms             uint
+	filePerms            uint
 	sidecar              string
 	nometa               bool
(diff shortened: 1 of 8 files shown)


versity-github:

cmd/internal/gwcli/posix.go
```diff
@@ -161,6 +175,10 @@ func runPosix(ctx *cli.Context) error {
 		return fmt.Errorf("invalid directory permissions: %d", dirPerms)
 	}
 
+	if filePerms > maxFilePerms {
+		return fmt.Errorf("invalid file permissions: %o, must be within 0000-0777", filePerms)
```
benmcclelland, line 179: This is reporting that 0000 is a valid setting, but the code appears to use the default 0644 if 0000 is set. I'm not sure if we should continue to disallow 0000 and just update the docs here for clarity, or actually allow setting 0000. Also not sure if there are any concerns here about allowing settings that will prevent the gateway from being able to read files after upload.

benmcclelland:
fixed up the commit and merged it in https://github.com/versity/versitygw/pull/2319

streamvsriver:
Thanks a lot @benmcclelland, sorry for the delay on response, was off for two weeks.

PR #2293, opened 2026-08-13
fix(posix): return empty config for GetBucketVersioning when unconfigured

lhjnano:
Validate bucket existence before checking the versioning state so a missing bucket still yields 404 NoSuchBucket. When no versioning directory is configured, respond with 200 and an empty configuration instead of VersioningNotConfigured, matching AWS semantics. GUI clients (e.g. S3 Browser) poll this API while browsing buckets and were failing on default deployments. PutBucketVersioning still rejects with VersioningNotConfigured.

Also fixes a swapped test body between the Get/Put "not configured" cases, and adds coverage for GetBucketVersioning on a non-existent bucket.

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -890,10 +890,6 @@ func (p *Posix) GetBucketVersioning(ctx context.Context, bucket string) (s3respo
 	}
 	defer release()
 
-	if !p.versioningEnabled() {
-		return s3response.GetBucketVersioningOutput{}, s3err.GetAPIError(s3err.ErrVersioningNotConfigured)
-	}
-
 	if !p.isBucketValid(bucket) {
 		return s3response.GetBucketVersioningOutput{}, s3err.GetBucketErr(s3err.ErrInvalidBucketName, bucket)
 	}
@@ -904,6 +900,15 @@ func (p *Posix) GetBucketVersioning(ctx context.Context, bucket string) (s3respo
 	}
 	if err != nil {
 		return s3response.GetBucketVersioningOutput{}, fmt.Errorf("stat bucket: %w", err)
+	}
+
+	if !p.versioningEnabled() {
+		// No versioning directory is configured: report the bucket as
+		// never-versioned (200 with an empty configuration) to match AWS,
+		// instead of failing with VersioningNotConfigured. GUI clients
+		// (S3 Browser, CloudBerry, ...) poll this API while browsing
+		// buckets. PutBucketVersioning still rejects the request.
+		return s3response.GetBucketVersioningOutput{}, nil
 	}
 
 	vData, err := p.meta.RetrieveAttribute(nil, bucket, "", versioningKey)
--- a/backend/posix/versioning_test.go
+++ b/backend/posix/versioning_test.go
@@ -1,0 +1,102 @@
+// Copyright 2026 Versity Software
+// This file is licensed under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package posix
+
+import (
+	"context"
+	"errors"
+	"os"
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/versity/versitygw/backend/meta"
+	"github.com/versity/versitygw/s3err"
+)
+
+// newUnversionedGateway creates a Posix backend over a temporary root
+// directory with no versioning directory configured, i.e. gateway-level
+// versioning disabled, matching the default gateway configuration.
+func newUnversionedGateway(t *testing.T) *Posix {
+	t.Helper()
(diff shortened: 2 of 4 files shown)


versity-github:

backend/posix/posix.go
```diff
@@ -906,6 +902,15 @@ func (p *Posix) GetBucketVersioning(ctx context.Context, bucket string) (s3respo
...
 
+	if !p.versioningEnabled() {
+		// No versioning directory is configured: report the bucket as
+		// never-versioned (200 with an empty configuration) to match AWS,
+		// instead of failing with VersioningNotConfigured. GUI clients
+		// (S3 Browser, CloudBerry, ...) poll this API while browsing
+		// buckets. PutBucketVersioning still rejects the request.
+		return s3response.GetBucketVersioningOutput{}, nil
+	}
+
```
benmcclelland, line 913: This early return is missing the bucket validation checks. This needs to come after the validation check for invalid bucket names and bucket existing.

benmcclelland (changes requested):

backend/posix/posix.go
lhjnano: Thanks for the review!

That reordering is actually what this PR does — the early return already sits after both checks. The diff splits the move across two hunks (removed above, added below), which makes it hard to see. Here's the resulting function:

```go
if !p.isBucketValid(bucket) {
    return s3response.GetBucketVersioningOutput{}, s3err.GetBucketErr(s3err.ErrInvalidBucketName, bucket)
}

_, err = os.Stat(bucket)
    return s3response.GetBucketVersioningOutput{}, s3err.GetBucketErr(s3err.ErrNoSuchBucket, bucket)
}
if err != nil {
    return s3response.GetBucketVersioningOutput{}, fmt.Errorf("stat bucket: %w", err)
}

if !p.versioningEnabled() {
    return s3response.GetBucketVersioningOutput{}, nil
}
```

On main, `versioningEnabled()` runs first, so a request for a non-existent bucket returns 501 VersioningNotConfigured instead of 404 NoSuchBucket. Moving it below the validation fixes that as well, and the new `VersioningDisabled_GetBucketVersioning_no_such_bucket` test covers that case.

Happy to restructure if you had something else in mind.

benmcclelland: I think the intended behavior here is to preserve the normal bucket validation flow before checking gateway-level versioning configuration.

Specifically:
* validate the bucket name first
* verify the bucket exists
* only then check whether versioning is enabled at the gateway level
* if the gateway is unconfigured but the bucket is valid and exists, GetBucketVersioning should return an empty configuration response with HTTP 200 (Status == nil)
* PutBucketVersioning should continue to return VersioningNotConfigured when gateway-level versioning is disabled

The implementation should be adjusted to match that ordering:
move the versioningEnabled() check below the bucket validation and existence checks
ensure GetBucketVersioning returns an empty config when unconfigured rather than ErrVersioningNotConfigured
keep PutBucketVersioning rejecting with VersioningNotConfigured

Otherwise we will be returning incorrect errors like VersioningNotConfigured for
bucket = bad bucket name
gateway versioning = disabled
instead of InvalidBucketName

lhjnano: Let me map your spec onto the functions involved, so we can confirm we're describing the same behavior.

| your requirement | function in `GetBucketVersioning` | result |
|---|---|---|
| validate the bucket name first | `p.isBucketValid(bucket)` → `backend.IsValidDirectoryName` | `GetBucketErr(ErrInvalidBucketName)` → 400 |
| verify the bucket exists | `os.Stat(bucket)` + `errors.Is(err, fs.ErrNotExist)` | `GetBucketErr(ErrNoSuchBucket)` → 404 |
| only then check gateway-level versioning | `p.versioningEnabled()`, i.e. `p.versioningDir != ""` | unconfigured → empty config + 200; configured → read the stored state and report `Enabled` / `Suspended` |
| empty config + 200 when unconfigured | `return s3response.GetBucketVersioningOutput{}, nil` | `Status` is a `*types.BucketVersioningStatus`, so the zero value is `nil` → 200 with an empty `VersioningConfiguration` |
| `PutBucketVersioning` keeps rejecting | `PutBucketVersioning`, unchanged | `GetAPIError(ErrVersioningNotConfigured)` → 501 |

For the case you called out specifically — a bad bucket name while gateway versioning is disabled — this yields `InvalidBucketName` instead of `VersioningNotConfigured`, because the name check runs ahead of `versioningEnabled()`.

One thing that may help: the configured path already returns an empty configuration with 200 when a bucket has never had versioning set (`meta.ErrNoSuchKey` → `return s3response.GetBucketVersioningOutput{}, nil`). So this makes the unconfigured-gateway case consistent with a response shape the function already produced, rather than introducing a new one.

I exercised the four paths against the posix backend to confirm:

```
invalid name           -> InvalidBucketName
non-existent bucket    -> NoSuchBucket
existing bucket        -> 200, Status == nil
PutBucketVersioning    -> VersioningNotConfigured
```

Does that match what you had in mind? If I've misread the intent anywhere, I'll happily rework it.

I'm also happy to add those four cases as `backend/posix` unit tests if you'd like them guarded in CI — the existing `VersioningDisabled_*` integration tests only run via `versitygw test posix` with versioning unconfigured, which no CI job currently exercises.

lhjnano: Went ahead and added the unit tests I mentioned: `backend/posix/versioning_test.go` now guards the validation ordering discussed above — invalid name → `InvalidBucketName`, missing bucket → `NoSuchBucket`, existing bucket → 200 with an empty config, and Put still rejecting with `VersioningNotConfigured`.

benmcclelland: looks good now, thanks

benmcclelland (approved):

PR #2300, opened 2026-08-18
feat(azure): carry Azure blob marker in continuation token for pagination

nileger:
## Problem

The Azure backend passed the S3 marker / continuation token straight to Azure
as a blob-listing marker. Azure blob markers are **opaque values that only
Azure may mint** — an S3 object key is not a valid marker. As a result,
paginated `ListObjectsV2` calls against Azure either failed or forced every
page to re-scan the prefix from the beginning, which does not scale.

This is exactly the issue we faced. The performance became really poor for buckets with many object.

## Change

Introduce `azMarkerToken`, an opaque S3 continuation token that carries the
Azure marker alongside the last key of the previous page:

- the **Azure marker** resumes the underlying blob listing where it stopped, and
- the **last key** filters out entries of that Azure page already returned.

Tokens are versioned with a `vgw1.` prefix. Anything without the prefix is
treated as a plain key, so continuation tokens issued by older versions and
hand-crafted markers keep working.

A shared `listBlobs` helper now backs both `ListObjects` and `ListObjectsV2`,
applying the S3 marker and delimiter **client side** (Azure has no equivalent
of "start after"):

- `ListObjectsV2` pages large listings efficiently via the token.
- `ListObjects` (v1) has no opaque token to carry state, so it walks the prefix
  from the start each page — callers who need efficient paging should use
  `ListObjectsV2`.

## Tests

- **Unit tests** (`backend/azure/azure_test.go`), driven by a fake Azure
  container that pages with opaque markers and rejects S3 keys the way real
  Azure does:
  - token round-trip / back-compat / corrupt / prefix-mismatch decoding
  - marker-resumed pagination over 2500 objects, asserting linear request count
  - delimiter + client-side marker filtering
  - common-prefix dedup across page and truncation boundaries
  - multipart-staging blobs filtered before counting toward `maxKeys`
- **Integration tests** (`tests/integration/`): full token-driven pagination and
  delimited pagination, asserting every object/prefix is returned exactly once,
  in order, with correct truncation/token semantics.

Verified with `gofmt -s`, `staticcheck`, `go test -race`, and the full Azurite
integration suite (`full-flow --azure`): **544 passed, 0 failed**.

### DIFF
--- a/backend/azure/azure.go
+++ b/backend/azure/azure.go
@@ -745,6 +745,173 @@ func (az *Azure) GetObjectAttributes(ctx context.Context, input *s3.GetObjectAtt
 	}, nil
 }
 
+// azMarkerToken is the decoded form of an S3 continuation token. Azure blob
+// markers are opaque values that only Azure may mint, so an S3 key can never be
+// used as one. The token therefore carries the Azure marker alongside the last
+// key of the previous page: the Azure marker resumes the blob listing close to
+// where it stopped, and the last key filters out the entries of that Azure page
+// which were already returned.
+type azMarkerToken struct {
+	Prefix  string `json:"p,omitempty"`
+	Marker  string `json:"m,omitempty"`
+	LastKey string `json:"k,omitempty"`
+}
+
+// azTokenPrefix marks continuation tokens minted by this backend. Anything
+// without it is treated as a plain key, which keeps tokens issued by older
+// versions and hand-crafted markers working.
+const azTokenPrefix = "vgw1."
+
+func encodeAzMarkerToken(t azMarkerToken) string {
+	data, err := json.Marshal(t)
+	if err != nil {
+		// the struct only holds strings, so this can't fail
+		return t.LastKey
+	}
+	return azTokenPrefix + base64.RawURLEncoding.EncodeToString(data)
+}
+
+// decodeAzMarkerToken returns the Azure marker and the last key encoded in the
+// continuation token. The Azure marker is only returned for a token minted for
+// the same prefix, as markers are tied to the listing that produced them.
+func decodeAzMarkerToken(token, prefix string) (azureMarker, lastKey string) {
+	if !strings.HasPrefix(token, azTokenPrefix) {
+		return "", token
+	}
+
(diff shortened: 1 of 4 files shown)


versity-github:

benmcclelland:
This looks ok to me. I think the issue was the azurite test system uses the blob names as the marker, so it was assumed this was sufficient. But clearly not how the real azure blob service works.

My only concern here is that the max-keys is probably not going to be aligned to the max blob listing counts. So we are still likely to be listing the same page from azure over and over until the s3 listing passes that page. I'm not sure if there is further optimization that could be done to prevent listing the objects in azure that have already been handled in the s3 responses.

Other minor concern is that the continuation token isnt bound to the delimiter, so reusing the token with a new delimiter might produce unexpected results. Since the token is encoding other request values, I wonder if the delimiter should be included as well.

Overall this looks like a good improvement even if the s3/azure blob listing count alignment cant be improved.

nileger:
> This looks ok to me. I think the issue was the azurite test system uses the blob names as the marker, so it was assumed this was sufficient. But clearly not how the real azure blob service works.
> 
> My only concern here is that the max-keys is probably not going to be aligned to the max blob listing counts. So we are still likely to be listing the same page from azure over and over until the s3 listing passes that page. I'm not sure if there is further optimization that could be done to prevent listing the objects in azure that have already been handled in the s3 responses.
> 
> Other minor concern is that the continuation token isnt bound to the delimiter, so reusing the token with a new delimiter might produce unexpected results. Since the token is encoding other request values, I wonder if the delimiter should be included as well.
> 
> Overall this looks like a good improvement even if the s3/azure blob listing count alignment cant be improved.

I think the first part cannot be improved further, but looking into this a bit more.

Regarding the second part: The token is now bound to the delimiter too. Good catch! 🤝🏻

nileger:
Checked on the first part again. Unfortunately, I came to the conclusion that this cannot be improved further.

benmcclelland (dismissed):

benmcclelland:
needs rebase to resolve conflicts

nileger:
@benmcclelland successfully rebased ✅

benmcclelland (approved):

PR #2306, opened 2026-08-23
fix: reject unsupported RDMA RC token schemes with 501

potatogim:
The RDMA token parser in the cuObject middleware accepts only the cuObject combined-token format, whose first colon-delimited field is a hex-encoded uint64 base address (at most 16 hex chars). Clients using fixed-width binary RC token schemes, such as AMD hipObject (a 44-byte payload hex-encoded to 88 chars, optionally suffixed with ":addr:size"), fail that parse and receive a 400 Bad Request. Since such clients decide between RDMA and the HTTP data path by looking for the x-amz-rdma-reply header, a 400 blocks their fallback entirely.

This rejects structurally incompatible token schemes with 501 Not Implemented instead. A first field longer than 16 hex chars cannot be a valid cuObject base address, so well-formed cuObject tokens never hit the new path, and non-hex fields still fall through to the existing malformed-token 400. The response is a standard S3 XML error sent directly from the middleware (the global error handler collapses non-fiber errors into 500), and the x-amz-rdma-reply header is deliberately left unset so RDMA-capable clients fall back to the HTTP path for this request.

Unit tests cover the RC token shapes (88-hex with and without suffix, uppercase, 17-char leading-zero boundary), pass-through of well-formed cuObject tokens, the malformed and 16-hex-no-colon 400 paths, and legacy descriptor precedence.

### DIFF
--- a/cumiddleware/cuobj.go
+++ b/cumiddleware/cuobj.go
@@ -33,6 +33,9 @@ import (
 
 	"github.com/gofiber/fiber/v3"
 	"github.com/valyala/fasthttp"
+
+	"github.com/versity/versitygw/s3api/utils"
+	"github.com/versity/versitygw/s3err"
 )
 
 // String keys used for fasthttp user-value storage.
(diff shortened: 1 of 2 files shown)


versity-github:

benmcclelland (approved):

PR #2318, opened 2026-08-27
test: separate package download and install for linux/amd config

lrm25:


tests/download_dependencies.sh
```diff
@@ -0,0 +1,191 @@
...
+}
+
+download_mc() {
+  if [ $# -gt 1 ]; then
+    echo "download_mc should only have optional download folder" >&2
+    return 1
+  fi
+  local download_folder="$1"
+
+  if ! response=$(get_mc_download_link "$os" 2>&1); then
```
anodelman, line 131: $os isn't set in this scope?

anodelman (changes requested):

anodelman (approved):

PR #2321, opened 2026-08-28
fix: harden posix conditional publish locking

benmcclelland:
Fail closed on unexpected advisory-lock errors instead of silently reducing cross-process exclusion to a local mutex. Make local publish-slot waits honor request cancellation. Move version snapshots under the publish lock so concurrent versioned PUTs preserve publication order. Add regression coverage for canceled lock waiters.

### DIFF
--- a/backend/posix/objlock.go
+++ b/backend/posix/objlock.go
@@ -38,13 +38,15 @@ import (
 // contention within one process is resolved cheaply and each process presents
 // at most one waiter to the filesystem lock.
 //
-// Lock identity: bucket/.sgwtmp/objlock/<shard>, where shard is the first
-// byte of sha256(object key) rendered as two hex characters. Hashing the key
-// gives a stable, traversal-safe, fixed-length name; sharding (256 slots per
-// bucket) keeps the number of lock files bounded while still letting writes
-// to unrelated keys proceed concurrently in the common case. Lock files are
-// empty, created on demand, and never unlinked: unlinking an flock file opens
-// a classic race where a waiter holds a lock on an unlinked inode while a new
+// Lock identity: <root>/.vgwlocks/<bucket-hash>/<shard>, where shard is the
+// first byte of sha256(object key) rendered as two hex characters. Hashing the
+// bucket and key gives stable, traversal-safe, fixed-length names; sharding
+// (256 slots per bucket) keeps the number of lock files bounded while still
+// letting writes to unrelated keys proceed concurrently in the common case.
+// Lock files are outside the bucket tree so Windows bucket deletion cannot
+// fail merely because an active request has the lock file open. They are empty,
+// created on demand, and never unlinked: unlinking an flock file opens a
+// classic race where a waiter holds a lock on an unlinked inode while a new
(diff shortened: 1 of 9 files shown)


backend/posix/objlock.go
```diff
@@ -83,46 +85,65 @@ func objLockShard(object string) uint8 {
...
+	}
+	return slots
+}
+
 // openObjLockFile opens (creating as needed) the lock file for the shard in
 // the given bucket.
 func (p *Posix) openObjLockFile(bucket string, shard uint8) (*os.File, error) {
-	name := filepath.Join(bucket, objLockDir, fmt.Sprintf("%02x", shard))
+	bucketHash := sha256.Sum256([]byte(bucket))
+	lockDir := filepath.Join(p.rootdir, objLockDir, fmt.Sprintf("%x", bucketHash))
```
niksis02, line 150: Moving locks under p.rootdir can break cross-gateway exclusion when buckets are shared through bucket-level mounts or symlinks but each gateway has a host-local root: each process locks a different inode, allowing competing conditional PUTs to both succeed. It also changes the path used by current main, so old and new binaries will not contend during a rolling upgrade

backend/posix/objlock.go
```diff
@@ -63,8 +65,8 @@ import (
 // process-local exclusion and logs a warning once.
 
 const (
-	// objLockDir is the per-bucket directory holding object publish lock files
-	objLockDir = MetaTmpDir + "/objlock"
+	// objLockDir is the root directory holding object publish lock files.
+	objLockDir = ".vgwlocks"
```
niksis02, line 69: Once the first publish creates `.vgwlocks`, `ListBucketsAndOwners` treats it as a bucket because that path enumerates every root directory without filtering invalid or reserved names. The admin bucket-list endpoint will therefore return .vgwlocks with an empty owner. Explicitly exclude `objLockDir` from root-level bucket discovery, or store locks outside the enumerated backend root?

backend/posix/objlock.go
```diff
@@ -83,46 +85,65 @@ func objLockShard(object string) uint8 {
...
 	shard := objLockShard(object)
 
-	mu := &p.objLockMus[shard]
-	mu.Lock()
+	slot := p.objLockSlots[shard]
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case <-slot:
+	}
```
niksis02, line 93: If cancellation and slot release become ready together, Go may select <-slot. The previous holder has already released the filesystem lock, so lockFileExclusive can immediately succeed without observing the canceled context and the PUT may continue publishing. Check ctx.Err() immediately after receiving the slot — releasing the slot before returning—to make cancellation deterministic.

niksis02 (changes requested):

backend/posix/objlock.go
benmcclelland: i dont think we care about either of these cases since this isnt in an actual release version yet and we only care about races between gateways at the same root.

niksis02 (approved):

PR #2324, opened 2026-08-28
fix: apply default Content-Type in POSTObject

### LINKED ISSUE #2307: [Bug] - POSTObject (browser-based upload) doesn't apply the default binary/octet-stream Content-Type
**Describe the bug**

When an object is uploaded via a browser-based POST (`POST /{bucket}`, presigned POST
policy) and the form contains no `Content-Type` field, the object is stored with an
**empty** Content-Type instead of the S3 default `binary/octet-stream`.

`POSTObject` reads the value straight out of the form fields with no fallback:

```go
// s3api/controllers/bucket-post.go:108
contentType := parsed.Fields["content-type"]
```

The equivalent header-driven paths all apply the default that #1849 / #1865 added:
```go
// s3api/controllers/object-put.go:481 (PutObject), :638 (CopyObject)
// s3api/controllers/object-post.go:150 (CreateMultipartUpload)
contentType := ctx.Get("Content-Type", defaultContentType)   // base.go:57 = "binary/octet-stream"
```
POSTObject looks to have been missed simply because of timing: #1865 merged
2026-02-18, and browser-based POST upload support (#1982) landed 2026-03-31.

The empty value has a surprising side effect on read. GetObject/HeadObject set
the header from the stored value (s3api/controllers/object-get.go:559), and because
it is an empty string, fasthttp substitutes its own default:

```go
// fasthttp/header.go — ResponseHeader.ContentType()
if !h.noDefaultContentType && len(h.contentType) == 0 { contentType = defaultContentType }
// fasthttp/strings.go
defaultContentType = []byte("text/plain; charset=utf-8")
```
So the object comes back as Content-Type: text/plain; charset=utf-8 — not just
missing, but actively mislabeled as text.

To Reproduce

1. Create a presigned POST policy with no Content-Type condition and upload a PNG
without a Content-Type form field:

```js
import boto3, requests
s3 = boto3.client("s3", endpoint_url="http://localhost:7070",
                  aws_access_key_id="...", aws_secret_access_key="...")
p = s3.generate_presigned_post("mybucket", "icon.png", ExpiresIn=3600)
with open("icon.png", "rb") as f:
    r = requests.post(p["url"], data=p["fields"], files={"file": f})
print(r.status_code)  # 204
```

2. Inspect the stored object:
```
$ aws --endpoint-url http://localhost:7070 s3api head-object \
      --bucket mybucket --key icon.png
...
"ContentType": "text/plain; charset=utf-8"

$ curl -sI "<presigned GET url for icon.png>" | grep -i content-type
content-type: text/plain; charset=utf-8
```

3. For contrast, the same file via PutObject with no Content-Type header is stored
as binary/octet-stream, as expected after #1865.

Expected behavior

POSTObject should apply defaultContentType when the form has no content-type
field, matching `PutObject/CopyObject/CreateMultipartUpload` and AWS S3. Both AWS
S3 and Ceph RGW return `binary/octet-stream` for this exact upload.

A one-line change appears to be enough:

```go
contentType, ok := parsed.Fields["content-type"]
if !ok || contentType == "" {
    contentType = defaultContentType
}
```

It may also be worth having the read path never emit an empty `Content-Type`, so
existing objects already stored with an empty value don't get fasthttp's `text/plain`
default.

Server Version

`v1.4.1`

Reproduced on **v1.4.1**; `s3api/controllers/bucket-post.go:108` is unchanged on main
as of today, so this should still be present in **v1.7.0**.

Related: #1849, #1865 (original default Content-Type fix), #1982 (POST upload support).

mu-adnan:
Browser-based POST uploads (POST `/{bucket}`) read Content-Type straight out of the form fields, so a form without a content-type field stored the object with an empty Content-Type. On read, fasthttp substitutes its own default, so the object came back as `text/plain; charset=utf-8` rather than just missing a type.

Fall back to defaultContentType (`binary/octet-stream`) when the form field is absent or empty, matching PutObject, CopyObject and CreateMultipartUpload, as well as AWS S3 and Ceph RGW.

Fixes: #2307

### DIFF
--- a/s3api/controllers/bucket-post.go
+++ b/s3api/controllers/bucket-post.go
@@ -118,6 +118,9 @@ func (c S3ApiController) POSTObject(ctx fiber.Ctx) (*Response, error) {
 	parsed := utils.ContextKeyObjectPostResult.Get(ctx).(middlewares.PostObjectResult)
 	bucket := ctx.Params("bucket")
 	contentType := parsed.Fields["content-type"]
+	if contentType == "" {
+		contentType = defaultContentType
+	}
 	contentEncoding := parsed.Fields["content-encoding"]
 	contentDisposition := parsed.Fields["content-disposition"]
 	contentLanguage := parsed.Fields["content-language"]
--- a/s3api/controllers/bucket-post_test.go
+++ b/s3api/controllers/bucket-post_test.go
@@ -621,6 +621,47 @@ func TestS3ApiController_POSTObject(t *testing.T) {
 			},
 		},
 		{
+			name: "default content type when field is omitted",
+			input: testInput{
+				beRes: s3response.PutObjectOutput{
+					ETag: "etag-123",
+				},
+				locals: postObjectLocalsForTest(middlewares.PostObjectResult{
+					Fields:        baseFields,
+					FileRdr:       newMockFileReader("payload"),
+					ContentLength: int64(len("payload")),
+				}),
+			},
+			output: testOutput{
+				response: &Response{
+					Headers: map[string]*string{
+						"Etag":                     utils.GetStringPtr("etag-123"),
+						"Location":                 &location,
+						"x-amz-checksum-crc32":     nil,
+						"x-amz-checksum-crc32c":    nil,
+						"x-amz-checksum-crc64nvme": nil,
+						"x-amz-checksum-sha1":      nil,
+						"x-amz-checksum-sha256":    nil,
+						"x-amz-checksum-sha512":    nil,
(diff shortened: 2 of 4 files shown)


versity-github:

niksis02 (changes requested):
It would be nice to add an integration test covering this case in `tests/integration/PostObject.go`.

mu-adnan:
> It would be nice to add an integration test covering this case in `tests/integration/PostObject.go`.

@niksis02 done

niksis02 (approved):

PR #2325, opened 2026-08-29
auth: add GetUserAccountFresh to bypass the IAM cache

potatogim:
Adds `GetUserAccountFresh` to the IAM cache, which bypasses the in-memory entry, reads the account directly from the underlying IAM service, and refreshes the cached entry with the result.

The cached `GetUserAccount` keeps serving the previous account for the configured TTL after the backing service changes, so credential revocation can lag by up to the cache expiry. The fresh variant lets callers that need revocations to take effect immediately observe the backing store state, without changing the behavior of the existing cached path.

The motivating caller is a multi-step RDMA transfer protocol where a session authorized at the first request must be re-checked at a later request within the session window; the test covers stale-entry bypass and cache refresh. No existing call sites change.

### DIFF
--- a/auth/iam_cache.go
+++ b/auth/iam_cache.go
@@ -177,6 +177,21 @@ func (c *IAMCache) ResolveAccounts(accessKeyIDs []string) ([]string, error) {
 	return resolveAccountsByLookup(accessKeyIDs, c.GetUserAccount)
 }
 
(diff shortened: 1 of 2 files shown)


versity-github:

benmcclelland (approved):

Issue #2330, opened 2026-08-29
[Feature] - RC data plane for vgwrdma (hipobj-rc-v2)


potatogim:
**Describe the solution you'd like**

Add a reliable-connection (RC) data plane to vgwrdma so hipObject clients that use two-phase prepare/ready transfers - QP pairing, staging MRs, cookie/PSN validation - can use the gateway as their object endpoint. Today vgwrdma speaks the cuObject single-request (ud-style) protocol only.

The work lands as a four-PR stack, each part reviewable on its own:

1. Session core: port the hipObject v2 RC session core into `cuwrapper/rc` - the session table and state machine, the hipobj-rc-v2 wire codec, QP/CQ lifecycle with INIT/RTR/RTS transitions, the data phase (RDMA write with immediate for GET, receive with immediate for PUT), injectable clock/randomness sources, the token codec, and a dlopen-based ibverbs shim.

2. C ABI: the server layer over that core - device lifecycle, prepare/ready/cancel/staging/put-view calls keyed by consume-once handles, global and per-principal resource limits, the READY completion-reference claim with rollback, reaping with deadline expiry, retryable peer-busy handling, and teardown that keeps verbs objects alive when a destroy fails.

3. Go binding: a cgo package (`rdma/rcserver`) with an admission gate around every call, a service-lifetime context for shutdown, and an idempotent Close that drains and destroys the server; stubbed on non-linux/non-cgo builds so `go vet ./...` keeps passing on standard CI.

4. Gateway routes: the three control routes (`POST /.hipobj-rc/prepare`, `/.hipobj-rc/ready`, `/.hipobj-rc/cancel`) behind the standard SigV4 middleware, fresh IAM revalidation between protocol phases, re-authorization against the percent-decoded bucket and key, readonly enforcement, the session ownership contract on READY, and the vgwrdma start/stop wiring.

Together the four parts give vgwrdma a full RC data plane: clients prepare a session, pair QPs, transfer through staging MRs, and finalize with object metadata, under the same auth, ACL, and policy enforcement as the S3 API.

PR #2331, opened 2026-08-29
rdma: port the hipObject v2 RC session core

potatogim:
Part of #2330.

Ports the hipObject v2 reliable-connection session core into `cuwrapper/rc` as the foundation for an RC data plane on vgwrdma. The port covers the session table and state machine, the hipobj-rc-v2 wire codec and request parsing, QP/CQ lifecycle with INIT/RTR/RTS transitions, staging MR registration, the data phase (RDMA write with immediate for GET, receive with immediate for PUT), injectable clock/randomness sources, the RDMA token codec, and a dlopen-based ibverbs shim.

The sources stay close to the upstream hipObject v2 core so the trees can be diffed during review. A Makefile rule builds the objects into `rdma/librcserver.a`; the archive is not linked into any binary yet, that happens with the ABI and binding PRs up the stack.


versity-github:

benmcclelland (approved):

PR #2332, opened 2026-08-29
rdma: add the RC session server C ABI

potatogim:
Part of #2330.

Builds on #2331.

The diff shown here is against `main`, so it includes the session core from #2331; the incremental change on top of that PR is `cuwrapper/rc/rc_server_abi.{h,cpp}` and the Makefile link change.

Adds the server-side C ABI over the ported v2 core (`cuwrapper/rc/rc_server_abi.{h,cpp}`): server lifecycle with GID-hint device selection, prepare/ready/cancel/staging/put-view session calls keyed by consume-once handles, global and per-principal resource limits, the READY completion-reference claim with rollback, reaping (mutating-call passes plus a deadline expiry thread), retryable peer-busy handling that re-arms the QP through RESET, and teardown that keeps verbs objects alive when a destroy or deregistration fails rather than freeing memory the NIC may still reference.

The `vgwrdma` target now links `rdma/librcserver.a`, which from this PR on includes the ABI objects.


versity-github:

benmcclelland (approved):

PR #2333, opened 2026-08-29
rdma: add the Go binding for the RC session server

potatogim:
Part of #2330.

Builds on #2332.

The diff shown here is against `main`, so it includes the session core and the C ABI from #2331 and #2332; the incremental change on top of those PRs is the `rdma/rcserver` package.

Binds the RC session server archive through cgo (`rdma/rcserver`). `Init` opens the verbs device with the configured limits; the `RCSvc` wrapper carries the admission gate (`TryEnter`/`Leave`) handlers use around every session call, a service-lifetime context that `Close` cancels first so in-flight handler I/O unblocks, and an idempotent `Close` that marks sessions for reaping, waits for the admission count to drain, and destroys the server. The session surface maps the C ABI one-to-one, with the READY transfer outcome returned atomically in the reply. Non-linux/non-cgo builds compile against a stub so the package stays portable.

Nothing imports the package yet, so no binary links it; the gateway integration in the next PR is where the cgo link gets exercised.

### DIFF
--- a/rdma/rcserver/rcserver_linux.go
+++ b/rdma/rcserver/rcserver_linux.go
@@ -1,0 +1,564 @@
+// Copyright 2026 Versity Software
+// This file is licensed under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


versity-github:

benmcclelland (approved):

PR #2334, opened 2026-08-29
rdma: add the hipobj-rc-v2 control routes to the vgwrdma gateway

### LINKED ISSUE #2330: [Feature] - RC data plane for vgwrdma (hipobj-rc-v2)

potatogim:
Fixes #2330.

Builds on #2333.

The diff shown here is against `main`, so it includes the session core, the C ABI, and the Go binding from #2331, #2332, and #2333; the incremental change on top of those PRs is the `rdma/rcroutes` package, the vgwrdma and embedgw wiring, and the `IAMCache` lookup method.

Mounts the three control routes (`POST /.hipobj-rc/prepare`, `/.hipobj-rc/ready`, `/.hipobj-rc/cancel`) on the S3 port behind the standard SigV4 middleware and wires the session server into the vgwrdma lifecycle.

The routes own the auth-adjacent policy the C server cannot see: READY and CANCEL re-read the account through a cache-bypassing IAM lookup (bounded against a stalled IAM by a small lookup pool) so mid-flow deletions and credential rotations take effect immediately; object access re-authorizes against the percent-decoded bucket and key; readonly gateways refuse RC PUTs like regular writes. The READY handler implements the session ownership contract: the completion-reference finalizer installs only after the transfer claim succeeds, the PUT path hands the reference to the put view exactly at the borrow point, the FINAL reply carries the stored object's ETag and version, and a peer-busy outcome answers 409 without a finalizer so the client can retry the same session.

Backend I/O runs under a context merged with the RC service context so `Close`, which runs after the gateway returns, converges without deadlocking. vgwrdma starts the session server when an RDMA interface is configured and tears it down on exit; `embedgw` passes the readonly flag into the shared object access checks; `IAMCache` gains one cache-bypassing lookup method. With this PR the stack is complete and vgwrdma has the full RC data plane.

### DIFF
--- a/cmd/vgwrdma/main.go
+++ b/cmd/vgwrdma/main.go
@@ -130,10 +130,6 @@ var (
 	BuildTime = "none"
 )
 
-type freshAccountReader interface {
-	GetUserAccountFresh(access string) (auth.Account, error)
-}
-
 type standaloneIAMExtensions interface {
 	auth.SigningKeyProvider
 	auth.PolicyEvaluator
@@ -153,37 +149,17 @@ func (s *shutdownOnceService) Shutdown() error {
 	return s.err
 }
 
-type shutdownOnceFreshService struct {
-	*shutdownOnceService
-	freshAccountReader
-}
-
 type shutdownOnceStandaloneService struct {
 	*shutdownOnceService
 	standaloneIAMExtensions
 }
 
-type shutdownOnceFreshStandaloneService struct {
-	*shutdownOnceService
-	freshAccountReader
-	standaloneIAMExtensions
-}
-
 func wrapIAMShutdownOnce(iam auth.IAMService) auth.IAMService {
 	wrapped := &shutdownOnceService{IAMService: iam}
-	fresh, hasFresh := iam.(freshAccountReader)
-	standalone, hasStandalone := iam.(standaloneIAMExtensions)
-
-	switch {
-	case hasFresh && hasStandalone:
-		return &shutdownOnceFreshStandaloneService{wrapped, fresh, standalone}
-	case hasFresh:
-		return &shutdownOnceFreshService{wrapped, fresh}
-	case hasStandalone:
+	if standalone, ok := iam.(standaloneIAMExtensions); ok {
 		return &shutdownOnceStandaloneService{wrapped, standalone}
-	default:
-		return wrapped
 	}
+	return wrapped
 }
 
 // gatewayCommands are the subcommands that call gwcli.RunGateway (and
--- a/cmd/vgwrdma/main.go
+++ b/cmd/vgwrdma/main.go
@@ -24,7 +24,9 @@ import (
 	"os"
 	"strings"
 
+	"github.com/gofiber/fiber/v3"
 	"github.com/urfave/cli/v2"
+	"github.com/versity/versitygw/auth"
 	"github.com/versity/versitygw/backend"
 	"github.com/versity/versitygw/cmd/internal/gwcli"
 	"github.com/versity/versitygw/cubackend"
@@ -33,7 +35,10 @@ import (
 	"github.com/versity/versitygw/embedgw"
 	"github.com/versity/versitygw/internal/netutil"
 	"github.com/versity/versitygw/rdma"
+	"github.com/versity/versitygw/rdma/rcroutes"
+	"github.com/versity/versitygw/rdma/rcserver"
 	"github.com/versity/versitygw/s3api"
+	"github.com/versity/versitygw/s3api/middlewares"
 )
 
 var (
@@ -104,6 +109,7 @@ var (
 	mpMaxParts                             int
 	socketPerm                             string
 	rdmaIP                                 string
+	rcGidHint                              string
 	rdmaPort                               uint
 	poolBufSize                            int
 	poolBufCount                           int
@@ -840,6 +846,12 @@ func initFlags() []cli.Flag {
 			EnvVars:     []string{"VGW_RDMA_IP"},
 			Destination: &rdmaIP,
 		},
+		&cli.StringFlag{
+			Name:        "rc-gid-hint",
+			Usage:       "dotted GID prefix selecting the verbs device for the hipobj-rc-v2 RC data plane (default: first device)",
+			EnvVars:     []string{"VGW_RC_GID_HINT"},
+			Destination: &rcGidHint,
+		},
 		&cli.UintFlag{
 			Name:        "rdma-port",
 			Usage:       "port for RDMA listener",
@@ -941,6 +953,9 @@ func runGateway(ctx context.Context, be backend.Backend) error {
 	}
 
 	var s3Opts []s3api.Option
+	// iamOwned tracks whether the RC IAM service is still ours to
+	// shut down; it is cleared only after RunVersityGW returns.
+	iamOwned := false
 	if rdmaIP != "" {
 		rdma.ConfigureTelemetry(debug)
 
@@ -969,7 +984,7 @@ func runGateway(ctx context.Context, be backend.Backend) error {
 		s3Opts = append(s3Opts, s3api.WithMiddleware("/", cumiddleware.CuObjMiddleware))
 	}
 
-	return embedgw.RunVersityGW(ctx, be, &embedgw.Config{
+	cfg := &embedgw.Config{
 		RootUserAccess:              gwcli.RootUserAccess,
 		RootUserSecret:              gwcli.RootUserSecret,
 		Region:                      region,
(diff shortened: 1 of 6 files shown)


versity-github:

cmd/vgwrdma/main.go
```diff
@@ -1065,9 +1080,83 @@ func runGateway(ctx context.Context, be backend.Backend) error {
...
+			if err := rcVerify(ctx); err != nil {
+				return err
+			}
+			return ctx.Next()
+		}
+		rcH := rcroutes.New(rcSvc, be, iamSvc, readonly, disableACLs)
+		cfg.S3Options = append(s3Opts,
+			s3api.WithRoute("POST", "/.hipobj-rc/prepare", rcAuth, rcH.Prepare),
+			s3api.WithRoute("POST", "/.hipobj-rc/ready", rcAuth, rcH.Ready),
+			s3api.WithRoute("POST", "/.hipobj-rc/cancel", rcAuth, rcH.Cancel),
```
benmcclelland, line 1172: should this call the Handler.Register in routes_stub.go to setup these routes instead? or else we should remove the unused route setup in Register().

cmd/vgwrdma/main.go
```diff
@@ -1065,9 +1080,83 @@ func runGateway(ctx context.Context, be backend.Backend) error {
...
+		}
+		cfg.IAMService = iamSvc
+		// The RC routes share this IAM service with the gateway.
+		// If a later step fails before RunVersityGW takes over,
+		// shut it down here so background workers do not leak;
+		// iamOwned is cleared after RunVersityGW returns.
+		iamOwned = true
+		defer func() {
+			if iamOwned {
+				_ = iamSvc.Shutdown()
```
benmcclelland: embedgw.RunVersityGW unconditionally runs iam.Shutdown() near the end of its shutdown sequence (whether it returns nil or a non-nil saveErr from a runtime failure, not just clean exit). this will call iamSvc.Shutdown() again in this defer. current iam servers are ok with this, but we should clean this up.

rdma/rcroutes/routes_linux.go
```diff
@@ -0,0 +1,722 @@
...
+	if resp.Etag != "" {
+		ctx.Set(hdrEtag, resp.Etag)
+	}
+	if resp.VersionID != "" {
+		ctx.Set(hdrVersionID, resp.VersionID)
+	}
+	return ctx.SendStatus(fiber.StatusOK)
+}
+
+func sizeOf(_ *rcserver.SessionInfo, resp *rcserver.ReadyResponse) uint64 {
```
benmcclelland: is there a reason *rcserver.SessionInfo is added here? looks unused and can be removed.

rdma/rcroutes/routes_linux.go
```diff
@@ -0,0 +1,722 @@
...
+//
+// The lookup runs bounded by the RC service context: the IAM
+// interface carries no context (its backends are shared with the
+// rest of the gateway), so an external IAM that stalls could
+// otherwise hold this handler - and through it the RCSvc.Close
+// ops drain - forever. When the service shuts down mid-lookup
+// the request fails instead of blocking the drain. The semaphore
+// caps how many lookup goroutines a stalled IAM can leave
+// behind: past the cap, further requests fail fast with 503.
+func (h *Handler) revalidateFresh(acct auth.Account) (auth.Account, error) {
```
benmcclelland: I think I see the reason for the revalidateFresh(): exists because PREPARE and READY/CANCEL are two separate HTTP requests separated by an arbitrary amount of time (the client does RDMA setup in between), and acct for READY/CANCEL comes from utils.ContextKeyAccount, which was populated by the SigV4 middleware from the (likely cached) IAMService.GetUserAccount lookup at the top of the request, not from PREPARE's earlier lookup.

Is that correct understanding?

I have concerns with overloading the backend IAM service under heavy load of simultaneous requests. This currently only bounds concurrency, but not the overall rate. And there could be significant latency on requests to IAM services such as LDAP.

Maybe a better approach would be more tunables for adjusting the cache TTL? Or is there some other error case where we are not getting the correct behavior?

benmcclelland (changes requested):

cmd/vgwrdma/main.go
potatogim: Good catch, thanks. As far as I can tell Register had no call sites, so I removed it from both build variants in 8b8a360 and kept the three s3api.WithRoute registrations as the only route setup.

One thing I found while removing it: the Linux Register also placed VerifyV4Signature first, and that middleware returns nil on success instead of calling Next, so I don't think it could have served as a drop-in replacement anyway. The rcAuth wrapper calls Next only after successful verification, so I believe the authentication path is equivalent to what Register intended. On unsupported builds rcserver.Init fails before any RC route could be mounted, so I don't think the stub removal changes observable behavior there, but happy to reconsider if you'd prefer the routes to exist there.

cmd/vgwrdma/main.go
potatogim: You're right, sorry about that. While fixing it I found the ownership was harder to express than the flag suggested: as far as I can tell, RunVersityGW's non-nil return mixes two cases - early setup failures that return before any IAM shutdown, and runtime failures that return after the service was already shut down - so I couldn't distinguish them from the return value alone.

I reworked this in 6f845aa: the IAM service is wrapped so Shutdown runs exactly once regardless of which side calls it, the deferred shutdown still covers every early failure path, and the gateway error is propagated to the caller again. The wrapper also re-exposes the optional interfaces other parts of the stack probe for (fresh account reads, signing keys, policy evaluation, fixed bucket ownership) - I might have missed something there, so please take a look if you see a cleaner way to handle that.

rdma/rcroutes/routes_linux.go
potatogim: You're right, that parameter was unused - sorry, it should not have been there. I removed it in 8b8a360. The SessionInfo lookup itself stays because READY uses it for owner validation and for the target/op reauthorization.

rdma/rcroutes/routes_linux.go
potatogim: Yes. More precisely, PREPARE and READY/CANCEL are separate signed HTTP requests. On a non-root request using IAMCache, SigV4 may perform one backing IAM read on a cache miss, and READY/CANCEL then perform one cache-bypassing read. PREPARE does not call revalidateFresh, and each retry is a new request.

Deletion may not be the only correctness case. Secret rotation changes the SHA-256(access:secret) principal, so an old prepared session would fail the READY/CANCEL owner check. A role change could also alter READY's authorization result because READY passes the freshly read account to the authorization check; CANCEL only rechecks credential ownership. I considered a shorter global IAM cache TTL instead, but it would not guarantee a fresh read at the PREPARE-to-READY boundary and would lower the cache hit rate for other non-root signed requests, so I went with the explicit re-read. Happy to revisit that choice.

Your load concern is valid. The current limit of eight bounds concurrency, not the aggregate request rate, and READY/CANCEL perform the fresh read before validating the required session fields, so signed requests with malformed or unknown session IDs can consume those slots and stall other READY/CANCEL flows with 503s while the backing IAM is slow. If it helps, I could move the cheap header validation before the fresh read, make the concurrency cap configurable, and coalesce concurrent reads for the same access key. Coalescing reduces same-key bursts, though it does not limit sequential request rate, and concurrent callers would share the in-flight read's freshness point. Limiting the sustained rate would take a per-key rate limit or a short freshness TTL, which trades availability or bounded staleness against immediate revocation - I'm not sure the extra complexity is warranted, but I'd appreciate your thoughts on which direction you'd prefer.

Separately, 0702734 makes only a confirmed missing account (ErrNoSuchUser) return 403; other fresh IAM lookup failures now return 503 and stay retryable. I suspect this also matters for the revocation-vs-transient distinction you raised, but let me know if you see a case it misses.

cmd/vgwrdma/main.go
benmcclelland: this is fine, mostly just wanted to get rid of duplicated code

rdma/rcroutes/routes_linux.go
benmcclelland: I’m not convinced READY/CANCEL should bypass the IAM cache. The cache TTL is already the gateway’s configured stale-auth window for successive requests, and READY/CANCEL are still separate signed requests in that model. Giving RC routes stricter freshness semantics than normal S3 requests adds backend IAM load and tail-latency risk, especially for LDAP/Vault/S3 IAM, without an obvious correctness requirement that differs from any other cached request after an account deletion or role change.

Can we use the normal cached account from SigV4 here, and rely on the existing cache TTL/operator tuning for revocation latency? If RC needs a stronger guarantee than the rest of the gateway, I think that should be an explicit configurable policy rather than unconditional fresh IAM reads on every READY/CANCEL.

CANCEL is cleanup/control-plane only, so forcing a fresh IAM dependency there is even harder to justify since we may prefer the cache credentials in this case.

potatogim: Yes, I agree. READY and CANCEL are independently authenticated SigV4 requests, and I removed the route-level fresh revalidation from both paths in bc5d643 so they now use the account resolved by the normal SigV4 path, following the same configured IAM cache TTL as other signed S3 requests.

With that in place, an out-of-band account deletion, secret rotation, or role change can remain effective through the remaining cache lifetime, including an old credential completing READY for a session it prepared. That is the same stale-auth window the gateway already accepts for a separate cached S3 request; the owner digest check compares principals and does not detect a deletion while the same stale account is cached. Operators that need tighter revocation can shorten the existing iam-cache-ttl or disable the cache. I also could not think of a stronger-freshness requirement specific to RC that would justify a separate flag, so I did not add one.

Along the way this removes the extra backend IAM lookup and its tail latency, the concurrency cap and saturation 503s, and the RC-specific 403/503 IAM error mapping. The session owner check and the READY reauthorization against the session's stored target and operation are unchanged. Please let me know if you would prefer anything done differently.

benmcclelland (approved):

benmcclelland:
@potatogim Thank you for the all the work on this! much appreciated.

PR #2353, opened 2026-09-03
fix: s3proxy sends bodyless CreateBucket when the configuration is empty

### LINKED ISSUE #1595: [Feature] - Add tagging support on bucket creation
S3 supports bucket tagging on bucket creation. The tags are sent in the request body. Currently the gateway ignores the `CreateBucket` request body. 

The idea is to parse and validate the request body, especially:
1. Tags
2. Location constraint(if incorrect, return an error)

Other fields in the request body has to be ignored.

**Additional context**
[S3 CreateBucket Doc](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateBucket.html)

StefanMarkmann:
## Problem

Since 9bde1ddb (`feat: implements tagging support for CreateBucket`, closes #1595,
shipped in every release since v1.0.19) the api layer always populates
`CreateBucketConfiguration` on the `CreateBucketInput` it hands to the backend —
the field only exists to carry bucket tags. The aws-sdk then serializes an empty
`<CreateBucketConfiguration/>` element (103 bytes) on **every** CreateBucket the
s3proxy backend issues, whether or not the client supplied tags or a location
constraint.

Strict backends reject that request:

| Backend | empty `<CreateBucketConfiguration/>` | bodyless CreateBucket |
|---|---|---|
| Ceph RGW Quincy 17.2.8 (`quay.io/ceph/demo:latest-quincy`) | **400 InvalidArgument** | 200 |
| Ceph RGW Squid 19.2.0 (`quay.io/ceph/demo:latest-squid`) | **400 InvalidArgument** | 200 |
| MinIO | 200 (tolerated) | 200 |
| SeaweedFS | 200 (tolerated) | 200 |

The RGW behavior is longstanding and consistent across releases — not a recent
Ceph change — so every versitygw release since v1.0.19 has been unable to
create buckets on a Ceph backend.

So bucket creation through the gateway is completely broken for s3proxy
deployments in front of Ceph RGW (API, admin UI and any S3 client alike), while
tests against MinIO pass — presumably why this went unnoticed since October
2025. The posix backend is unaffected (no backend HTTP call).

## Fix

In the s3proxy backend, drop `CreateBucketConfiguration` before calling the
backend when it carries no content (no tags, no location constraint, no
location/bucket info). A bodyless CreateBucket is accepted by all tested
backends for this no-location case. Configurations that do carry content are
forwarded unchanged, so the tagging feature from #1595 keeps working. The check lives in the s3proxy backend rather
than the api layer because the posix and azure backends dereference
`input.CreateBucketConfiguration` unconditionally.

## Verification

- **New wire-level unit tests** (`backend/s3proxy/s3_test.go`, the package's
  first): an injected capturing HTTP client asserts what the sdk actually
  sends. `TestCreateBucketOmitsEmptyConfiguration` fails before this fix
  (empty element on the wire, non-zero Content-Length) and passes with it;
  four preservation tests prove that configurations carrying tags (#1595),
  a location constraint, location info, or bucket info are still forwarded.
- Wire capture (socat dump backend) of the full gateway binary: a no-tags
  CreateBucket is now bodyless (`Content-Length: 0`; previously the 103-byte
  empty element), a CreateBucket with `<Tags>` still sends the configuration.
- **Verified end-to-end against Ceph RGW (squid-era)**: CreateBucket through
  the patched gateway succeeds (200, bucket usable and deletable) where the
  unpatched gateway fails with 400 InvalidArgument. The raw-element behavior
  was additionally confirmed with direct requests: bodyless CreateBucket →
  200, empty element → 400, consistent across three RGW nodes.
- `go build ./cmd/versitygw`, `go vet ./backend/s3proxy/`, `gofmt` clean.
- **Regression runs:** full unit suite (`go test -race -tags=github ./...`) all
  green; the functional suite (`runtests.sh`) passes 39/39; `versitygw test
  full-flow` against an s3proxy→posix chain yields byte-identical results with
  and without this fix (188 PASS / 36 FAIL — the 36 are pre-existing chain
  failures unrelated to CreateBucket, identical failure list on both builds).

### DIFF
--- a/backend/s3proxy/s3.go
+++ b/backend/s3proxy/s3.go
@@ -204,6 +204,15 @@ func (s *S3Proxy) CreateBucket(ctx context.Context, input *s3.CreateBucketInput,
 		}
 	}
 
+	// drop an empty CreateBucketConfiguration: strict backends (Ceph RGW) reject the empty element
+	if input.CreateBucketConfiguration != nil &&
+		len(input.CreateBucketConfiguration.Tags) == 0 &&
+		input.CreateBucketConfiguration.LocationConstraint == "" &&
+		input.CreateBucketConfiguration.Location == nil &&
+		input.CreateBucketConfiguration.Bucket == nil {
+		input.CreateBucketConfiguration = nil
+	}
+
 	_, err := s.client.CreateBucket(ctx, input)
 	if err != nil {
 		return handleError(err)
--- a/backend/s3proxy/s3_test.go
+++ b/backend/s3proxy/s3_test.go
@@ -1,0 +1,171 @@
+// Copyright 2023 Versity Software
+// This file is licensed under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package s3proxy
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+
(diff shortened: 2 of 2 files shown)


versity-github:

benmcclelland (approved):

PR #2355, opened 2026-09-04
rdma: report RC route errors through the S3 error boundary

### LINKED ISSUE #2354: [Bug] - RC route errors collapse to 500 and lose protocol diagnostics
**Describe the bug**

After #2334, requests hitting the hipobj-rc-v2 control routes fail in ways the gateway misreports. On a cgo-disabled build the stub answer for an RC route is classified as an internal error, so the client sees a 500 and the log shows an internal-error line for a request that is not implemented on this platform. Transport-level failures inside the RC session layer surface as bare 500s with no S3 error code, and the error XML body drops the per-type diagnostic fields (access key, string-to-sign) the same failures produce on the regular S3 routes, so client tooling shows different output for the same failure depending on which route produced it.

**To Reproduce**

Build vgwrdma with CGO_ENABLED=0 and POST to /.hipobj-rc/prepare (answers a bare 500 instead of 501), or send a request with a bad signature (answers a bare 500 without the diagnostic fields the regular routes return).

**Expected behavior**

The RC routes answer with the same S3 error codes and the same XML body shape as the regular routes: 501 for the platform stub without an internal-error log line, SignatureDoesNotMatch with the access key and string-to-sign for a bad signature, and response body identifiers consistent with the request-id headers.

potatogim:
Fixes #2354.

Reports hipobj-rc-v2 route failures through the S3 error boundary instead of collapsing them to bare 500s. Transport-level failures inside the RC session layer map to S3 protocol codes (bad request for rejected arguments and short or oversized transfers, internal error for the rest), and the response keeps the full S3 error XML body, so per-type diagnostics such as the access key and string-to-sign survive the route boundary the same way they do on the regular S3 routes. Auth failures take the same path, and the platform stub answers 501 on every build, classified before the internal-error logging decision so it no longer logs as an internal 500.

The s3api boundary tests cover the mapping through the production server and link the RC session-server archive on the general CI platform, so the workflow builds it before the test step. Regression tests assert a wrapped signature failure keeps the status, the code, and both diagnostic fields, and that the response body identifiers equal the request-id headers.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -24,6 +24,9 @@ jobs:
       run: |
         go get -v -t -d ./...
 
+    - name: Build RDMA session server archive
+      run: make rdma/librcserver.a
+
     - name: Test
       run: go test -coverprofile profile.txt -race -v -timeout 30s -tags=github ./...
 
--- a/cmd/vgwrdma/main.go
+++ b/cmd/vgwrdma/main.go
@@ -1162,7 +1162,7 @@ func runGateway(ctx context.Context, be backend.Backend) error {
 		// chain as usual.
 		rcAuth := func(ctx fiber.Ctx) error {
 			if err := rcVerify(ctx); err != nil {
-				return err
+				return rcroutes.WriteRouteError(ctx, err)
 			}
 			return ctx.Next()
 		}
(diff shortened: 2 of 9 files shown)


versity-github:

PR #2357, opened 2026-09-04
rdma: gate the RC data plane behind a dedicated flag

### LINKED ISSUE #2356: [Bug] - RC data plane requires the v1 RDMA stack and tears down in the wrong order
**Describe the bug**

After #2334, the hipobj-rc-v2 RC data plane has no dedicated switch: the RC control routes and data plane only run when a v1 RDMA interface is configured (--rdma-ip). Running v2 without v1 - the standalone v2 deployment shape - is impossible, and conversely configuring a v1 interface silently brings up the v2 session server on gateways that did not opt in.

When both stacks run, teardown closes the v1 CuServer before the RC session server has drained, so in-flight RC transfers can hit a dead backend. The rollback path on startup failure only closes the base backend, leaking the v1 resources it had already initialized. The --rdma-cq-depth bound validation is also lost during option parsing: a value above the 32-bit range wraps to zero and is silently accepted.

**To Reproduce**

Start vgwrdma without --rdma-ip: the RC control routes are not mounted, so the RC data plane cannot run standalone. With --rdma-ip the v2 session server starts even though nothing opted in. For the bound, pass --rdma-cq-depth=4294967296: it is accepted.

**Expected behavior**

A dedicated flag admitting v2 independently of --rdma-ip, a startup rollback that closes everything it initialized, a teardown order that drains RC before the v1 stack, and the CQ depth bound enforced.

potatogim:
Fixes #2356.

Builds on #2355.

The diff shown here is against `main`, so it includes the RC route error boundary from the first PR; the incremental change on top of it is the `internal/rdmamode` package, the vgwrdma startup/teardown changes, and the CQ depth bound.

Gates the RC data plane behind a dedicated flag. --rdma-rc-enable brings up the v2 RC session server independently of --rdma-ip (default false; the existing v1 stack is unchanged), and starting neither flag keeps the plain S3 gateway. The startup validation moves into a mode resolver so the v1 checks and their error messages are shared with the combined mode.

Startup failure rolls back everything the attempt initialized: the rollback closes the outermost backend chain, so a v1 failure after RC init no longer leaks the RC service. Teardown runs in reverse startup order - the gateway returns, RC transfers drain, then the v1 stack and the backend close - with the backend wrapped so its shutdown waits for the RC drain, and the v1 completion chain shuts down exactly once even when both the rollback and the normal exit reach it.

The --rdma-cq-depth bound is enforced again: values above the 32-bit range are rejected at startup with an explicit message instead of wrapping to zero.

### DIFF
--- a/cmd/vgwrdma/main.go
+++ b/cmd/vgwrdma/main.go
@@ -16,13 +16,12 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
-	"math"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
-	"strings"
 	"sync"
 
 	"github.com/gofiber/fiber/v3"
@@ -35,6 +34,7 @@ import (
 	"github.com/versity/versitygw/debuglogger"
 	"github.com/versity/versitygw/embedgw"
 	"github.com/versity/versitygw/internal/netutil"
+	"github.com/versity/versitygw/internal/rdmamode"
 	"github.com/versity/versitygw/rdma"
 	"github.com/versity/versitygw/rdma/rcroutes"
 	"github.com/versity/versitygw/rdma/rcserver"
@@ -111,6 +111,7 @@ var (
 	socketPerm                             string
 	rdmaIP                                 string
 	rcGidHint                              string
+	rdmaRCEnable                           bool
 	rdmaPort                               uint
 	poolBufSize                            int
 	poolBufCount                           int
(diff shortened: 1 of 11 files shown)


versity-github:

cmd/vgwrdma/main.go
```diff
@@ -967,27 +979,36 @@ func runGateway(ctx context.Context, be backend.Backend) error {
...
 	}
 
+	// The gateway command owns the input backend from here on:
+	// every later failure path must close the whole chain it has
+	// built so far, exactly once. The closure reads be at run
+	// time, so the rollback covers the v1 chain and the RC
+	// wrapper once those layers are added; the once guard keeps
+	// the shared close with the RunVersityGW lifecycle single.
+	be = rdmamode.WrapShutdownOnce(be)
+	defer func() { be.Shutdown() }()
```
benmcclelland, line 1008: this should be moved to the begining of the function so that the shutdown can be called for any of the above error cases as well.

benmcclelland: we dont need to make this a blocker, can be fixed in a new PR

benmcclelland (approved):

PR #2359, opened 2026-09-04
fix: make conditional publish lock modes explicit

### LINKED ISSUE #2351: [Bug] - flock-based publish lock is node-local on GPFS/Storage Scale, so conditional-write atomicity is silently per-gateway
**Title:** [Bug] - objlock's flock publish lock is silently node-local on several clustered filesystems (measured on GPFS), and no single lock primitive is cluster-coherent everywhere

---
UPDATED - other filesystems besides GPFS

**Describe the bug**

`backend/posix/objlock.go` takes an advisory `flock` on `bucket/.sgwtmp/objlock/<shard>` to make
conditional PUT evaluation and publication atomic per key. This assumes `flock(2)` excludes
holders on *other nodes* when several gateways share one filesystem.

**On IBM GPFS / Storage Scale that assumption does not hold, and it fails silently.** `flock(2)`
succeeds and scopes the lock to the node. Two gateways on different nodes both "hold" the lock for
the same key, and the check-then-act race the lock was added to close remains open.

Silence is the core of the report. The fallback and its warning are reachable only when the lock
call returns an **error**:

```go
// backend/posix/objlock_unix.go:35-41
err := unix.Flock(int(f.Fd()), unix.LOCK_EX|unix.LOCK_NB)
if err == nil {
    return nil                      // <-- GPFS takes this path: success, node-local scope
}
if !errors.Is(err, unix.EWOULDBLOCK) && !errors.Is(err, unix.EAGAIN) &&
    !errors.Is(err, unix.EINTR) {
    return err
}
```

```go
// backend/posix/objlock.go:95-110
err = lockFileExclusive(ctx, f)
if err != nil {
    ...
    // The filesystem does not support advisory locking (e.g. NFS
    // mounted with -o nolock). Fall back to process-local exclusion
    // and warn once: conditional writes are then only atomic within
    // this gateway process.
    p.objLockWarn.Do(func() { ... })
    return mu.Unlock, nil
}
```

GPFS returns no error, so neither the fail-closed path nor the warn-once path runs, and the
gateway proceeds as though it has cross-gateway exclusion. This is the variant of the concern
raised in #2288:

> "I'm a little nervous allowing to fallback to in process locking if the filesystem doesnt support
> flock. Since this will not be able to enforce behavior if multiple gateways are deployed. But I
> guess the warning log might be good enough…"

The warning is good enough for a filesystem that **rejects** `flock`. It never fires for one that
**accepts** `flock` and honours it only locally — which is the larger group.

**This is not GPFS-specific**

GPFS is where I measured it, but the property is common, and it is not always `flock` that is the
node-local primitive. Documentation references at the end.

| Filesystem / mount mode | `flock` cross-node | `fcntl` (`F_SETLK`) cross-node | How it fails here |
|---|---|---|---|
| **GPFS / Storage Scale** | **No** *(measured, below)* | Yes *(measured)* | **Silent** |
| **BeeGFS**, default config | **No** | **No** | **Silent** — both are node-local unless `tuneUseGlobalFileLocks=true` |
| **Lustre** `-o localflock` | **No** | **No** | **Silent** — upstream docs call this mode "potentially unsafe" |
| **Lustre** `-o noflock` (the default before 2.13) | errors `ENOSYS` | — | Detected: fallback + warning |
| **Lustre** `-o flock` (default 2.13+) | Yes | Yes | Safe |
| **GFS2** `-o localflocks` | **No** | **No** | **Silent** — and this mode is *required* when the filesystem is NFS-exported |
| **GFS2** default cluster mode | Yes (DLM) | Yes (DLM) | Safe |
| **OCFS2** with the default `o2cb` stack | **Yes** (since 2.6.26) | **No** — clustered `fcntl` needs a userspace stack (`pcmk`/`cma
| **NFS** `local_lock=flock` / `local_lock=all` / `nolock` | **No** | varies | **Silent** (see next section) |
| **NFS** `local_lock=posix` | Yes (NLM) | **No** | **Inverse of GPFS** |
| **NFS** default (`local_lock=none`) | Yes | Yes | Safe |
| **CephFS** | Yes | Yes | Safe on modern kernels |

**A possible error in the existing doc comment**

`objlock.go:58-59` gives `-o nolock` as the example of a filesystem that "does not support advisory
locking". But `nolock` makes the Linux NFS client perform locking **locally**, and local locks
*succeed* — so `flock()` returns 0, the fallback never runs, and nothing is logged. If that is
right, the one worked example of the detectable case is actually a member of the silent class, and
the comment is reassuring about a configuration it does not actually cover. I have not tested NFS,
so this is offered as a question rather than a claim.

**Scope of what I measured, stated plainly**

I measured the **lock primitive** on the filesystem, from two nodes. I did **not** run a
multi-replica gateway on GPFS and observe two winners on one conditional PUT — that consequence
follows from reading `objlock.go` and is not itself measured. The non-GPFS rows in the table above
are from vendor documentation, **not** measured by me.

**To Reproduce**

Two hosts mounting the same GPFS filesystem. The holder takes an exclusive lock and keeps it; the
prober attempts a non-blocking exclusive lock every 250 ms and counts outcomes. No file data is
read or written, so the result cannot be confused with a cache-coherence effect.

```python
# lockprobe.py  —  usage: lockprobe.py holder|prober <path> <seconds> flock|fcntl
import errno, fcntl, os, sys, time
role, path, secs, kind = sys.argv[1], sys.argv[2], float(sys.argv[3]), sys.argv[4]
def acq(fd, block):
    op = fcntl.LOCK_EX if block else fcntl.LOCK_EX | fcntl.LOCK_NB
    fcntl.flock(fd, op) if kind == "flock" else fcntl.lockf(fd, op, 0, 0, 0)
def rel(fd):
    fcntl.flock(fd, fcntl.LOCK_UN) if kind == "flock" else fcntl.lockf(fd, fcntl.LOCK_UN, 0, 0, 0)
open(path, "a").close()
f = open(path, "r+b")
if role == "holder":
    acq(f.fileno(), True); print("held", flush=True); time.sleep(secs); rel(f.fileno())
else:
    time.sleep(5)                      # let the holder win first
    got = blocked = 0
    end = time.time() + secs
    while time.time() < end:
        try:
            acq(f.fileno(), False); got += 1; rel(f.fileno())
        except OSError as e:
            if e.errno in (errno.EWOULDBLOCK, errno.EAGAIN, errno.EACCES): blocked += 1
            else: raise
        time.sleep(0.25)
    print(f"{kind}: acquired_while_held={got} blocked={blocked}", flush=True)
```

```
# node A                                  # node B
lockprobe.py holder /gpfs/x/lk 120 flock   lockprobe.py prober /gpfs/x/lk 40 flock
lockprobe.py holder /gpfs/y/lk 120 fcntl   lockprobe.py prober /gpfs/y/lk 40 fcntl
```

Make the holder outlast the prober, or the prober's tail attempts land after release and look like
lock failures.

**Results** — same filesystem, same run, holder and prober on different nodes:

| primitive | acquired while held | blocked | verdict |
|---|---|---|---|
| `flock(2)` — what `objlock.go` uses | **160** | **0** | no cross-node exclusion |
| `fcntl` record locks (`F_SETLK`) | **0** | **160** | exclusion holds |

A separate read-modify-write probe under `flock` from the two nodes agrees: 10,475 guarded
increments left a shared counter at ~25, with directly observed overlapping critical sections.

**Expected behavior**

An advisory lock taken on the shared filesystem should exclude publication across all gateways
sharing it — or, where it cannot, the gateway should say so instead of assuming it worked.

**Suggested fix**

~~Use `fcntl` / `F_SETLK` record locks instead of `flock`.~~ **Retracted.** It fixes GPFS and
breaks others: clustered `flock` on OCFS2 works with every cluster stack including the default
`o2cb`, while clustered `fcntl` requires a *userspace* stack and is unsupported on `o2cb`. NFS
`local_lock=posix` inverts the same way, and BeeGFS defaults both primitives to node-local so
neither choice helps there. **There is no primitive that is cluster-coherent everywhere**, so
swapping one for the other just moves the silent failure to a different filesystem.

That leaves the real defect: **the gateway cannot distinguish "I hold a cluster-wide lock" from
"I hold a node-local lock", and currently assumes the former.** Options, roughly in increasing
cost:

1. **Document a hard requirement** and stop implying the warning covers it — cheapest, and worth
   doing regardless of the rest. Say that conditional writes with more than one gateway require a
   filesystem whose advisory locks are cluster-coherent, name the known-unsafe modes
   (`localflock`, `localflocks`, `local_lock=`, BeeGFS default, GPFS `flock`), and correct the
   `-o nolock` example.
2. **Make the primitive selectable** — e.g. `--objlock-mode=flock|fcntl|none`. Cheap, and lets an
   operator pick what their filesystem actually supports. `none` should refuse conditional writes
   rather than silently degrade when more than one gateway is configured.
3. **Verify coherence at startup rather than inferring it from an error.** A self-check needs a
   second participant, so a single process cannot fully test it — but a *fork-based* probe would
   catch nothing GPFS does wrong, so this only works as a documented operator-run test, or via a
   short-lived helper the gateway runs at startup on the shared filesystem while other replicas
   are up. Honestly, this is where the cost is.
4. **Fail closed for multi-gateway deployments** unless the operator explicitly asserts a
   coherent filesystem — an `--i-know-my-locks-are-cluster-coherent`-shaped flag. Ugly, but it
   converts silent data loss into a deployment-time decision, which is the direction #2288 was
   already leaning.

If `fcntl` is adopted anyway, one implementation note: `fcntl` locks are owned per **process** and
are dropped when *any* descriptor on the file is closed by that process, unlike `flock`, which is
tied to the open file description. The current code holds one fd for the lock's lifetime so this
should be compatible, but the paths that may open the same lock file twice deserve a check.

**Server Version**

Analysis is against `main` at `e9354400`. `objlock.go` arrives in `f7e13d71` (with a follow-up in
`fce9d9bf`) and is not in any released tag — latest is `v1.7.0`. The measurement is of the lock
primitive on a GPFS filesystem rather than of a running gateway binary, so there is no
`./versitygw -version` output to give.

**Additional context**

The deployment shape that motivated this: four stateless gateway replicas on Kubernetes sharing
one GPFS filesystem via a `ReadWriteMany` volume — the "multiple gateways" case #2288's discussion
flagged. Also relevant: #2321 (hardening), #2183 and #1522 (conditional write support), and #701
(multi-gateway testing).

Documentation for the non-GPFS rows:

- Lustre — [mount options](https://wiki.lustre.org/Mounting_a_Lustre_File_System_on_Client_Nodes),
  [mount.lustre(8)](https://github.com/Xyratex/lustre-stable/blob/master/lustre/doc/mount.lustre.8)
- GFS2 — [Red Hat: mounting conditions for gfs2](https://access.redhat.com/articles/3244841),
  [gfs2_mount(8)](https://linux.die.net/man/8/gfs2_mount),
  [Red Hat: how POSIX fcntl locks work on GFS2](https://access.redhat.com/articles/48659)
- BeeGFS — [POSIX file locking / `tuneUseGlobalFileLocks`](https://groups.google.com/g/fhgfs-user/c/7XY9qwepZFk)
- OCFS2 — [ocfs2(7)](https://manpages.ubuntu.com/manpages/focal/man7/ocfs2.7.html),
  [user file locks design doc](https://oss.oracle.com/osswiki/OCFS2/DesignDocs/UserFileLocks.html)
- NFS — [nfs(5)](https://linux.die.net/man/5/nfs),
  [cluster-coherent NFS and byte-range locking](http://wiki.linux-nfs.org/wiki/index.php/Cluster_Coherent_NFS_and_Byte_Range_Locking)

benmcclelland:
Conditional PUTs require a lock primitive that excludes competing gateway processes sharing a backend filesystem. A successful flock call does not prove that property: some clustered filesystem configurations accept flock but scope it to one node, silently leaving cross-gateway check-and-publish races open.

Add an object-lock mode that lets operators select flock or fcntl for filesystems where that primitive is cluster-coherent, local for the existing per-process behavior, or none to reject conditional writes with NotImplemented. Keep the legacy disable flag as an alias for local.

Shared lock modes now verify the selected primitive on the root lock filesystem during startup and fail closed if it cannot be used. Runtime lock failures no longer silently downgrade to process-local exclusion. The startup check cannot establish cross-node coherence, so that remains an explicit operator requirement.

Fixes #2351

### DIFF
--- a/backend/posix/objlock.go
+++ b/backend/posix/objlock.go
@@ -23,7 +23,6 @@ import (
 	"time"
 
 	"github.com/versity/versitygw/backend"
-	"github.com/versity/versitygw/debuglogger"
 )
 
 // Object publish locking
@@ -33,8 +32,8 @@ import (
 // replacement happen as one atomic step per bucket/key. The gateway is
 // stateless and multiple gateway processes may share the same backend
 // filesystem, so an in-process mutex alone is not sufficient: exclusion is
-// provided by an advisory lock (flock on unix, LockFileEx on windows) on a
-// shared lock file, combined with a process-local striped mutex so that
+// provided by a configured advisory lock (flock or fcntl on unix, LockFileEx
+// on windows) on a shared lock file, combined with a process-local striped mutex so that
 // contention within one process is resolved cheaply and each process presents
 // at most one waiter to the filesystem lock.
 //
@@ -56,13 +55,11 @@ import (
 // the file handle is closed or the process exits, so failures, cancellation,
 // or crashes cannot leave a permanently stale lock.
 //
-// NFS notes: flock on Linux NFS clients is mapped to NFSv4 byte-range locks
-// (or NLM on NFSv3), giving cross-client exclusion. Mounting with
-// "-o nolock" or "-o local_lock=flock"/"local_lock=all" disables server-side
-// locking and reduces exclusion to a single client; conditional-write
-// atomicity across gateways requires server-backed locking. If the filesystem
-// does not support advisory locking at all, the gateway falls back to
-// process-local exclusion and logs a warning once.
+// Filesystem lock scope varies by filesystem and mount options. Operators using
+// multiple gateway processes must select a mode that their filesystem provides
+// with cluster-wide exclusion. The local mode provides only process-local
+// exclusion and is unsafe for conditional writes across gateway processes. The
+// none mode rejects conditional writes.
 
 const (
 	// objLockDir is the root directory holding object publish lock files.
@@ -70,6 +67,63 @@ const (
 	// objLockShards is the number of lock shards per bucket
 	objLockShards = 256
 )
+
+// ObjectLockMode selects the advisory locking primitive used to serialize
+// conditional object publishes.
+type ObjectLockMode string
+
+const (
+	ObjectLockModeFlock ObjectLockMode = "flock"
+	ObjectLockModeFcntl ObjectLockMode = "fcntl"
+	ObjectLockModeLocal ObjectLockMode = "local"
+	ObjectLockModeNone  ObjectLockMode = "none"
+)
+
+func resolveObjectLockMode(mode ObjectLockMode, forceNoObjLockFile bool) (ObjectLockMode, error) {
+	if mode == "" {
+		if forceNoObjLockFile {
+			return ObjectLockModeLocal, nil
+		}
+		return ObjectLockModeFlock, nil
+	}
+	if forceNoObjLockFile && mode != ObjectLockModeLocal {
+		return "", fmt.Errorf("disable object lock file conflicts with object lock mode %q", mode)
+	}
+	switch mode {
+	case ObjectLockModeFlock, ObjectLockModeFcntl, ObjectLockModeLocal, ObjectLockModeNone:
+		return mode, nil
+	default:
+		return "", fmt.Errorf("invalid object lock mode %q (want flock, fcntl, local, or none)", mode)
+	}
+}
+
+// verifyObjectLockMode verifies that the configured shared lock primitive is
+// available on the filesystem containing the gateway's lock directory. It
+// cannot verify cross-node lock coherence.
+func (p *Posix) verifyObjectLockMode() error {
+	if p.objectLockMode == ObjectLockModeLocal || p.objectLockMode == ObjectLockModeNone {
+		return nil
+	}
+	if err := validateObjectLockMode(p.objectLockMode); err != nil {
+		return err
+	}
+
+	lockDir := filepath.Join(p.rootdir, objLockDir)
+	if err := backend.MkdirAll(lockDir, 0, 0, false, p.newDirPerm); err != nil {
+		return fmt.Errorf("make object lock directory: %w", err)
+	}
+	f, err := os.CreateTemp(lockDir, ".startup-lock-*")
+	if err != nil {
+		return fmt.Errorf("create object lock probe: %w", err)
+	}
+	defer os.Remove(f.Name())
+	defer f.Close()
+
+	if err := lockFileExclusive(context.Background(), f, p.objectLockMode); err != nil {
+		return fmt.Errorf("verify object lock mode %q: %w", p.objectLockMode, err)
+	}
+	return nil
+}
 
 // objLockShard returns the shard index for an object key.
 func objLockShard(object string) uint8 {
@@ -96,7 +150,7 @@ func (p *Posix) lockObjectPublish(ctx context.Context, bucket, object string) (f
 		releaseLocal()
 		return nil, err
 	}
-	if p.forceNoObjLockFile {
+	if p.objectLockMode == ObjectLockModeLocal || p.objectLockMode == ObjectLockModeNone {
 		return releaseLocal, nil
 	}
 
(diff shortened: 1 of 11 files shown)


niksis02 (commented):
ScoutFS backend still hardcodes local-only locking and wasn't migrated to `ObjectLockMode`. It silently stays on process-local exclusion with no startup warning and no way to opt into fcntl. Worth extending this fix to ScoutFS ?

benmcclelland:
> ScoutFS backend still hardcodes local-only locking and wasn't migrated to `ObjectLockMode`. It silently stays on process-local exclusion with no startup warning and no way to opt into fcntl. Worth extending this fix to ScoutFS ?

scoutfs doesnt have cluster consistent posix locks. but now i'm thinking that maybe we should default to none and allow local-only for single node systems if people need this.

niksis02 (approved):

PR #2360, opened 2026-09-05
rdma: install the gateway shutdown wrapper at function entry

potatogim:
Moves the shutdown-once wrapper installation from after the gateway option validation to the top of runGateway, so every early error return (copy-object-threshold, RDMA mode validation) closes the backend chain exactly once instead of leaking it, as discussed in #2357.


versity-github:

benmcclelland (approved):

Issue #2361, opened 2026-09-05
[Feature] - Operational controls for the RC data plane (tuning flags, diagnostics, session visibility)


potatogim:
**Describe the solution you'd like**

I'd like to propose operational controls for the RC data plane introduced in #2330. As it stands, the RC plane runs with its resource limits hardcoded at the `rcserver.Init` call site, emits no diagnostics beyond stderr errors, and keeps its session table invisible to operators, while the v1 cuObject stack offers tuning flags, telemetry hooks, and wire-status visibility. I believe an operator sizing vgwrdma for RC traffic currently has no knob to turn and nothing to look at when a transfer stalls.

The work is delivered as a four-PR stack so each part can be reviewed on its own:

1. Tuning flags (#2362): one gateway flag per RC resource limit (sessions, per-principal sessions, staging budgets, queue pairs, READY admission slots, prepare/exec timeouts), defaulting to today's hardcoded values, validated through `rdmamode` only when the RC data plane is enabled so stale environment values never block a v1-only or plain-S3 startup.

2. Diagnostics (#2363): a log-callback ABI so the RC core reports session and transfer lifecycle events beyond the stderr error lines, wired to the existing `--debug` switch. The sink runs outside session locks, copies messages before returning, and falls back to stderr when the sink would block.

3. Session visibility (#2365): a snapshot ABI over the session table plus an admin endpoint listing active RC sessions - id, principal, state, age, and pending-transfer summary - so a stalled gateway can be inspected without a debugger.

4. S3 service parity (#2366): RC completions flow through the same audit, metrics, and event pipeline as native GET/PUT, with an operational-context adapter that carries the bucket/key and region from the PREPARE request into the completion path.

**How the parts fit together**

```mermaid
flowchart LR
    client["S3 client (RDMA)"]
    backend["object backend<br/>(posix / s3 proxy)"]

    subgraph gw["vgwrdma gateway process"]
        auth["SigV4 auth + IAM (existing)"]
        routes["RC routes (existing)<br/>prepare / ready / cancel"]
        rcsvc["rcserver.RCSvc<br/>Go bindings"]
        flags["gateway flags via rdmamode"]
        admin["admin server<br/>GET /rc-sessions"]
        dbg["debug logger<br/>--debug"]
        ops["opsTracker<br/>single publication worker"]
        credit["session admission<br/>credit held until published"]
        audit["audit log"]
        met["metrics"]
        evt["event delivery"]

        subgraph cd["C data plane"]
[...]

PR #2362, opened 2026-09-05
rdma: expose the RC data plane resource limits as gateway flags

potatogim:
Part of #2361.

The `hipobj-rc-v2` data plane currently starts with its session, queue pair, staging and timeout limits hardcoded at the `rcserver.Init` call site, so as far as I can tell there is no way for an operator to size the RC plane for their hardware the way they can for the cuObject backend. This adds one flag per limit plus the READY admission slot count, all defaulting to the values the gateway passes today: `--rdma-rc-max-sessions`, `--rdma-rc-max-user-sessions`, `--rdma-rc-max-staging-bytes`, `--rdma-rc-max-user-staging-bytes`, `--rdma-rc-max-qps`, `--rdma-rc-max-user-qps`, `--rdma-rc-max-ready-slots`, `--rdma-rc-prep-timeout-ms` and `--rdma-rc-exec-timeout-ms`, each with a `VGW_RDMA_RC_*` environment counterpart.

The values are validated through a new `rdmamode.V2ValidationError` consulted only when `--rdma-rc-enable` is set, mirroring the stale-value handling of the v1 settings: counts are parsed as `uint64` and range-checked against the `uint32` narrowing at the `DeviceOpts` boundary, and the timeouts carry an upper bound that keeps the `nowMs + timeout` deadline arithmetic in the C core from wrapping. Deployments that set nothing keep the previous behavior unchanged, so I believe the change is safe to pick up as-is.

I exercised the flags on the RC test host while driving the control plane end to end (see #2363 for that work), including short `--rdma-rc-prep-timeout-ms` values used to observe session reaping.

### DIFF
--- a/cmd/vgwrdma/main.go
+++ b/cmd/vgwrdma/main.go
@@ -120,6 +120,15 @@ var (
 	rdmaCQDepth                            uint
 	rdmaRetryCount                         uint
 	rdmaTunablesSet                        bool
+	rcMaxSessions                          uint64
+	rcMaxUserSessions                      uint64
+	rcMaxStagingBytes                      uint64
+	rcMaxUserStagingBytes                  uint64
+	rcMaxQPs                               uint64
+	rcMaxUserQPs                           uint64
+	rcMaxReadySlots                        uint64
+	rcPrepTimeoutMs                        uint64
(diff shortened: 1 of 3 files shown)


versity-github:

PR #2363, opened 2026-09-05
rdma: add a log callback ABI to the RC data plane

potatogim:
Part of #2361.

Adds a log callback ABI to the RC data plane so gateway operators can see session-level diagnostics (reap events, READY data-phase outcomes, init failures) instead of only having raw stderr.

`rc_server_set_log_sink()` installs a plain function pointer after init that stays valid until destroy. The gateway registers a fixed cgo trampoline (closures cannot cross the boundary), copies each message immediately per the lifetime contract, and never calls the sink while holding the session map lock. Error-level lines keep their existing stderr output, and `--debug` enables the level-2 diagnostic stream.

While verifying the init failure path I noticed a loader bug on rdma-core 61+ hosts: as far as I can tell, `ibv_poll_cq`, `ibv_post_send`, and `ibv_post_recv` are no longer exported as library symbols in modern rdma-core. verbs.h ships them as static inline wrappers that dispatch through `cq->context->ops`, so the dlsym-based loader treated the installed library as incomplete and init failed with a bare `RC_E_INTERNAL`. The loader now resolves those three calls through the ops table of a briefly opened device, and deviceless and gid-hint-mismatch init failures now print which step failed.

While verifying the sink end to end I ran into two more defects, both fixed here. First, the `Prepare` and `ReadyTransfer` Go wrappers embedded string views built from heap strings inside request structs passed to C by pointer, so every live PREPARE or READY call with header-derived strings panicked at the cgo pointer-check boundary and the route returned a 500; the string bytes are now pinned with `runtime.Pinner` for the duration of the call. Second, the session id only lived as the sessions map key, so the reap record logged an empty id; the record now carries its own copy. A deviceless cgo regression test covers the panic path by calling the real wrapper with heap-backed strings and an opcode that returns from C argument validation before the server handle is touched.

Verified on a host with rdma-core 61 and a passthrough mlx5_0 VF: archive rebuild, full vgwrdma link, stub and unit tests, the new cgo boundary test under `GODEBUG=cgocheck=1` and `GOEXPERIMENT=cgocheck2`, `--debug` startup reaching a listening gateway, a signed `POST /.hipobj-rc/prepare` returning a complete session (id, QPN, PSN, staging address and rkey) instead of the previous 500, reap records reaching the gateway log through the callback with the session id present, and clean SIGTERM shutdown.

### DIFF
--- a/cmd/vgwrdma/main.go
+++ b/cmd/vgwrdma/main.go
@@ -120,6 +120,15 @@ var (
 	rdmaCQDepth                            uint
 	rdmaRetryCount                         uint
 	rdmaTunablesSet                        bool
+	rcMaxSessions                          uint64
+	rcMaxUserSessions                      uint64
+	rcMaxStagingBytes                      uint64
+	rcMaxUserStagingBytes                  uint64
+	rcMaxQPs                               uint64
+	rcMaxUserQPs                           uint64
+	rcMaxReadySlots                        uint64
+	rcPrepTimeoutMs                        uint64
+	rcExecTimeoutMs                        uint64
 )
 
 var (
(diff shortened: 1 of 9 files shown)


versity-github:

potatogim:
The MacOS s3cmd check failure looks unrelated to this change - the log stops at
`error installing go dependencies: go: -d flag is deprecated. -d=true is a no-op`
before reaching the build. That reads like the workflow's go install invocation hit the
deprecation on a newer Go, not something this diff touches. I may well be missing
something. Whenever convenient, could someone take a look or retry?

PR #2365, opened 2026-09-06
rdma: expose live RC sessions on the admin server

potatogim:
Part of #2361.
Builds on #2363.

Adds live RC session observability: `GET /rc-sessions` on the admin server returns a point-in-time JSON snapshot of every session in the RDMA control-plane map.

The C ABI gains `rc_server_sessions_snapshot()`, which copies each live session into a fixed `rc_session_snapshot` record under the map lock and invokes the callback once per record outside the lock, so the callback cannot race the reaper moving or erasing entries. Each record carries the session id, op, target, state machine value combined with a reap-pending marker (CANCEL and expiry only set that marker, they do not move the state), age in milliseconds, and staged bytes. Age comes from a new monotonic creation timestamp on the session, because the prepare/ready deadlines move as the session progresses and cannot serve as an age reference. Records whose op or target does not fit the fixed fields are skipped rather than truncated.

On the Go side an `RCSvc.SessionsSnapshot()` wrapper marshals the records through the same fixed-trampoline pattern as the log sink, and the admin server gains a `WithAdminRoute` option so an embedding binary can register extra admin routes. `vgwrdma` registers the snapshot route when `--rdma-rc-enable` is set, running it behind the same signature verification and admin checks as the built-in admin endpoints, so unsigned or non-admin requests get the usual 403 XML error rather than a generic 500. Stub builds return a not supported error, keeping the build matrix unchanged.

Verified on a host with rdma-core 61 and a passthrough mlx5_0 VF: archive rebuild, full vgwrdma link, stub and unit tests, and an end-to-end run where a signed `POST /.hipobj-rc/prepare` creates a session and the snapshot endpoint immediately returns it (id, `op=GET`, target, `state=prepared`, staged bytes), while the same request without a signature is rejected with 403.

### DIFF
--- a/cmd/vgwrdma/main.go
+++ b/cmd/vgwrdma/main.go
@@ -120,6 +120,15 @@ var (
 	rdmaCQDepth                            uint
 	rdmaRetryCount                         uint
 	rdmaTunablesSet                        bool
+	rcMaxSessions                          uint64
+	rcMaxUserSessions                      uint64
+	rcMaxStagingBytes                      uint64
+	rcMaxUserStagingBytes                  uint64
+	rcMaxQPs                               uint64
+	rcMaxUserQPs                           uint64
+	rcMaxReadySlots                        uint64
+	rcPrepTimeoutMs                        uint64
+	rcExecTimeoutMs                        uint64
 )
 
 var (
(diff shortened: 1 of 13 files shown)


versity-github:

benmcclelland (approved):

PR #2366, opened 2026-09-06
rdma: publish RC transfer outcomes into the gateway operational services

### LINKED ISSUE #2361: [Feature] - Operational controls for the RC data plane (tuning flags, diagnostics, session visibility)

potatogim:
Fixes #2361.
Builds on #2363.
Builds on #2365.

Publishes RC transfer outcomes into the gateway's operational services, so RDMA sessions appear in the audit log, metrics, and event delivery: every PREPARE, READY, CANCEL, and completed transfer leaves the same structured record a native GET/PUT would, with the real error classification, bytes moved, and object-created events for committed PUTs.

Publication ownership follows outcome ownership. Teardown callbacks own the records of sessions that became unowned (expired, abandoned, or cancelled) and the request path owns its own outcomes; a reservation token closes the window between a synchronous callback and the in-flight request that has not reported yet, so outcomes land exactly once per session and match the wire response the client saw.

The teardown callback never runs sink work on a native thread. A single worker detaches the queue under the publication lock and performs all audit, metrics, and event sinks outside it, so a stalled downstream blocks the callback only for as long as it takes to hand off the record. Backpressure is explicit: session admission holds a credit until publication completes and bounds the outstanding backlog by the native session quota (excess registrations get a SlowDown response), and session-less records such as failed authentications are capped with a drop counter reported at shutdown rather than growing without bound.

Two changes sit outside the rdma tree. The metrics interface gains SendWithBucket because RC publications run on a synthetic fiber context without a matched route, so the bucket dimension must be passed explicitly rather than read from Params. While wiring the shutdown ordering I also ran into two latent defects in the metrics manager close path (a send on a closed channel from a late handler, and Close waiting on a context it did not own) and fixed them here. I can split these into a separate PR if that reads better.

Verified on a development host with an mlx5_0 VF passthrough (RoCE not available there, so the data-path probes run on the emulated transport used for RC bring-up): archive rebuilt from source, full link, stub and unit suites with the race detector, and end-to-end runs covering expired-session publication, one-shot auth-failure publication, and one-shot signature-failure publication, plus repeated stress tests of the publication lifecycle - concurrent admissions against a small session quota, overflow drain, stalled-sink recovery, and a test hook that pauses a registration between its admission check and the record insert so a concurrent registration races it.

### DIFF
--- a/cmd/vgwrdma/main.go
+++ b/cmd/vgwrdma/main.go
@@ -1287,13 +1287,19 @@ func runGateway(ctx context.Context, be backend.Backend) error {
 		// nil on success without doing so. Wrap it so a verified
 		// request reaches the route handler, while errors end the
 		// chain as usual.
+		rcH := rcroutes.New(rcSvc, be, iamSvc, readonly, disableACLs)
 		rcAuth := func(ctx fiber.Ctx) error {
+			// The RC routes run before the default-values
+			// middleware sets the request locals; the access
+			// logger and event schema read the region from
+			// there, so set it for every verified request.
+			utils.ContextKeyRegion.Set(ctx, region)
 			if err := rcVerify(ctx); err != nil {
+				rcH.PublishAuthFailure(ctx, err)
 				return rcroutes.WriteRouteError(ctx, err)
 			}
 			return ctx.Next()
 		}
-		rcH := rcroutes.New(rcSvc, be, iamSvc, readonly, disableACLs)
 		// The gateway builds the access logger, metrics manager,
 		// and event sender inside RunVersityGW; hand them to the
 		// RC routes as soon as they exist so finished transfers
--- a/cmd/vgwrdma/main.go
+++ b/cmd/vgwrdma/main.go
@@ -1276,7 +1276,6 @@ func runGateway(ctx context.Context, be backend.Backend) error {
 		// rollback closure and the RunVersityGW lifecycle a
 		// single, ordered owner of both steps.
 		be = rdmamode.WrapBackendShutdownAfterRC(be, rcSvc)
-
 		rcVerify := middlewares.VerifyV4Signature(
 			middlewares.RootUserConfig{
 				Access: gwcli.RootUserAccess,
@@ -1288,6 +1287,13 @@ func runGateway(ctx context.Context, be backend.Backend) error {
 		// request reaches the route handler, while errors end the
 		// chain as usual.
 		rcH := rcroutes.New(rcSvc, be, iamSvc, readonly, disableACLs)
+		// The RC shutdown wrapper drains queued operational
+		// publications before the sinks close; the drain hook is
+		// the route handler, which is built only now (it needs
+		// the wrapped backend).
+		if w, ok := be.(*rdmamode.BackendShutdownAfterRC); ok {
+			w.SetOpsDrainer(rcH)
+		}
 		rcAuth := func(ctx fiber.Ctx) error {
 			// The RC routes run before the default-values
 			// middleware sets the request locals; the access
--- a/cmd/vgwrdma/main.go
+++ b/cmd/vgwrdma/main.go
@@ -1286,7 +1286,7 @@ func runGateway(ctx context.Context, be backend.Backend) error {
 		// nil on success without doing so. Wrap it so a verified
 		// request reaches the route handler, while errors end the
 		// chain as usual.
-		rcH := rcroutes.New(rcSvc, be, iamSvc, readonly, disableACLs)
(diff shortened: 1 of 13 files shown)


versity-github:

benmcclelland (approved):

PR #2368, opened 2026-09-07
feat: add s3:if-match and s3:if-none-match policy condition keys

### LINKED ISSUE #2294: [Feature] - Support s3:if-none-match in bucket-policy conditions

I'd like to configure a write-only / blind-drop account that may create new objects but cannot overwrite existing ones.

VersityGW already supports `If-None-Match: *` on `PutObject`. I tried enforcing use of that header with this bucket policy:

```json
{
  "Version": "2012-10-17",
  "Statement": [{
    "Sid": "AllowCreateOnly",
    "Effect": "Allow",
    "Principal": {
      "AWS": "jweb"
    },
    "Action": "s3:PutObject",
    "Resource": "arn:aws:s3:::testasuser/*",
    "Condition": {
      "Null": {
        "s3:if-none-match": "false"
      }
    }
  }]
}
```

The policy is accepted, but the condition is currently ignored.

**Additional context**

Looking at the code, `BucketPolicyItem` does not currently have a `Condition` field, while `PutObject` already parses `If-None-Match` and passes it to the backend.

Would you consider adding narrow support for the `Null` / `s3:if-none-match` condition so the existing request-header behavior can be enforced by a bucket policy?

This would enable simple create-only / blind-drop archival workflows without adding a new VersityGW-specific permission mode.

niksis02:
Closes #2294

Support AWS's conditional-write condition keys in both S3 bucket policies and IAM identity policies, so a policy can require that an upload or delete carry an `If-Match`/`If-None-Match` precondition.

`requestConditionContext` now populates both keys from the request headers, which covers both policy types at once: the same map is evaluated in-process for bucket policies and shipped to the IAM service for identity policies. Only the three requests whose preconditions the gateway actually enforces contribute - PutObject, CompleteMultipartUpload and DeleteObject. Copies, form uploads, DeleteObjects batches, upload parts and the sub-resource writes all ignore the headers, and a policy must never grant on a precondition that won't be checked; reads are excluded for the same reason, since GET and HEAD take these headers as ordinary HTTP cache preconditions. The value is the ETag with its surrounding quotes stripped, matching what the gateway enforces against.

Bucket policies validate condition keys against a fixed catalogue at `PutBucketPolicy` time, so both keys are added there with the action sets AWS accepts: `s3:if-match` on `s3:PutObject` and `s3:DeleteObject`, `s3:if-none-match` on `s3:PutObject` alone. Identity policies validate only the operator vocabulary, matching AWS, so they need no change.

### DIFF
--- a/auth/condition_context.go
+++ b/auth/condition_context.go
@@ -15,6 +15,7 @@
 package auth
 
 import (
+	"slices"
 	"strconv"
 	"time"
 
@@ -28,7 +29,11 @@ import (
 // aws:username, aws:PrincipalTag/*, …) are deliberately absent: the S3
 // gateway has no way to know them, so the IAM service fills them in itself
 // when it evaluates an identity policy.
-func requestConditionContext(ctx fiber.Ctx) map[string][]string {
+//
+// actions is the action set this request is being authorized under, used by
+// the keys whose applicability AWS defines per action. Pass nil where no
+// such key can apply.
+func requestConditionContext(ctx fiber.Ctx, actions []Action) map[string][]string {
 	now := time.Now().UTC()
 	condCtx := map[string][]string{
 		"aws:CurrentTime":     {now.Format(time.RFC3339)},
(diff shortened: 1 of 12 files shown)


auth/condition_context.go
```diff
@@ -63,6 +64,68 @@ func requestConditionContext(ctx fiber.Ctx) map[string][]string {
...
+// CompleteMultipartUpload and DeleteObject.
+//
+// Everything else ignores those headers, and a policy must never grant on a
+// precondition that won't be checked — otherwise a form upload, a
+// DeleteObjects batch, a copy or an upload part could satisfy a statement
+// demanding a conditional write by sending a header that changes nothing.
+// Reads are the same case: GET and HEAD take these headers as ordinary HTTP
+// cache preconditions. Excluding a request leaves both keys absent, which
+// denies it under such a policy rather than letting it through.
+func isConditionalWrite(ctx fiber.Ctx) bool {
```
benmcclelland: every DELETE request returns true unless it has one of the excluded subresource query parameters. That includes requests with versionId.

However, object-delete.go:125 maps a request with versionId to DeleteObjectVersionAction, and the PR’s own bucket-policy validation explicitly rejects s3:if-match for s3:DeleteObjectVersion. The backend still evaluates If-Match for versioned deletes, so an IAM identity policy such as:
```
{
  "Effect": "Allow",
  "Action": "s3:DeleteObjectVersion",
  "Resource": "...",
  "Condition": {
    "StringEquals": {
      "s3:if-match": "etag"
    }
  }
}
```
could be satisfied even though the PR says this key applies only to s3:DeleteObject. This also contradicts the added unit test that expects a versioned delete to populate s3:if-match.

The classifier should exclude versionId deletes, or the action applicability rules and stated behavior should be changed consistently. A focused regression test should verify both bucket-policy and IAM-policy behavior for DeleteObjectVersion.

auth/condition_context.go
```diff
@@ -63,6 +64,95 @@ func requestConditionContext(ctx fiber.Ctx) map[string][]string {
...
+// denies it under such a policy rather than letting it through.
+func conditionalWriteAction(ctx fiber.Ctx) Action {
+	// A copy carries its preconditions in the X-Amz-Copy-Source-If-*
+	// headers, which name the source object and populate neither key.
+	if ctx.Get("X-Amz-Copy-Source") != "" {
+		return ""
+	}
+	query := ctx.Request().URI().QueryArgs()
+	switch string(ctx.Request().Header.Method()) {
+	case fiber.MethodPut:
```
benmcclelland, line 146: derive conditional-key applicability from the actual resolved S3 action, or otherwise restrict it to the exact object-write routes. s3:if-match should be populated only for PutObject, CompleteMultipartUpload (authorized as PutObject), and unversioned DeleteObject; s3:if-none-match should be populated only for PutObject and CompleteMultipartUpload. Keep both keys absent for DeleteObjectVersion, bucket-level writes, copies, upload parts, batch deletes, and reads. Add regression tests for bucket-level PUT operations and versioned deletes.

auth/condition_context.go
```diff
@@ -63,6 +64,95 @@ func requestConditionContext(ctx fiber.Ctx) map[string][]string {
...
+		return ""
+	default:
+		return ""
+	}
+}
+
+// trimETagQuotes strips one leading and one trailing double quote from an
+// ETag-valued header, leaving any other value (notably If-None-Match's "*")
+// untouched.
+func trimETagQuotes(s string) string {
```
benmcclelland, line 179: trimETagQuotes strips quotes only when they surround the entire header value. A valid header such as:
`If-Match: "etag-a", "etag-b"`
becomes one context value containing both ETags rather than two values. This does not match the gateway’s intended conditional semantics, where any matching ETag should satisfy If-Match.

benmcclelland: is it well defined from AWS how multiple entity-tags and * are represented in the condition context? We might need some tests covering these cases.

niksis02: I checked this against live AWS and it does not apply HTTP list semantics to a conditional write. `PutObject` with
`If-Match: "<wrong>", "<right>"` is `412 PreconditionFailed`, and so is the reverse order —it's a whole-string compare against the single current ETag, exactly what `EvaluatePreconditions` does. "Any matching ETag satisfies If-Match" is the read semantics, which applies to `GET/HEAD`, and those don't populate either key.

The condition key matches that. Probed with a bucket-policy Deny on s3:if-match, the value for If-Match: "A", "B" is the single string  A", "B  — the raw header with only the outermost quotes stripped, 68 chars, no comma split and no per-element unquoting. `ForAnyValue:StringEquals` and `ForAllValues:StringEquals` naming either ETag alone do not match (both operators do work on this key when the header is a single ETag), so it is not
multivalued. `StringLike` with `A + four ? + B` matches as a full-string pattern, which pins both the length and the absence of outer quotes.

Splitting would therefore be wrong twice over: it wouldn't match AWS's representation, and it would let a policy grant on a header shape the write itself rejects with 412 — the same "grant on a precondition that isn't enforced" hazard the classifier avoids elsewhere.

One deliberate difference worth naming: AWS strips a leading quote and a trailing quote independently, so `"abc` (unbalanced) becomes abc there while `trimETagQuotes` leaves it as `"abc`. That's on purpose — the context value has to equal what `utils.TrimQuotes` hands `EvaluatePreconditions`, or a policy could match a value the precondition check doesn't. Both forms fail the precondition anyway.

benmcclelland (approved):

PR #2372, opened 2026-09-08
feat: add storage class to put object input

benmcclelland:
A new glaicer mode for the archiving backend needs the storage class supplied to the back for put object input.


niksis02 (commented):
Shouldn't we add `x-amz-storage-class` for browser based object `POST` and `CreateMultipartUpload`?

benmcclelland:
> Shouldn't we add `x-amz-storage-class` for browser based object `POST` and `CreateMultipartUpload`?

I think its already there for CreateMultipartUpload, I'll check the POST.

niksis02 (approved):

PR #2373, opened 2026-09-08
feat: add OIDC endpoint relaxations for private/isolated networks

### LINKED ISSUE #2364: [Feature] - disable ssrfSafeHTTPClient
**problem description**
I started to look at wiring up the new oidc support.

> The blocker: versitygw's standalone IAM service fetches the OIDC discovery doc + JWKS through an SSRF-guarded HTTP client that unconditionally rejects loopback, RFC1918 and link-local targets ([isDisallowedFetchTarget](https://github.com/versity/versitygw/blob/v1.8.0/iamapi/internal/iamutil/oidc_thumbprint.go#L136), used by webidentity.go:816). There is no flag to disable it — versitygw's own test suite says a real AssumeRoleWithWebIdentity is "unreachable from this suite by design". A SPIRE OIDC discovery provider on a normal ClusterIP (10.96/12) will always be refused. It also requires https with no explicit port and does real chain verification (no skip-verify).

This design is problematic for running on local and isolated networks where private ip's are normal. You can run both the versitygw and the spiffe-oidc-discovery-provider in the same kubernetes cluster and use its private ips safely.

Or, you should be able to run the spire-oidc-discovery-provider as a sidecar in the same pod as versitygw and bind it to 127.0.0.1 and have it completely unspoofable so even http would be reasonable for this configuration.

Its also normal and good to run such services on non root only ports such as 8443 to allow the process to run as non-root.

Allow some configuration flags to reduce the strict checking. Its too strict in some circumstances.

niksis02:
Closes #2364

`AssumeRoleWithWebIdentity` only ever trusted an `OIDC` provider reachable over verified `https`, at a publicly routable address, on the implicit `:443`. That posture is right for an internet-facing IdP but rejects every address an internal one can have, so a `SPIFFE/SPIRE` OIDC discovery provider in the same cluster — or as a sidecar in the same pod — could never be registered, let alone verified against, and no setting could express "this private address is the IdP".

Two opt-in flags on `versitygw iam`, both off by default:

`--oidc-allow-private-endpoints`
Permit a provider `Url` resolving to a loopback/private/link-local address, and an explicit port. Transport is unchanged: still `https`, still fully verified (a self-signed in-cluster cert is trusted the way AWS documents, through `ThumbprintList`).

`--oidc-allow-insecure-transport`
Additionally permit plaintext `http` provider URLs, discovery/JWKS endpoints and redirects, and drop TLS verification (`thumbprint` pinning included) for `https` ones.

Both apply uniformly to the thumbprint auto-fetch at `CreateOpenIDConnectProvider` time and to the discovery-document plus `JWKS` fetch at `AssumeRoleWithWebIdentity` time. Neither weakens anything past the endpoint: signature verification, issuer matching, audience and trust policy evaluation are untouched, and the DNS-resolve-once/dial-the-resolved-IP shape stays in place so a rebind still cannot redirect a connection.

An `http` provider keeps its scheme in its stored `Url`, `ARN` and `iss` matching, rather than being stripped like an `https` one — otherwise `"http://host"` and `"https://host"` would collapse onto a single ARN and storage key and each could satisfy the other's trust policy. It also stores an empty `ThumbprintList` rather than failing: a plaintext provider presents no certificate to thumbprint.

Helm: `iamServer.oidc.{allowPrivateEndpoints,allowInsecureTransport}`, alongside `disableThumbprintAutofetch` moved into the same block (the flat `iamServer.disableOidcThumbprintAutofetch` stays honored). Chart `0.4.1 -> 0.4.2`.

The WebUI's create-provider form no longer rejects `http` URLs and ports client-side; it cannot see the service's settings, so those two rules are left to the server, whose error surfaces as a toast like any other.


kfox1111:
This is looking good. One more thing that could be helpful in this context...

Somethings like Kubernetes own implementation allow specifying a discoveryURL separate from the issuerURL. When the discoveryURL is set, it is used for the key lookup, while issuerURL is still used as the validation url against the issuer in the JWT, as the spec requires. This enables looking up the keys through an optimized private path when in cluster, but the JWT tokens can still be handed out to services on the internet and they can still look up the token validity based on the issuerURL only, as the JWT spec requires.

benmcclelland (approved):

benmcclelland:
> This is looking good. One more thing that could be helpful in this context...
> 
> Somethings like Kubernetes own implementation allow specifying a discoveryURL separate from the issuerURL. When the discoveryURL is set, it is used for the key lookup, while issuerURL is still used as the validation url against the issuer in the JWT, as the spec requires. This enables looking up the keys through an optimized private path when in cluster, but the JWT tokens can still be handed out to services on the internet and they can still look up the token validity based on the issuerURL only, as the JWT spec requires.

I agree this is a useful distinction, but it expands the provider model beyond the endpoint relaxations in this PR. I’d prefer to land the current scoped change and address a separate DiscoveryURL field in a follow-up PR, preserving Url for issuer and trust-policy validation while using DiscoveryURL only for discovery/JWKS retrieval.

We can follow up with the DiscoveryURL next.

kfox1111:
Would you like me to file an issue for the followup?

niksis02:
> Would you like me to file an issue for the followup?

If you've found any issue, feel free to open a bug report issue, I'll take a look.

PR #2374, opened 2026-09-08
s3api: report the bound addresses from the listen hook

RaduBerinde:
`WithOnListen` tells an embedder when the S3 server is serving, but not where: the callback takes no arguments, and the server reports the addresses it bound nowhere else. An embedder that wants an ephemeral port therefore cannot ask for port 0; it has to pick a free port itself, release it, and pass it in, which loses to any other process that binds the same port in between.

Add `WithOnListenAddrs`, which passes the callback the address of every listener the server bound, in port-specification order, so an embedder can serve on `127.0.0.1:0` and learn the port the kernel chose. `WithOnListen` is unchanged. `MultiListener` gains `Addrs`, the every-listener counterpart of `Addr`, to supply them.

### DIFF
--- a/internal/netutil/multi_listener.go
+++ b/internal/netutil/multi_listener.go
@@ -124,6 +124,23 @@ func (ml *MultiListener) Addr() net.Addr {
 	return nil
 }
 
+// Addrs returns the address of every underlying listener, in the order the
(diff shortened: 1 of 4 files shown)


versity-github:

benmcclelland (approved):

PR #2375, opened 2026-09-08
tests: build the integration SDK config without the host's AWS setup

RaduBerinde:
The integration harness went through `config.LoadDefaultConfig` to build its `aws.Config`, even though it supplies the region, credentials, endpoint, and HTTP client itself. That made it read the host's shared AWS configuration, and with `AWS_PROFILE` set in the environment the SDK insists the named profile exist: on a machine whose shell sets a profile the SDK cannot find, every test in `cmd/versitygw` died at client setup with

    error: failed to get shared config profile, <name>

Build the `aws.Config` directly from the harness settings instead. Nothing the shared configuration could supply was used -- credentials and region were always overridden -- and disabling only the shared files would not have helped, since the SDK still requires a profile named by `AWS_PROFILE` to resolve. The default stderr logger `LoadDefaultConfig` installed is kept so `--debug` output is unchanged.

### DIFF
--- a/tests/integration/s3conf.go
+++ b/tests/integration/s3conf.go
@@ -24,13 +24,12 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	v4 "github.com/aws/aws-sdk-go-v2/aws/signer/v4"
-	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/credentials"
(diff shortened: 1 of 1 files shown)


versity-github:

RaduBerinde:
CI failures seem like flaky infrastructure.

benmcclelland (approved):

benmcclelland:
> CI failures seem like flaky infrastructure.

its on the todo list to see if we can better deal with the connection timeouts in the github actions.

PR #2377, opened 2026-09-08
posix: add AbsolutePaths option for embedding without chdir

RaduBerinde:
## Problem

`posix.New` calls `os.Chdir(rootdir)` and uses cwd-relative paths for every bucket and object. That is the cheapest way to address files, but the cwd is process-wide: embedding the gateway (`embedgw`) silently moves the host program's cwd. In particular, Go unit tests that embed the gateway can no longer read their test data files by relative path.

## Change

- New `PosixOpts.AbsolutePaths`. When set, `New` leaves the working directory alone and builds every path from the absolute root; a relative `VersioningDir`/`SideCarDir` is then resolved against the working directory rather than the root. **The default is unchanged: chdir and relative paths.**
- All bucket and object paths go through new `BucketPath`/`ObjectPath`, which return the name as-is by default and prefix the root with `AbsolutePaths`. An absolute "bucket" (the versioning directory substitution) is passed through unchanged.
- `tmpfile` records the bucket directory path so `link()` and its fallbacks use the same addressing; `ListBuckets` reads the root through the same helper.
- `meta.XattrMeta` needs the same root with `AbsolutePaths`. New `meta.RootDirSetter` interface; `posix.New` calls `WithRootDir` on storers that implement it in that mode. A zero `XattrMeta` keeps resolving against the cwd. `SideCar`/`NoMeta` unchanged. A type that embeds `XattrMeta` inherits a `WithRootDir` that returns a bare `XattrMeta`, so it needs its own (documented on `RootDirSetter`).
- `DeleteObject` (directory object), `ListParts`, and `UploadPartCopy` passed filesystem paths where the metadata API expects bucket/object names; they now pass names, so the sidecar layout is unchanged in both modes.
- Windows `handleParentDirError` walks up until `filepath.Dir` is a fixed point, which works for relative and absolute paths.
- scoutfs used cwd-relative bucket/object paths in `CreateBucket`, `GetObject`, `HeadObject`, `RestoreObject` and the glacier walk; they now go through `BucketPath`/`ObjectPath`. `scoutfs.New` resolves `rootdir` before `posix.New` so a relative root no longer reopens `rootdir/rootdir` after the chdir.
- `isBucketValid` unconditionally rejects names that do not denote a single entry under the root: `""`, `.`, `..`, names containing a path separator, and absolute paths. `XattrMeta` rejects `""`, `.` and `..` likewise. With relative paths `os.Stat("")` and `os.RemoveAll(".")` failed by accident; with absolute paths they would act on the root directory itself (reachable with strict bucket names disabled, or via the admin `change-bucket-owner` endpoint which does not validate `bucket`).
- scoutfs had its own `isBucketValid` whose `validateBucketName` flag was never set, so it accepted everything. It now delegates to the new exported `Posix.IsBucketValid`.
- `UploadPartCopy` did not validate the copy source's bucket name (unlike `CopyObject`); it does now.
- `New` opens the root after validating the versioning and sidecar directories, so those error paths no longer leak the root handle. The chdir still happens first, so a relative directory resolves against the root as before.

## Tests

- New `TestDefaultModeChangesWorkingDirectory` documents the default.
- New `TestRootDirIndependentOfWorkingDirectory`: `AbsolutePaths` with a relative root from an unrelated cwd, checks cwd is untouched and that put/get/list/delete, copy, multipart upload with checksums and part copy, directory-object delete, and invalid bucket names behave correctly under the root, for both metadata storers.
- New `TestVersioningDirIndependentOfWorkingDirectory`: same setup with a relative versioning directory; versions land there and not under the root or cwd.
- New `TestXattrMetaPath` covers cwd-relative and root resolution, absolute pass-through and the rejected names.
- New `BenchmarkPosix*` benchmarks (small-object head/get/put/list, both storers, both path modes).

## Benchmarks

Why the flag is opt-in: the default mode matches `main` within noise on both platforms, and absolute paths cost about 0.2µs (Linux) to 0.4µs (macOS) per path lookup, which on Linux is a sizable share of the metadata-heavy small-object operations. "Flag cost" compares the branch's `AbsolutePaths` column against the branch's default column.

Run with:

```
go test -run xxx -bench Posix -benchtime 300ms -count 6 ./backend/posix/
```

**Linux arm64 (podman VM, go1.25.14, overlayfs). Median ± stddev of 6 runs.**

| benchmark (µs/op) | main (chdir) | branch, default (chdir) | branch, `AbsolutePaths` | flag cost |
|---|---|---|---|---|
| GetObject/sidecar | 23.1 ± 0.5 | 22.7 ± 5.1 | 24.5 ± 1.0 | +8% |
| GetObject/xattr | 16.1 ± 0.2 | 15.7 ± 0.3 | 18.0 ± 0.9 | +15% |
| HeadBucket/sidecar | 0.46 ± 0.00 | 0.47 ± 0.00 | 0.70 ± 0.02 | +51% |
| HeadBucket/xattr | 0.46 ± 0.01 | 0.47 ± 0.00 | 0.70 ± 0.01 | +49% |
| HeadObject/sidecar | 23.4 ± 0.4 | 23.2 ± 0.2 | 25.6 ± 1.4 | +11% |
| HeadObject/xattr | 18.4 ± 1.1 | 18.2 ± 0.3 | 23.1 ± 0.5 | +27% |
| ListObjectsV2/sidecar | 1323.8 ± 34.2 | 1318.8 ± 36.3 | 1360.0 ± 12.4 | +3% |
| ListObjectsV2/xattr | 570.7 ± 4.4 | 565.0 ± 3.6 | 672.7 ± 28.6 | +19% |
| PutObject/sidecar | 98.4 ± 4.9 | 98.2 ± 8.4 | 100.1 ± 3.5 | +2% |
| PutObject/xattr | 68.6 ± 2.4 | 69.0 ± 4.4 | 70.8 ± 3.4 | +3% |

**macOS arm64 (go1.26.8, APFS, root not behind a symlink). Median ± stddev of 6 runs.**

| benchmark (µs/op) | main (chdir) | branch, default (chdir) | branch, `AbsolutePaths` | flag cost |
|---|---|---|---|---|
| GetObject/sidecar | 92.7 ± 6.1 | 94.3 ± 2.3 | 93.8 ± 4.0 | -1% |
| GetObject/xattr | 94.3 ± 1.1 | 92.0 ± 1.8 | 95.6 ± 3.0 | +4% |
| HeadBucket/sidecar | 0.54 ± 0.01 | 0.56 ± 0.02 | 0.91 ± 0.07 | +64% |
| HeadBucket/xattr | 0.54 ± 0.02 | 0.56 ± 0.01 | 0.91 ± 0.02 | +61% |
| HeadObject/sidecar | 81.1 ± 2.6 | 82.6 ± 2.8 | 83.3 ± 2.4 | +1% |
| HeadObject/xattr | 90.1 ± 1.8 | 89.2 ± 18.3 | 94.6 ± 2.3 | +6% |
| ListObjectsV2/sidecar | 4160.4 ± 175.5 | 4381.6 ± 225.4 | 4397.4 ± 1043.0 | +0% |
| ListObjectsV2/xattr | 2631.5 ± 565.6 | 2651.7 ± 82.4 | 2708.8 ± 122.9 | +2% |
| PutObject/sidecar | 693.7 ± 22.7 | 677.2 ± 14.3 | 691.6 ± 9.6 | +2% |
| PutObject/xattr | 362.8 ± 5.0 | 362.8 ± 7.5 | 373.5 ± 13.1 | +3% |

`os.Root` was also evaluated and rejected: Go's `Root` opens every intermediate path component with a separate `openat` (no `openat2`), which measured 6-50x slower than a plain `stat` depending on key depth, and it refuses to follow symlinks that leave the root, which breaks `--bucketlinks`.

---
Code and description generated with Claude Code; the design decisions are mine (Radu).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/meta/meta.go
+++ b/backend/meta/meta.go
@@ -45,3 +45,18 @@ type MetadataStorer interface {
 	// directory for an object is renamed so that metadata stays in sync.
 	RenameObject(bucket, oldObject, newObject string) error
 }
+
+// RootDirSetter is implemented by metadata storers that keep metadata on the
+// bucket and object files themselves and therefore need to know where those
+// files live. A backend that does not change the process working directory
+// to its root calls WithRootDir with the absolute path of the root before
+// using the storer; the returned storer resolves bucket names against that
+// directory instead of the working directory. Storers that keep metadata
+// elsewhere (SideCar, NoMeta) do not implement it.
+//
+// A type that embeds XattrMeta inherits its WithRootDir, which returns a bare
+// XattrMeta and so drops the outer type; such a type must provide its own
+// WithRootDir.
+type RootDirSetter interface {
+	WithRootDir(rootdir string) MetadataStorer
+}
--- a/backend/meta/xattr.go
+++ b/backend/meta/xattr.go
@@ -31,7 +31,41 @@ var (
 	ErrNoSuchKey = errors.New("no such key")
 )
 
-type XattrMeta struct{}
+// XattrMeta stores metadata as extended attributes on the bucket and object
+// files. The zero value resolves bucket names against the process working
+// directory, which the posix backend sets to its root directory; a backend
+// that keeps the working directory supplies the root through WithRootDir.
+type XattrMeta struct {
+	// rootdir is the absolute path bucket names are resolved under, or ""
+	// to resolve them against the process working directory.
+	rootdir string
+}
+
+var _ RootDirSetter = XattrMeta{}
+
+// WithRootDir returns a copy of x that resolves bucket names against rootdir.
+func (x XattrMeta) WithRootDir(rootdir string) MetadataStorer {
+	x.rootdir = rootdir
+	return x
+}
+
+// path returns the filesystem path holding the attributes of object in
+// bucket (of the bucket itself when object is empty).
+//
+// The bucket argument is normally a bucket name, resolved under the root
+// directory (or the working directory when no root is set). The versioning
+// code instead passes the absolute path of a bucket's versioning directory;
+// an absolute bucket is used as given.
+func (x XattrMeta) path(bucket, object string) (string, error) {
+	if filepath.IsAbs(bucket) {
+		return filepath.Join(bucket, object), nil
+	}
+	if bucket == "" || bucket == "." || bucket == ".." {
+		// Would resolve to the root directory itself or its parent.
+		return "", fmt.Errorf("xattr metadata: invalid bucket name %q", bucket)
+	}
+	return filepath.Join(x.rootdir, bucket, object), nil
+}
 
 // RetrieveAttribute retrieves the value of a specific attribute for an object in a bucket.
 func (x XattrMeta) RetrieveAttribute(f *os.File, bucket, object, attribute string) ([]byte, error) {
@@ -43,7 +77,11 @@ func (x XattrMeta) RetrieveAttribute(f *os.File, bucket, object, attribute strin
 		return b, err
 	}
 
-	b, err := xattr.Get(filepath.Join(bucket, object), xattrPrefix+attribute)
+	name, err := x.path(bucket, object)
+	if err != nil {
+		return nil, err
+	}
+	b, err := xattr.Get(name, xattrPrefix+attribute)
 	if errors.Is(err, xattr.ENOATTR) {
 		return nil, ErrNoSuchKey
 	}
@@ -63,7 +101,11 @@ func (x XattrMeta) StoreAttribute(f *os.File, bucket, object, attribute string, 
 		return err
 	}
 
-	err := xattr.Set(filepath.Join(bucket, object), xattrPrefix+attribute, value)
+	name, err := x.path(bucket, object)
+	if err != nil {
+		return err
+	}
+	err = xattr.Set(name, xattrPrefix+attribute, value)
 	if errors.Is(err, syscall.EROFS) {
 		return s3err.GetAPIError(s3err.ErrMethodNotAllowed)
 	}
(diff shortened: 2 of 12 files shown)


versity-github:

benmcclelland (changes requested):
os.Chdir() was an intentional performance optimization in the POSIX backend, so removing it entirely changes the behavior of the primary VersityGW deployment path.

I would prefer retaining chdir by default and introducing an explicit option for embedding scenarios that resolves paths relative to the configured root instead. That would preserve the existing performance characteristics for the primary use case while allowing embedded applications to opt out of the process-wide working-directory change.

also, current test failures are from staticcheck

RaduBerinde:
Interesting! I believe using [os.OpenRoot](https://pkg.go.dev/os#OpenRoot) would get the best of both worlds? (and would improve security by not allowing any access above that directory)

RaduBerinde:
No, `os.Root` is actually much slower (it walks the path one component at a time):

key depth | chdir-relative | absolute | os.Root
-- | -- | -- | --
bucket/obj | 1.4 µs | 2.0 µs | 12.7 µs
bucket/k/k/obj | 1.3 µs | 2.1 µs | 39.4 µs
bucket/k/k/k/k/k/obj | 1.5 µs | 2.2 µs | 78.5 µs

I will work on adding a flag and also a benchmark.

benmcclelland:
> Interesting! I believe using [os.OpenRoot](https://pkg.go.dev/os#OpenRoot) would get the best of both worlds? (and would improve security by not allowing any access above that directory)

Oh, interesting. I'll check this out. Looks promising. The specific cases that were causing some slowdowns were with distributed filesystems that were all trying to constantly get locks on root inode with full path access. The chdir allowed the access to prevent getting cluster locks for anything above the gateway root. I dont really expect this overhead to be an issue on single node filesystems.

RaduBerinde:
Updated the change to use a flag. I also added a benchmark and posted some results.

benmcclelland (approved):

PR #2378, opened 2026-09-08
fix: close connections with unread chunked bodies

benmcclelland:
A streamed chunked request can leave bytes queued after the handler returns. Reusing that connection lets fasthttp parse those bytes as the next request, allowing a shared upstream proxy connection to mix requests across tenants. Mark chunked requests Connection: close when the middleware cannot safely drain them, preventing leftover bytes from crossing the request boundary.

Draining is intentionally skipped because fasthttp's request stream reads past the terminating chunk when probing for EOF and can block waiting for another chunk header. The connection-reuse sacrifice is therefore required to avoid both request desynchronization and delaying the response. Content-Length bodies retain the existing bounded drain behavior.

### DIFF
--- a/s3api/middlewares/drain-request-body.go
+++ b/s3api/middlewares/drain-request-body.go
@@ -81,11 +81,12 @@ func drainRequestBody(ctx fiber.Ctx) {
 	// body, but not for a chunked one: past the terminating chunk it goes back
 	// to the socket for another chunk header that will never come. Reading a
 	// chunked body the handler already finished would block until the deadline
-	// and hold the response back with it, so only Content-Length framing is
-	// drained. Nothing is lost for the aws-chunked uploads this exists for --
-	// STREAMING-* payloads carry a Content-Length.
+	// and hold the response back with it, so do not drain chunked framing. The
+	// stream may still have unread bytes, though, so the connection must not be
+	// reused for another request.
 	cLength := ctx.Request().Header.ContentLength()
-	if cLength <= 0 {
+	if cLength < 0 {
+		ctx.Response().Header.SetConnectionClose()
 		return
 	}
 
--- a/s3api/middlewares/drain-request-body.go
+++ b/s3api/middlewares/drain-request-body.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/gofiber/fiber/v3"
 	"github.com/versity/versitygw/debuglogger"
+	"github.com/versity/versitygw/s3api/utils"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


s3api/middlewares/drain-request-body.go
```diff
@@ -81,11 +81,12 @@ func drainRequestBody(ctx fiber.Ctx) {
...
 	// chunked body the handler already finished would block until the deadline
-	// and hold the response back with it, so only Content-Length framing is
-	// drained. Nothing is lost for the aws-chunked uploads this exists for --
-	// STREAMING-* payloads carry a Content-Length.
+	// and hold the response back with it, so do not drain chunked framing. The
+	// stream may still have unread bytes, though, so the connection must not be
+	// reused for another request.
 	cLength := ctx.Request().Header.ContentLength()
-	if cLength <= 0 {
+	if cLength < 0 {
```
niksis02: This forces `Connection: close` on every chunked request, including ones that finished cleanly and succeeded — not just the ones we bailed out of early.

There's no branch here for "the handler already read the entire body, including the terminating chunk" vs. "the handler stopped partway through." Both hit this same path and both pay the same price: no keep-alive, ever, for any chunked-body request.

This isn't just theoretical. `SignedStreamingPayload_invalid_chunk_size` test sends `Content-Length: -1` + `Transfer-Encoding: chunked` for a signed-streaming payload and expects several cases to succeed. So a normal, successful upload that happens to use chunked encoding without an announced `Content-Length` now permanently loses connection reuse.

The reason we can't already tell the two cases apart is that fasthttp's chunked stream doesn't expose "did I already hit the real end" — but we can track that ourselves instead of asking fasthttp a second, unsafe question: wrap the body stream in a small sticky-state reader before the handler touches it, remembering whether the first terminal result was `io.EOF` (fully, cleanly finished — safe to keep the connection alive), no terminal result yet (handler just stopped — safe to keep draining through the same wrapper, same as we do for `Content-Length`), or a real parse error (genuinely broken — keep today's close-without-draining behavior). That way "always close" only applies to the genuinely ambiguous case instead of every chunked request.

s3api/middlewares/drain-request-body.go
```diff
@@ -81,11 +81,12 @@ func drainRequestBody(ctx fiber.Ctx) {
...
+		ctx.Response().Header.SetConnectionClose()
 		return
```
niksis02, line 96: We give up on draining entirely here, which brings back the exact "client sees a connection reset instead of our S3 error".

The reasoning for not draining a finished chunked stream is sound (re-reading past the real end risks blocking or misreading a next chunk header). But by this point we've already committed to `Connection: close` above, so the original desync/cross-tenant risk this file was built to close is already fully neutralized regardless of whether we attempt a drain here — reusing the connection is what made stray bytes dangerous, and that's off the table either way now.

Given that, it should be safe to still attempt a bounded, best-effort drain here (reusing the existing drainReader/drainIdleTimeout/drainTotalTimeout/maxDrainBytes machinery) before closing — worst case it eats up to the timeout, best case the client gets to finish writing and read the actual S3 error instead of an abrupt reset. Since we're throwing the connection away no matter what happens in this drain, there's no downside to trying?

niksis02 (changes requested):

benmcclelland:
approved now with changes.

PR #2383, opened 2026-09-11
fix: object verson mtime

### LINKED ISSUE #2311: [Bug] - ListObjectVersions - non-current versions report the archive time as LastModified instead of the original upload time
**Describe the bug**
When using the POSIX backend with `--versioning-dir`, overwriting a versioned object creates a copy of the previous object version in the versioning directory. This copy is created as a brand-new file (`io.Copy` into a fresh tmp file), so its mtime is set to the moment the **next** upload happens — not to the moment when that version was originally uploaded.

As a result, `ListObjectVersions` reports `LastModified` of every non-current version as the time when it was archived (i.e. the timestamp of the subsequent PUT), which breaks client logic that relies on per-version timestamps (e.g. rendering version history).

**To Reproduce**
1. Start versitygw:
   `./versitygw --access=... --secret=... --region=us-east-1 posix --versioning-dir /tmp/versioning /tmp/gw`
2. Create a bucket and enable versioning (`PutBucketVersioning`)
3. `aws s3api put-object --bucket test-bucket --key testfile --body /tmp/data` at time t0
4. Wait at least a few seconds
5. `aws s3api put-object --bucket test-bucket --key testfile --body /tmp/data` at time t1
6. `aws s3api list-object-versions --bucket test-bucket --prefix testfile`

**Actual result:** both versions report essentially the same `LastModified` ≈ t1.
Observed on our setup (two uploads minutes apart):

01M0TNG9HB8G9FGC1PMRF699S7 | latest: true  | LastModified: 2026-08-24T19:56:04.000Z
01M0TMBY6MV5QGSPBZH1SPA589 | latest: false | LastModified: 2026-08-24T19:56:04.000Z  ← uploaded earlier

Filesystem evidence from the backend directories (mtimes differ by only 8 ms although
the versions were uploaded minutes apart):

$ stat /tmp/gw/test-bucket/testfile                     # latest version
2026-08-24 21:56:04.519661779 +0200
$ stat /tmp/versioning/test-bucket/<hash>/01M0TMBY...   # archived version
2026-08-24 21:56:04.511661786 +0200

**Expected behavior**
A non-current object version should keep the `LastModified` value of its own original upload (matching AWS S3 behavior), i.e. the older version above should report its original upload time t0, not t1.

**Server Version**
Version: v1.7.0 (build 8a3dbcf, 2026-07-15)

**Additional context (from AI findings)**

Root cause (possible):
- [`createObjVersion()` in `backend/posix/posix.go`](https://github.com/versity/versitygw/blob/762e244b4352/backend/posix/posix.go#L995) creates the archived version by opening the current object and copying it via [`io.Copy(f.File(), sf)`](https://github.com/versity/versitygw/blob/762e244b4352/backend/posix/posix.go#L1029) into a freshly created tmp file. The new file therefore receives the current time as mtime. A `rename()`/`link()` of the existing file would preserve the original mtime.
- After copying xattrs, no `os.Chtimes()` (or equivalent) is called to restore the original modification time — there is no occurrence of `Chtimes`/`Utimes` anywhere in the file.
- `ListObjectVersions` then reports [`LastModified: nf.ModTime()`](https://github.com/versity/versitygw/blob/762e244b4352/backend/posix/posix.go#L1364) (also [L1389](https://github.com/versity/versitygw/blob/762e244b4352/backend/posix/posix.go#L1389), [L1497](https://github.com/versity/versitygw/blob/762e244b4352/backend/posix/posix.go#L1497)) straight from the archived file's mtime, propagating the wrong timestamp to clients.

JoshPigott:
Fixes #2311

## Why
- All non-latest versions have the wrong mtime.
- This happens because the archived version is created by opening the current object and copying it, which changes the mtime instead of preserving the original.

## Changes
- Keeping the original mtime.

Please let me know if you would like any changes. Thanks

### DIFF
--- a/backend/posix/posix.go
+++ b/backend/posix/posix.go
@@ -1166,8 +1166,18 @@ func (p *Posix) createObjVersion(bucket, key string, size int64, acc auth.Accoun
 	}
 	defer f.cleanup()
 
+	// Keeps original time
+	srcInfo, _ := sf.Stat()
+	beforeMTime := srcInfo.ModTime()
+
 	// Prioritize copy_file_range for internal file-to-file version copies.
 	_, err = io.Copy(f.File(), sf)
+
+	err = os.Chtimes(f.File().Name(), time.Now(), beforeMTime)
+	if err != nil {
+		return versionPath, err
+	}
+
 	if err != nil {
 		return versionPath, err
 	}


versity-github:

backend/posix/posix.go
```diff
@@ -1166,8 +1166,18 @@ func (p *Posix) createObjVersion(bucket, key string, size int64, acc auth.Accoun
 	}
 	defer f.cleanup()
 
+	// Keeps original time
+	srcInfo, _ := sf.Stat()
+	beforeMTime := srcInfo.ModTime()
```
benmcclelland, line 1171: you can't ignore the err from stat here. if err != nil, then srcInfo = nil and the srcInfo.ModTime() will panic. I would recommend a best effort approach where you only call srcInfo.ModTime() and os.Chtimes() if sf.Stat() err == nil.

benmcclelland (changes requested):

backend/posix/posix.go
JoshPigott: Thanks for that Ben, I appreciate it. Silly mistake.

benmcclelland (approved):

PR #2385, opened 2026-09-11
webui: show upload progress in the Explorer

crsolucoes:
## Summary
Uploading a large file through the Explorer gives no feedback at all today — the toast only fires once at the start ("Uploading...") and once at the end ("Uploaded"), so a multi-GB multipart upload looks completely stalled to the user for however long it actually takes, with no way to tell it apart from a hung connection.

Adds a small persistent widget (bottom-right, same visual style as the existing toasts) that tracks bytes uploaded / total bytes across the whole upload batch, updating after each part of a multipart upload completes (or once a small single-PUT file finishes). It disappears once the batch is done, right before the existing success/failure toast.

No changes to any upload/network behavior — this only observes progress already being made by the existing `uploadMultipart`/`putObject` calls and renders it.

## Test plan
- [x] Tested the progress-accumulation logic in isolation (mocked DOM, simulated a 4GB multipart upload + a small single-PUT file in the same batch) — caught and fixed one real bug in the process (the small-file branch was reporting progress *before* the PUT resolved instead of after, which under-reported the final byte count).
- [ ] Manual smoke test against a running instance (no local Go toolchain available in my environment to build/run the full server — relying on CI + maintainer review for this)

### DIFF
--- a/webui/web/explorer.html
+++ b/webui/web/explorer.html
@@ -2965,6 +2965,41 @@ under the License.
     // Upload
     // ============================================
 
+    // Persistent widget (unlike showToast's fire-and-forget toasts) tracking
+    // overall byte progress across every file/part of the current upload
+    // batch, so a large file no longer looks stalled with no feedback.
+    let uploadProgressEl = null;
+
+    function showUploadProgress(fileName, fileIndex, fileCount, bytesDone, bytesTotal) {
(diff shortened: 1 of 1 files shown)


versity-github:

benmcclelland (approved):

PR #2389, opened 2026-09-13
Return IllegalLocationConstraintException for a mismatched LocationConstraint

### LINKED ISSUE #2246: [Bug] - Return `IllegalLocationConstraintException` when `CreateBucket` specifies an incorrect `LocationConstraint`
**Describe the bug**
If the gateway is running in one region (for example, `us-west-1`) and the specified `LocationConstraint` is either a different region (for example, `eu-central-1`) or is omitted, the gateway should return an `IllegalLocationConstraintException` error.

The exception is `us-east-1`, which is a special case and requires `LocationConstraint` to be omitted.

**To Reproduce**
Run the gateway in `us-west-1`, then:
```bash
aws s3api create-bucket --bucket test --region us-west-1 --create-bucket-configuration LocationConstraint=eu-central-1
```
**Expected behavior**
The following error should be returned:
```xml
<Error><Code>IllegalLocationConstraintException</Code><Message>The eu-central-1 location constraint is incompatible for the region specific endpoint this request was sent to.</Message><RequestId>0D43NRDGX6VWDESA</RequestId><HostId>b8uJ8XzVZU8wnOBvlOTwegtCzfQQJ+F6imPiFuzJ4TEKZ+nQdsz4yEub6UtO3GtlukbJ6luREGCy00tALh24D5n8HZEKS9fq</HostId></Error>
```

dualfroz:
A region endpoint returned `InvalidLocationConstraint` for a different region and accepted a missing constraint; both now return `IllegalLocationConstraintException` like AWS.
Note this is a behaviour change: outside us-east-1, CreateBucket without a body will now be rejected.

Fixes #2246


versity-github:

dualfroz:
hey, was just about to go to sleep but fixed it real quick. the test helpers created buckets without a location constraint, so everything outside us-east-1 died at setup. they send the region now

s3api/controllers/bucket-put.go
```diff
@@ -546,6 +546,32 @@ func (c S3ApiController) PutBucketAcl(ctx fiber.Ctx) (*Response, error) {
 	}, err
 }
 
+// validateLocationConstraint checks a CreateBucket location constraint. The
+// global endpoint serves defaultRegion and takes no constraint; any other
+// region is a region specific endpoint and requires the constraint to name it.
+func validateLocationConstraint(constraint *string, region string) error {
```
niksis02: Move this utility function definition to the `s3api/utils/utils.go` package.

niksis02 (changes requested):
Overall looks good to me. Just a minor code style fix is needed.

s3api/controllers/bucket-put.go
dualfroz: done, moved it to utils

niksis02 (approved):

benmcclelland (approved):

Issue #2169, opened 2026-06-05
[Bug] - ListObject versions - `encoding-type` query is ignored


lrm25:
**Describe the bug**
If a user sends a message to S3 with an `encoding-type` query param that isn't `url`, the following error is returned:
```
HTTP/1.1 400 Bad Request
x-amz-request-id: F60HC8TTZGFY1491
x-amz-id-2: g+FuGIkfPuE3u1DrJCBRgclnB5xtyammUy2oeQU1BxwEyz2iaG19SZMLSBXGgMYYNhh098Qt5yA=
Content-Type: application/xml
Transfer-Encoding: chunked
Date: Fri, 05 Jun 2026 19:47:22 GMT
Connection: close
Server: AmazonS3

<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidArgument</Code><Message>Invalid Encoding Method specified in Request</Message><ArgumentName>encoding-type</ArgumentName><ArgumentValue>{invalid encoding type}</ArgumentValue><RequestId>F60HC8TTZGFY1491</RequestId><HostId>g+FuGIkfPuE3u1DrJCBRgclnB5xtyammUy2oeQU1BxwEyz2iaG19SZMLSBXGgMYYNhh098Qt5yA=</HostId></Error>
```
Also, when `encoding-type` is set to `url`, and a key with the name of, for example, `a+ b.txt`, is used, the key is returned correctly encoded in the XML:
```
<Key>a%2B+b.txt</Key>
```

**To Reproduce**
1. Start versitygw, e.g.:  `./versitygw --access=AKIA**** --secret=******** --region=us-east-1 --cert /Users/lukemccrone/devel/versitygw/tester/cert.pem --key /Users/lukemccrone/devel/versitygw/tester/versitygw.pem posix --versioning-dir /tmp/versioning /tmp/gw-tester`
[...]

PR #2390, opened 2026-09-14
fix: return Marker and ContinuationToken when the request omits them

### LINKED ISSUE #2191: [Bug] - ListObjects/ListObjectsV2 missing Marker/ContinuationToken when empty
Describe the bug
When a client sends a ListObjects request without a Marker parameter (or with an empty Marker), VersityGW does not include the <Marker> element in the XML response. Similarly, for ListObjectsV2, the <ContinuationToken> element is missing when the request does not contain a continuation-token (or the token is empty).

According to the AWS S3 specification, these fields must be present in the response, even if their value is an empty string. The absence of these fields causes the Ceph s3‑tests suite to fail with KeyError: 'Marker' (or KeyError: 'ContinuationToken'), breaking compatibility.

To Reproduce
Start VersityGW with POSIX backend.
Create a bucket and upload some objects.
Send a ListObjects request without the Marker parameter:
aws s3api list-objects --bucket test-bucket --endpoint-url http://localhost:7070
Actual response: XML does not contain a <Marker> element.
Expected response: XML contains <Marker></Marker> (empty string).

Send a ListObjectsV2 request without continuation-token:
aws s3api list-objects-v2 --bucket test-bucket --endpoint-url http://localhost:7070
Actual response: XML does not contain a <ContinuationToken> element.
Expected response: XML contains <ContinuationToken></ContinuationToken>.

Run the Ceph s3‑tests:
S3TEST_CONF=s3tests.conf pytest s3tests/functional/test_s3.py::test_bucket_list_marker_none -v
Result: KeyError: 'Marker' (FAILED).

Expected behavior
ListObjects responses must always include the <Marker> element, with the value being the marker used in the request (or an empty string if none was provided).
ListObjectsV2 responses must always include the <ContinuationToken> element, with the value being the token used in the request (or an empty string if none was provided).
This is required for AWS S3 compatibility and for client libraries that expect these fields to exist.

Proposed fix
1. In ListObjects controller (s3api/controllers/bucket-get.go)
Convert an empty marker query parameter to nil before passing to the backend, so that the backend knows no marker was provided.
After receiving the response, if res.Marker == nil, set it to an empty string.

2. In ListObjectsV2 controller (s3api/controllers/bucket-get.go)
Convert an empty continuation-token to nil before passing to the backend.
After receiving the response, if res.ContinuationToken == nil, set it to an empty string.

Diff:
```diff
--- s3api/controllers/bucket-get.go.orig        2026-06-16 13:06:20.704421695 +0300
+++ s3api/controllers/bucket-get.go     2026-06-16 13:50:18.985679617 +0300
@@ -595,11 +595,18 @@
                }, err
        }
 
+       // If continuation token is an empty string, treat it as "no token" by passing nil.
+       // This ensures the backend starts listing from the first object.
+       var tokenPtr *string
+       if cToken != "" {
+               tokenPtr = &cToken
+       }
+
        res, err := c.be.ListObjectsV2(ctx.Context(),
                &s3.ListObjectsV2Input{
                        Bucket:            &bucket,
                        Prefix:            &prefix,
-                       ContinuationToken: &cToken,
+                       ContinuationToken: tokenPtr,
                        Delimiter:         &delimiter,
                        MaxKeys:           &maxkeys,
                        StartAfter:        &sAfter,
@@ -613,6 +620,14 @@
                }, err
        }
 
+       // AWS S3 always includes the ContinuationToken field in the response, even if it's empty.
+       // If the backend returns nil (e.g., when no token was provided), set it to an
+       // empty string to ensure the XML response contains the <ContinuationToken> element.
+       if res.ContinuationToken == nil {
+               empty := ""
+               res.ContinuationToken = &empty
+       }
+
        return &Response{
                Headers: map[string]*string{
                        "x-amz-bucket-region": &region,
@@ -669,11 +684,18 @@
                }, err
        }
 
+       // If marker is an empty string, treat it as "no marker" by passing nil.
+       // This ensures the backend starts listing from the first object.
+       var markerPtr *string
+       if marker != "" {
+               markerPtr = &marker
+       }
+
        res, err := c.be.ListObjects(ctx.Context(),
                &s3.ListObjectsInput{
                        Bucket:    &bucket,
                        Prefix:    &prefix,
-                       Marker:    &marker,
+                       Marker:    markerPtr,
                        Delimiter: &delimiter,
                        MaxKeys:   &maxkeys,
                })
@@ -685,6 +707,14 @@
                }, err
        }
 
+       // AWS S3 always includes the Marker field in the response, even if it's empty.
+       // If the backend returns nil (e.g., when no marker was provided), set it to an
+       // empty string to ensure the XML response contains the <Marker> element.
+       if res.Marker == nil {
+               empty := ""
+               res.Marker = &empty
+       }
+
        return &Response{
                Headers: map[string]*string{
                        "x-amz-bucket-region": &region,
```

Test results
After applying the fix, the following Ceph s3‑tests now pass:
test_bucket_list_marker_none
test_bucket_list_marker_empty
test_bucket_listv2_continuationtoken_empty

$ S3TEST_CONF=s3tests.conf pytest s3tests/functional/test_s3.py::test_bucket_list_marker_none -v
...
s3tests/functional/test_s3.py::test_bucket_list_marker_none PASSED

$ S3TEST_CONF=s3tests.conf pytest s3tests/functional/test_s3.py::test_bucket_list_marker_empty -v
...
s3tests/functional/test_s3.py::test_bucket_list_marker_empty PASSED

$ S3TEST_CONF=s3tests.conf pytest s3tests/functional/test_s3.py::test_bucket_listv2_continuationtoken_empty -v
...
s3tests/functional/test_s3.py::test_bucket_listv2_continuationtoken_empty PASSED

Additional context
This fix improves AWS S3 compliance and prevents KeyError exceptions in client code that expects these fields.

LennardGeissler:
Fixes #2191.

## Problem

`ListObjects` and `ListObjectsV2` drop the `Marker` and `ContinuationToken` elements from the listing response whenever the request carries no marker or token. S3 includes both unconditionally, echoing the request value back, so a client that reads the field without a presence check breaks against the gateway.

## Root cause

The elements go missing *before* serialization, not during it. The backends build their result with `backend.GetPtrFromString`, which collapses `""` to a nil pointer, and `encoding/xml` omits nil pointers entirely:

```go
func GetPtrFromString(str string) *string {
	if str == "" {
		return nil
	}
	return &str
}
```

Fixing this in `backend/posix` alone would leave the other backends behind, so the controller restores the field from the request after the backend call. That covers posix, the s3 proxy, azure and scoutfs in one place.

## Verification

The three Ceph s3-tests named in the issue, run against a posix-backed gateway before and after (two gateways side by side, same objects):

| test | `main` | this branch |
| --- | --- | --- |
| `test_bucket_list_marker_none` | `KeyError: 'Marker'` | pass |
| `test_bucket_list_marker_empty` | `KeyError: 'Marker'` | pass |
| `test_bucket_listv2_continuationtoken_empty` | `KeyError: 'ContinuationToken'` | pass |
| marker echoed back when one *is* supplied (regression guard) | pass | pass |

The response difference is exactly the presence of the element:

```diff
-<ListBucketResult ...><Name>b</Name><MaxKeys>1000</MaxKeys>...
+<ListBucketResult ...><Name>b</Name><Marker></Marker><MaxKeys>1000</MaxKeys>...
```

## Changes

- `s3api/controllers/bucket-get.go` — the fix.
- `s3api/controllers/bucket-get_test.go` — the controller unit tests asserted the old behaviour (`Marker: nil`) and are updated. Worth noting: `utils.GetStringPtr` collapses `""` to nil in the same way `backend.GetPtrFromString` does, so the tests take the address of an empty string directly. Two helpers quietly turning `""` into nil is how this class of bug arises in the first place.
- `tests/drivers/list_objects/list_objects_rest.sh`, `tests/test_rest_list_objects.sh` — a REST test per version, asserting the element is present and empty. `check_xml_element` distinguishes the two cases correctly: a missing element fails on the existence check, an empty one returns `""`.

`go test ./s3api/... ./backend/... ./auth/... ./s3response/... ./s3err/...` passes; `gofmt` clean.

I could not run the bats suite locally (needs bash >= 4, macOS ships 3.2), so the REST tests are written but unverified by me — please shout if the CI run shows I got the harness conventions wrong. The behaviour they assert is verified by the table above.

## Adjacent, deliberately not fixed here

Same root cause, but outside this issue — happy to fold either in if you'd prefer one PR:

1. `<Prefix>` is dropped from both v1 and v2 responses for the same reason. AWS returns it as an empty element.
2. `<EncodingType></EncodingType>` is emitted on *every* listing, empty, although S3 only returns it when `encoding-type` was requested. That one belongs to #2169, and I've left a note there.


versity-github:

benmcclelland:
The test failures look related to this change. Might just need to update the tests for the new behavior for this.

LennardGeissler:
You're right — both failures were the integration tests asserting the old behaviour. Pushed in 6c2e8f5.

`ListObjects_list_all_objs` and `ListObjectsV2_list_all_objs` checked that `Marker` and `ContinuationToken` come back nil. With the fix they arrive as an empty element, which the SDK parses into a pointer to `""`, so the assertions now check for present-and-empty rather than nil. That distinction matters here, so they are explicit nil and value checks rather than a `getString` comparison, which would collapse the two cases.

I left the neighbouring assertions for `NextMarker`, `NextContinuationToken`, `Delimiter`, `Prefix` and `StartAfter` untouched — none of those are echoed request values, and the fix doesn't affect them.

All five failing jobs were these same two tests. Verified locally against a posix gateway on the merged branch, and `go test ./s3api/...` passes.

Also worth noting for the open question in my description: the two new bats tests did run in CI and passed (`ok 17`, `ok 18` in `tests/test_rest_list_objects.sh`), so the harness conventions were fine.

benmcclelland:
The AWS docs for the ListObjectsV2 states that ContinuationToken is only present in the response if it was sent with the request.
https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_ResponseElements
```
ContinuationToken
If ContinuationToken was sent with the request, it is included in the response. You can use the returned ContinuationToken for pagination of the list response.
```
And I just confirmed with a test against an s3 bucket. I think this behavior is correct for the V1 response, but not for the V2 response.

Issue #2388, opened 2026-09-12
[Feature] - Unify cuObject clients onto the RC v2 data plane


potatogim:
**Describe the solution you'd like**

The gateway now has two RDMA object-transfer protocols. The v1 cuObject stack speaks a single-request ud-style protocol whose token describes only the remote buffer, and its backend rejects RDMA multipart uploads because the part payload arrives out of band while the backend expects it in the request body. The v2 RC data plane (#2330, #2361) carries the object offset and transfer size explicitly in PREPARE, stages received bytes so part uploads dispatch to the backend like any other write, and uses standard ibverbs memory registrations in the gateway transfer path with nothing that encodes the client's memory vendor. As it stands, cuObject clients cannot use the RC multipart support or future v2 wire extensions without separate v1 protocol work.

Keeping the v1 stack in place for existing clients and adding a v2 client path for cuObject would close that gap; the work could be grouped into four reviewable areas.

A client session core in `cuwrapper/rc-client` ports the client half of the hipobj-rc-v2 session: the phase machine for the PREPARE/READY exchange, whose final result is returned by READY, the wire codec, the QP INIT/RTR/RTS transitions, and the token codec, mirroring the server work tracked in #2330 behind the same dlopen-based ibverbs shim so no link-time verbs dependency is added. Control requests are SigV4-signed, transport waits - connect, DNS resolution, poll loops, and the PUT grace interval - are bounded by cancellation and deadlines, and target paths are percent-encoded on the wire.

Memory registration covers CUDA device memory and host memory as ibverbs MRs (nvidia-peermem for device memory, ordinary host registration otherwise), and a cgo binding (`rdma/rcclient`) exposes the client to Go tooling with the same guard-and-drain pattern the server binding uses. The client probes the live connection before admission and binds admission to the connection identity, so a replacement connection is re-probed before it carries transfers.

On the gateway side, the target's query string is preserved when the target is split (`rdma/rcroutes`), uploadId and partNumber are passed to UploadPart, part reads are dispatched as GetObject with PartNumber, the effective multipart part limit is enforced on part numbers, and multipart part-transfer support is advertised with the mp capability. On the wire the gateway cannot tell the two v2 clients apart.

A v2 mode in cuobjtest drives the path end to end over host memory, including multipart part uploads over RC, and a regression suite covers the client behaviors above.

[...]

PR #2391, opened 2026-09-14
rdma: carry multipart semantics through the RC routes

potatogim:
Part of #2388.

Carries multipart semantics through the RC routes so part uploads and part reads dispatched by an RC v2 client behave like their REST counterparts. Target parsing keeps the raw query, UploadPart and GetObject receive uploadId and partNumber, the configured part limit is enforced, and READY advertises the mp capability.

Route tests cover target parsing, multipart validation and dispatch, the part limit, and capability advertisement.


versity-github:

benmcclelland:
We have a staticcheck warning on these along with the above security comments:
```
  Error: rdma/rcroutes/capabilities_linux_test.go:44:2: field stageFail is unused (U1000)
```

The multipart semantics added in this PR are not actually carried through the live RC route path. Only the tests call parseSemanticOp(). The actual Prepare/Ready flow still calls splitTarget(), then it strips the query string before parsing, then it authorizes and records the session as a plain object GET/PUT.

That means uploadId and partNumber never survive into the real route lifecycle, so a multipart upload/read is indistinguishable from a normal object transfer in production. In other words, the PR adds a semantic layer but never wires it into the code path that actually executes.

Do we need to make sure hipobject repo will accept these new capabilities, or have you already got the green light from them?

potatogim:
Thanks for the review, and for taking the time to look through this. The staticcheck warning and the two CodeQL findings are addressed in 6922377b: the unused fixture field is removed, and both int32 conversions now check the range first, since the part
limit flag itself has no upper bound.

On the semantic wiring -- I think that concern is fair to raise, so I went back through the live path again before replying. As far as I can trace it in the current head, prepareCore() parses the target into a semantic op before authorizing (routes_linux.go), and the parsed op flows into authorize() and the audit record;
stageGet() dispatches part reads through GetObjectInput.PartNumber and commitPut() dispatches part uploads through UploadPart with the parsed uploadId. READY re-parses the stored target the same way before the transfer. It is entirely possible we are looking at different revisions, so if what you are seeing does not match that, I would appreciate a pointer and I will gladly walk through the call chain together.

While re-checking, I also noticed and corrected four things in the same flow, in case they help: part reads now reject a mismatch between the announced size and the backend's reported ContentLength when that value is present and nonnegative, and probe one byte past the staging buffer otherwise, so an undeclared oversized part cannot be silently truncated. Part uploads now skip the overwrite/object-lock check, matching the REST UploadPart controller. Audit synthesis keeps the query separate from the object key. PREPARE authorization, session-creation, staging, and
audit-admission failures now retain the parsed multipart identity;
malformed header failures still use the plain operation. I would appreciate a second pair of eyes on all of these.

On hipObject: I have submitted the RC v2 client implementation in ROCm/hipObject#32, which carries the target-query wire support.
The PR is still open with maintainer review requested, so I cannot yet say whether they will accept it. Plain GET/PUT targets without
a query remain supported by these routes independently of that PR landing, but I would of course defer to you on whether you would rather hold this until that side is settled.

Issue #32, opened 2023-05-26
internal server error log


benmcclelland:
The internal server errors usually mean something happened server side that an admin would like to know about.  We should consider an internal serve error log so that we can monitor for when these types of errors happen.

PR #2394, opened 2026-09-14
rdma/rcobj: add a Go wrapper on libhipobj for the RC v2 data path

### LINKED ISSUE #2388: [Feature] - Unify cuObject clients onto the RC v2 data plane

potatogim:
Fixes #2388. Builds on #2391.

Adds the Go side of the RC v2 data path: a cgo binding that links libhipobj (the hipobj-rc-v2 client library, ROCm/hipObject#32) and drives it from Go. The library owns the RDMA data plane and the transfer state machine; this wrapper implements the four control-plane callbacks (PREPARE, READY request/response, CANCEL) with net/http and the gateway SigV4 signing, per the hipobj.h callback contract, so control traffic reaches the RC routes from #2391 through the standard request path.

Admission is fail-closed: a one-byte GET probe through the real API must succeed before any v2 transfer, capability headers are captured by the PREPARE callback, and admission turns negative on connection replacement or redirect until a fresh probe succeeds. The control dials pin the peer that answered the probe and classify transport loss by error identity, not timing.

cuobjtest gains a -v2 mode that exercises the wrapper end to end. The cgo binding is opt-in behind a hipobj build tag wired through HIPOBJ_ENABLED in the Makefile: the default build compiles a non-cgo stub whose entry points return a not-linked error, so the existing CI images keep building without libhipobj, and a hipobj-provisioned environment links the real client with HIPOBJ_INC_DIR/HIPOBJ_LIB_DIR.

Tests cover the control-plane wire exchanges against an in-process fake (pinned-peer enforcement, transport-loss classification, dial fallback ordering, budget split), the admission layer (probe fail-closed, connection-change and redirect invalidation), and the stub contract.

Note: stacked on #2391; the changes added by this PR are the commits after the #2391 head.


versity-github:

benmcclelland:
PR [#32](https://github.com/ROCm/hipObject/pull/32) only includes X-Amz-Rdma-Offset when offset is nonzero. Its normal GET/PUT calls use offset 0, so the header is absent. The local PREPARE route currently requires and parses that header unconditionally in routes_linux.go. Those requests will receive a 400.

The local C++ reference parser in v2_request.cpp already treats offset as optional, defaulting to zero. The Go route should do the same.

benmcclelland:
Add the rdma build tags requirements to these if we need to bypass these in github CI:
```
FAIL	github.com/versity/versitygw/rdma/rcroutes [build failed]
	github.com/versity/versitygw/rdma/rcserver		coverage: 0.0% of statements
FAIL	github.com/versity/versitygw/s3api [build failed]
```

some staticcheck errors:
```
  -: # github.com/versity/versitygw/rdma/rcroutes
  Error: rdma/rcroutes/semantic_linux.go:82:49: undefined: errRouteBadRequest (compile)
  -: # github.com/versity/versitygw/rdma/rcroutes [github.com/versity/versitygw/rdma/rcroutes.test]
  Error: rdma/rcroutes/semantic_linux_test.go:164:3: undefined: opsEmitter
  Error: rdma/rcroutes/semantic_linux.go:82:49: undefined: errRouteBadRequest
  Error: rdma/rcroutes/capabilities_linux_test.go:105:16: unknown field svc in struct literal of type Handler
  Error: rdma/rcroutes/capabilities_linux_test.go:105:26: unknown field be in struct literal of type Handler
  Error: rdma/rcroutes/capabilities_linux_test.go:105:46: unknown field mpMaxParts in struct literal of type Handler
  Error: rdma/rcroutes/capabilities_linux_test.go:112:17: unknown field svc in struct literal of type Handler
  Error: rdma/rcroutes/capabilities_linux_test.go:112:28: unknown field be in struct literal of type Handler
  Error: rdma/rcroutes/capabilities_linux_test.go:112:48: unknown field mpMaxParts in struct literal of type Handler
  Error: rdma/rcroutes/capabilities_linux_test.go:128:17: unknown field svc in struct literal of type Handler
  Error: rdma/rcroutes/capabilities_linux_test.go:128:33: unknown field be in struct literal of type Handler
  Error: rdma/rcroutes/capabilities_linux_test.go:128:33: too many errors (compile)
  Error: cmd/cuobjtest/main.go:83:30: this comparison is always true (SA4023)
  Error: 	cmd/cuobjtest/main.go:83:13: the lhs of the comparison is the 1st return value of this function call
  Error: 	cmd/cuobjtest/mode_v2_stub.go:24:6: github.com/versity/versitygw/cmd/cuobjtest.runV2Mode never returns a nil interface value
  Error: Process completed with exit code 1.
```

And I guess we need to make sure https://github.com/ROCm/hipObject/pull/32 is accepted to solidify this wire protocol contract.

potatogim:
Sorry for the slow reply, and thank you for the detailed analysis. routes_linux.go now parses x-amz-rdma-offset as optional and defaults the offset to zero, matching the C++ reference parser in v2_request.cpp. I also added a regression test covering the omitted, explicit-zero, and malformed cases: the first two reach the session server with the same zero offset, and only a malformed non-empty value rejects with 400.

Pushed in a8f6fdf2.

potatogim:
Thank you for the CI output as well, and sorry for the slow reply here too. The rcroutes test files and semantic_linux.go now carry the rdma build tag, so the github CI job skips them instead of failing on undefined symbols.

For the staticcheck findings, the stub build of runV2Mode no longer returns an error. main now gates the -v2 mode behind a hasV2Mode constant and exits directly when the hipobj backend is absent, and the error handling moved into the hipobj-backed implementation (runV2ModeErr). This removes the always-true comparison itself rather than suppressing it, and staticcheck comes back clean on both the default and the github,rdma tag sets.

go vet -tags=github ./... and the full rcroutes suite (with the rdma tag) both pass.

And I agree we need ROCm/hipObject#32 accepted to solidify the wire protocol contract.

PR #2399, opened 2026-09-15
fix: omit ObjectLockConfiguration Rule when there is no default retention

### LINKED ISSUE #2397: [Bug] - PutObjectRetention/versioning: keys ending in "/" (directory placeholders) return no usable version-id — breaks Nakivo immutability.
After escalation, Nakivo's development team confirmed that the empty element I originally reported is not the cause — their SDK parses that response correctly and detects Object Lock as enabled. I'm correcting this report with the actual root cause, which I've now reproduced directly against the gateway.

Actual root cause: objects whose key ends with / (zero-byte directory placeholder objects, created by Nakivo Backup) do not expose a usable version-id on a versioning-enabled bucket. This breaks per-object retention (PutObjectRetention) for AWS-SDK-based clients.

Reproduction (bucket with versioning + object lock enabled, versitygw v1.8.0, POSIX backend on ZFS):

list-object-versions on the folder key returns the literal string "null" as the version-id, while regular data objects in the same prefix return real version-ids:

json
{
"Key": "f9cc4668-cf1f-442b-a808-0e54d954da67/",
"VersionId": "null",
"IsLatest": true,
"Size": 0,
"ETag": ""d41d8cd98f00b204e9800998ecf8427e""
},
{
"Key": "f9cc4668-cf1f-442b-a808-0e54d954da67/ab9ca750-a8d0-4857-b850-a804f79802a3-001.raw",
"VersionId": "01M2KZR55M1NXY9E7QYT7WMCCK",
"IsLatest": true,
"Size": 5363236657
}

head-object on the same folder key returns no x-amz-version-id header at all (note ContentType: application/x-directory):

json
{
"AcceptRanges": "bytes",
"LastModified": "Wed, 16 Sep 2026 02:12:58 GMT",
"ContentLength": 0,
"ETag": ""d41d8cd98f00b204e9800998ecf8427e"",
"ContentType": "application/x-directory",
"Metadata": {}
}

So list-object-versions and head-object disagree for these keys: the listing reports "null", the metadata response omits the version-id entirely.

Why it breaks Nakivo (confirmed by their dev team): Nakivo applies object-lock retention per object and takes the version-id from the head-object (metadata) response, not from the listing. It processes the folder placeholder object first. Because that object has no usable version-id, Nakivo cannot form a valid PutObjectRetention request and stops before sending anything; the remaining objects in the recovery point are then skipped, so none get retention. Nakivo reports this is consistent across 10 lock attempts / 4 repositories / 2 buckets. This also explains why the gateway access log never shows any PutObjectRetention request — the client aborts before issuing it.

Expected behavior: objects with keys ending in / on a versioned bucket should return a valid version-id, both in list-object-versions and in the x-amz-version-id header of head-object, and the two should agree — so that lock/retention operations can target them. These placeholders are stored as real directories on the POSIX backend (ContentType: application/x-directory), which appears to be why they carry no version-id.

Likely the same interoperability area as https://github.com/versity/versitygw/issues/1864 (Veeam, object lock + no default retention).

++++++ old investigation ++++++

**Describe the bug**
When a bucket has Object Lock enabled but no default retention configured, GET /<bucket>?object-lock returns an ObjectLockConfiguration containing an empty <Rule></Rule> element:
<ObjectLockConfiguration>
  <ObjectLockEnabled>Enabled</ObjectLockEnabled>
  <Rule></Rule>
</ObjectLockConfiguration>

AWS S3 omits the <Rule> element entirely in this case (it is only present when a default retention rule exists). The empty <Rule></Rule> appears to break the AWS Java SDK v2 used by Nakivo Backup & Replication: immediately after the client parses this response, it issues a follow-up request with an empty host and empty headers, which fails, and the repository immutability setup is aborted.

Environment

versitygw v1.8.0, POSIX backend on ZFS (Ubuntu)
Bucket created with --object-lock-enabled-for-bucket, versioning enabled, no default bucket retention (as required by Nakivo)
Client: Nakivo Backup & Replication (uses aws-sdk-java/2.44.12)

**To Reproduce**
Create a bucket with object lock enabled and versioning, no default retention.
From an AWS Java SDK v2 client, add it as a Generic S3-Compatible immutable repository (Nakivo).
During the connection/verification phase the client calls GET /<bucket>?object-lock.

Gateway access log (with debug), showing the exact failure sequence:

01:31:17 | vgw | 200 | GET | /test3-immutabile | object-lock=
   RESPONSE BODY:
   <ObjectLockConfiguration>
     <ObjectLockEnabled>Enabled</ObjectLockEnabled>
     <Rule></Rule>
   </ObjectLockConfiguration>

   (next request from the client:)
   [URL]: https:///                 <-- empty host
   REQUEST HEADERS: (empty)
01:31:38 | vgw | 200 | GET | /
[INTERNAL ERROR]: Bad Gateway
The well-formed request just before this one has Host: ncloud-s3.nemesigroup.eu:8445 and a full User-Agent: aws-sdk-java/2.44.12 .... Only after parsing the object-lock response with the empty <Rule> does the client emit a request with an empty host.

**Expected behavior**
When no default retention is configured, GetObjectLockConfiguration should omit the <Rule> element (matching AWS S3), rather than returning an empty <Rule></Rule>:

xml
<ObjectLockConfiguration>
  <ObjectLockEnabled>Enabled</ObjectLockEnabled>
</ObjectLockConfiguration>

**Server**
v1.8.0

**Additional context**
**Impact**
This makes versitygw unusable as an immutable backup target for Nakivo Backup & Replication (and likely other AWS-Java-SDK-based backup products): the immutability verification fails, the repository is created as non-immutable (backupImmutabilitySupport=false), and recovery points are never locked. This is likely the same class of interoperability problem as #1864 (Veeam, same "object lock enabled, no default retention" scenario).

Additional verification (storage side is otherwise correct):

aws s3api get-bucket-versioning         -> Status: Enabled
aws s3api get-object-lock-configuration -> ObjectLockEnabled: Enabled
aws s3api put-object-retention (manual) -> success (COMPLIANCE)

So Object Lock itself works; only the XML shape of the empty-rule response is non-standard.

tunglambk:
Fixes #2397.

`ParseBucketLockConfigurationOutput` always built the response with `Rule` set, so a bucket with object lock enabled and no default retention answered `GET /<bucket>?object-lock` with an empty `<Rule></Rule>`. The response is marshalled with `encoding/xml` straight from the AWS SDK type, so a non-nil `Rule` with a nil `DefaultRetention` renders as the empty element the report shows:

```xml
<ObjectLockConfiguration>
  <ObjectLockEnabled>Enabled</ObjectLockEnabled>
  <Rule></Rule>
</ObjectLockConfiguration>
```

`Rule` is now only set when a default retention exists, which leaves the element out of the response the way AWS S3 does. The default retention case is unchanged.

`auth/object_lock_test.go` gains a table test over the stored JSON for both cases, asserting the marshalled XML: it reproduces the response body from the report before the change (`"…<Rule></Rule>" should not contain "<Rule>"`) and passes after it.

Validation:
- `go test -race -tags=github -timeout 120s ./auth/` passes.
- `go vet ./auth/` and `gofmt -s -l auth/` are clean.
- `go test ./s3api/... ./backend/...` passes except `TestVersioningDirIndependentOfWorkingDirectory` in `backend/posix`, which fails the same way on `main` here (xattr/sidecar subtests report `ExistingObjectIsDirectory`, unrelated to this change).


versity-github:

niksis02 (commented):
It would be nice to have a regression integration test coverring this case, where it expects the `Rule` to be `nil`.

tunglambk:
Added the regression case as `GetObjectLockConfiguration_no_default_retention` in `tests/integration/GetBucketLockConfiguration.go`, registered in `TestGetObjectLockConfiguration` and in `GetIntTests` so it also runs on its own.

It asserts that a lock-enabled bucket with no default retention returns `Rule == nil`, then configures a default retention and asserts the rule comes back non-nil, so a response that omits the rule unconditionally cannot pass. I also updated `s3api/controllers/bucket-get_test.go`, whose `TestS3ApiController_GetObjectLockConfiguration` case still expected the old empty non-nil `Rule` and was the failing test in the `general` workflow.

Ran against a local posix gateway: the new case fails on the pre-fix server with `expected nil object lock rule, instead got {DefaultRetention:<nil> ...}` and passes with the fix, alongside the existing `_success` and `_unset_config` cases. `gofmt -s`, `go vet`, `go build -tags github ./tests/...` and `go test ./s3api/controllers/ ./auth/` are clean.

masterzambo-hue:
Thanks for the quick fix on the empty <Rule> element — that's a real non-standard-response bug and the fix looks correct.

I want to flag a distinction, though, so this issue isn't closed prematurely. After escalating to Nakivo's development team, they confirmed that the empty <Rule> is not what blocks their immutability — their SDK parses that response correctly and detects Object Lock as enabled. So PR #2399, while a valid fix, may not by itself resolve the Nakivo failure described in this issue.

The blocker they identified (and which I've since reproduced directly) is a separate problem: objects whose key ends with / (zero-byte directory placeholders created by the backup client) return no usable version-id. On a versioning-enabled bucket:

list-object-versions reports "VersionId": "null" (the literal string) for the folder key, while real data objects in the same prefix return valid version-ids.
head-object on the folder key returns no x-amz-version-id header at all (ContentType: application/x-directory).

Nakivo takes the version-id from the head-object metadata to build per-object PutObjectRetention, processes the folder placeholder first, and — getting no version-id — stops before sending any retention request (confirmed consistent across 10 attempts / 4 repos / 2 buckets).

So there are two independent issues:

Empty <Rule> on GetObjectLockConfiguration — addressed by #2399. ✅
No usable version-id for keys ending in / on versioned buckets — still open, and this is the one that actually breaks Nakivo immutability.

Could #2399 be scoped to the <Rule> fix, and this issue kept open for the separate version-id problem (directory placeholders)? I can provide the full list-object-versions and head-object outputs. Happy to test any patch for the version-id case against my gateway.

niksis02 (approved):

niksis02:
> Thanks for the quick fix on the empty element — that's a real non-standard-response bug and the fix looks correct.
> 
> I want to flag a distinction, though, so this issue isn't closed prematurely. After escalating to Nakivo's development team, they confirmed that the empty is not what blocks their immutability — their SDK parses that response correctly and detects Object Lock as enabled. So PR #2399, while a valid fix, may not by itself resolve the Nakivo failure described in this issue.
> 
> The blocker they identified (and which I've since reproduced directly) is a separate problem: objects whose key ends with / (zero-byte directory placeholders created by the backup client) return no usable version-id. On a versioning-enabled bucket:
> 
> list-object-versions reports "VersionId": "null" (the literal string) for the folder key, while real data objects in the same prefix return valid version-ids. head-object on the folder key returns no x-amz-version-id header at all (ContentType: application/x-directory).
> 
> Nakivo takes the version-id from the head-object metadata to build per-object PutObjectRetention, processes the folder placeholder first, and — getting no version-id — stops before sending any retention request (confirmed consistent across 10 attempts / 4 repos / 2 buckets).
> 
> So there are two independent issues:
> 
> Empty on GetObjectLockConfiguration — addressed by #2399. ✅ No usable version-id for keys ending in / on versioned buckets — still open, and this is the one that actually breaks Nakivo immutability.
> 
> Could #2399 be scoped to the fix, and this issue kept open for the separate version-id problem (directory placeholders)? I can provide the full list-object-versions and head-object outputs. Happy to test any patch for the version-id case against my gateway.

@masterzambo-hue Thanks for diging this deep and finding the failure reason. Currently versioning is not supported for directory objects at all. Could you please open a bug report in github issues with some clear reproductions steps, so we could understand where exactly it fails and decide how deep we should go in adding versioning support for directory objects. As directory objects can't contain data in posix, it might not make a lot of sense to have versions of directories, that's why we need to understand the exact failure reason.

masterzambo-hue:
Heads-up before this merges: the linked issue #2397 has been re-scoped to a different problem (missing version-id for keys ending in /, which is the actual Nakivo blocker). This PR fixes the empty <Rule> element — a valid but separate fix. Nakivo's dev team confirmed the empty <Rule> is not what blocks them. Could the "Fixes #2397" be changed to "Relates to #2397" so the version-id issue stays open after merge? Details and reproduction are in #2397. Happy to test any version-id patch against a live Nakivo setup.

niksis02:
> Heads-up before this merges: the linked issue #2397 has been re-scoped to a different problem (missing version-id for keys ending in /, which is the actual Nakivo blocker). This PR fixes the empty element — a valid but separate fix. Nakivo's dev team confirmed the empty is not what blocks them. Could the "Fixes #2397" be changed to "Relates to #2397" so the version-id issue stays open after merge? Details and reproduction are in #2397. Happy to test any version-id patch against a live Nakivo setup.

The versionId support for directory objects (objects ending with /) has already been merged as a separate PR and has landed on current main. Here is the PR https://github.com/versity/versitygw/pull/2413. It will be included in the new release (1.9.0), but the release date is not clear yet.

For testing purposes you can manually compile the latest version from current main and test against a live Nakivo setup. Would be glad to hear feedback for this. If you find any new issues, feel free to open a new bug report issue ticket.

PR #2400, opened 2026-09-15
fix: strip `aws-chunked` from stored `Content-Encoding` on streaming uploads

### LINKED ISSUE #2398: [Bug] - `Content-Encoding: aws-chunked` is persisted as object metadata and replayed on GET
**Describe the bug**
When a client uploads with a trailing checksum, the request body is framed in AWS's chunked format and the request carries `Content-Encoding: aws-chunked`. versitygw decodes the framing correctly, but stores that header on the object. Every later GET replays it, and any client that decodes content codings strictly fails to read an object whose bytes are perfectly intact.

Real S3 treats `aws-chunked` as transport framing and removes the token before storing, keeping any other coding present in the same header.

**To Reproduce**
Using AWS CLI v2.23 or newer, whose `request_checksum_calculation` defaults to `when_supported`:

```
aws --endpoint-url $GATEWAY s3 cp ./somefile.csv s3://bucket/somefile.csv
aws --endpoint-url $GATEWAY s3api head-object --bucket bucket --key somefile.csv
```

Result:

```json
{
    "AcceptRanges": "bytes",
    "ContentLength": 17445,
    "ContentEncoding": "aws-chunked",
    "Metadata": {},
    "StorageClass": "STANDARD"
}
```

`ContentLength` is the true decoded length, so the stored object is correct — only the header is wrong.

Reading it back with any strictly decoding client then fails:

```
curl -v --output /dev/null $GATEWAY/bucket/somefile.csv     # with credentials
curl: (61) Unrecognized content encoding type
```

The same request against real S3 returns no `Content-Encoding` at all.

**Expected behavior**
`aws-chunked` should be stripped from `Content-Encoding` before the value reaches the backend, matching S3. If the header contains other codings, those should be preserved — S3 removes only the `aws-chunked` token.

**Server Version**
```
Version  : v1.7.0
Build    : 8a3dbcf97d996ea4955426e6b8a76eb96d2cd92a
BuildTime: 2026-07-15T02:34:54Z
Linux s3gateway--hyfkwjk-7d7fdf4d-7fxvk 6.6.150.1-1.azl3 #1 SMP PREEMPT_DYNAMIC Mon Aug 10 11:36:52 UTC 2026 x86_64 Linux
```

**Additional context**
**Impact**

Two independent reproductions, both on objects written days earlier:

- AWS CLI upload, later read by a PHP application through Guzzle — `cURL error 61: Unrecognized content encoding type` on an HTTP 200 response
- boto3 upload from an application, same header stored on every object it has ever written

The failure is delayed and silent. Browsers ignore an unrecognised content coding, and SDK clients return raw bytes without decoding, so a deployment whose read path is browser-only or SDK-only carries the header indefinitely with no symptom. It only surfaces when some later consumer — a script, an integration, a different library — decodes properly.

**Workaround**

Set `request_checksum_calculation = when_required` on the client (`AWS_REQUEST_CHECKSUM_CALCULATION` or the shared config file). No trailer is sent, so the body needs no chunked framing and no header is stored. This is client-side only and does nothing for objects already written.

Notably, PR #1868 applied this same setting inside the `s3proxy` backend for a related symptom. The equivalent problem on the storing backends is the one reported here.

loongyh:
Fixes #2398

## Problem

An object uploaded with a trailing checksum comes back carrying `Content-Encoding: aws-chunked`, and any client that decodes content codings strictly then fails to read it, although the stored bytes are intact:

```
$ curl -v --output /dev/null "$GATEWAY/bucket/file.csv"
< HTTP/1.1 200 OK
< Content-Encoding: aws-chunked
curl: (61) Unrecognized content encoding type
```

It surfaced on an azure-backed deployment as `cURL error 61` from a PHP application, on attachments the AWS CLI had uploaded a week earlier. Browsers had been rendering the same objects throughout, because they ignore a coding they do not recognise.

## Root cause

Since `request_checksum_calculation` began defaulting to `when_supported`, the SDKs send a CRC32 as a trailer. A trailer cannot precede the body, so the request is framed in the aws-chunked format and announced with `Content-Encoding: aws-chunked`.

For a [SigV4 streaming upload](https://docs.aws.amazon.com/AmazonS3/latest/dev/API/sigv4-streaming.html) that header describes the framing rather than the object, and `s3api/utils/unsigned-chunk-reader.go` consumes the framing while reading the body. S3 removes the token before storing the header, keeping any other coding the client sent. The gateway stored the value verbatim and replayed it on every later GET and HEAD.

`s3proxy` is the one backend that does not exhibit this, because the upstream S3 strips the token itself, which is likely why it has gone unnoticed: the backend validated against genuine S3 is the one that cannot show it.

## Only PutObject needs this

The token is only transport when the request was actually framed that way. On any other request `aws-chunked` is a value the client chose, and S3 stores it — so the strip is conditional on `x-amz-content-sha256` naming a streaming payload type, not applied wherever the header appears.

That confines the change to one controller, because a streaming payload type is rejected in `authentication.go` for every route registered with `streamBody` false:

| handler | takes `Content-Encoding` from | can carry aws-chunked framing |
| --- | --- | --- |
| `PutObject` | request header, `streamBody` true | yes — fixed here |
| `UploadPart` | not read at all | yes, but nothing to store |
| `CopyObject` | request header, `streamBody` false | no |
| `CreateMultipartUpload` | request header, `streamBody` false | no |
| `POSTObject` | the `content-encoding` form field | no — `multipart/form-data` |

## Also fixed

`UNSIGNED-PAYLOAD` combined with `Content-Encoding: aws-chunked` is a contradiction — the payload declares the body is sent as-is, the header says it is framed — and S3 rejects it:

```
400 InvalidArgument: "aws-chunked encoding is not supported when x-amz-content-sha256 UNSIGNED-PAYLOAD is supplied"
```

The gateway accepted it. The check sits in the authentication middleware beside the existing payload-type validation, so it covers `PutObject` and `UploadPart` alike.

## Verification

A posix-backed gateway, same file each time:

| case | `main` | this branch |
| --- | --- | --- |
| chunked upload, then `head-object` | `"ContentEncoding": "aws-chunked"` | field absent |
| hex payload with `Content-Encoding: aws-chunked` | stored as sent | stored as sent |
| `UNSIGNED-PAYLOAD` with `Content-Encoding: aws-chunked` | 200 | 400 `InvalidArgument` |

The controller test fails if the call site is reverted:

```
object-put_test.go: expected content encoding "gzip", got "aws-chunked,gzip"
```

## Changes

- `s3api/utils/utils.go` — `ParseContentEncoding`, dropping the token only for streaming payload types and keeping any other codings and their order, plus `HasAwsChunkedEncoding` for the middleware check.
- `s3api/controllers/object-put.go` — `PutObject` uses it. `CopyObject` is untouched, per the table above.
- `s3api/middlewares/authentication.go`, `s3err/invalid-argument.go` — the `UNSIGNED-PAYLOAD` rejection and its `InvalidArgAwsChunkedUnsignedPayload` code.
- `s3api/utils/utils_test.go`, `s3api/controllers/object-put_test.go` — the parsing, including ordering, case, a coding that merely contains the token, and both sides of the streaming condition.
- `tests/test_rest_chunked.sh`, `tests/test_rest_put_object.sh` — a REST test per case in the table.
- `tests/rest_scripts/put_object.sh`, `tests/drivers/head_object/head_object_rest.sh`, `tests/drivers/put_object/put_object_rest.sh` — `CONTENT_ENCODING` support and a `check_content_encoding` driver, needed by those tests.

`go test ./s3api/...` passes; `go vet` and `gofmt` clean; `shellcheck` reports nothing new.

## Adjacent, deliberately not fixed here

1. This changes newly written objects only. Objects already stored keep the header until something rewrites it — on the azure backend that can be done in place with Set Blob HTTP Headers, since only the metadata was ever wrong.


versity-github:

niksis02 (changes requested):
Blindly stripping `aws-chunked` from `Content-Encoding` solves one AWS compatibility issue, but introduces another.

AWS documents that `aws-chunked` is removed from the stored `Content-Encoding`, but this behavior is described specifically in the context of [SigV4 streaming uploads](https://docs.aws.amazon.com/AmazonS3/latest/dev/API/sigv4-streaming.html). This means `aws-chunked` should only be stripped for streaming uploads(I have already verified against S3).

For regular PUT requests using `x-amz-content-sha256: <hex>`, it should not be removed. Otherwise, the following request will not behave like S3:

```bash
aws s3api put-object --bucket test --key obj --content-encoding aws-chunked
```

In this case, `aws-chunked` would be stripped unconditionally and would not be stored as the object's `Content-Encoding`, which differs from AWS S3 behavior.

Preferably, it would be good to have integration tests covering these cases.

niksis02 (commented):
There's another important case to cover in this context. The `UNSIGNED-PAYLOAD` and `Content-Encoding: aws-chunked` combo is rejected with and error: 
```
400 InvalidArgument: "aws-chunked encoding is not supported when x-amz-content-sha256 UNSIGNED-PAYLOAD is supplied"
```

loongyh:
Thanks for the review, addressed them in the new commit.

s3api/middlewares/authentication.go
```diff
@@ -132,6 +132,11 @@ func VerifyV4Signature(root RootUserConfig, iam auth.IAMService, region string,
 		if !streamBody && utils.IsStreamingPayload(hashPayload) {
 			return s3err.GetAPIError(s3err.ErrInvalidSHA256PayloadUsage)
 		}
+		// aws-chunked frames the body, so it contradicts an unsigned payload,
+		// which declares the body is sent as-is
+		if utils.IsUnsignedPaylod(hashPayload) && utils.HasAwsChunkedEncoding(ctx.Get("Content-Encoding")) {
+			return s3err.GetInvalidArgumentErr(s3err.InvalidArgAwsChunkedUnsignedPayload, hashPayload)
```
niksis02, line 138: S3 returns the following error for this case:
```xml
<Error>
  <Code>InvalidArgument</Code>
  <Message>aws-chunked encoding is not supported when x-amz-content-sha256 UNSIGNED-PAYLOAD is supplied.</Message>
  <ArgumentName>Content-Encoding</ArgumentName>
  <ArgumentValue>aws-chunked</ArgumentValue>
  <RequestId>0J6W2XDW4WDX3CC2</RequestId>
  <HostId>XJRCbf3RQX14Wsg4h+C5vYBXnVZan+EkkRxsLHgtEZiB9iLMEsH9ivpO8+A5aU5D+6ThFl2td74=</HostId>
</Error>
```

Note that `ArgumentName` is `Content-Encoding` not `x-amz-content-sha256` and `ArgumentValue` is `aws-chunked`.

Note that when `Content-Encoding: gzip,aws-chunked` is sent, `<ArgumentValue>` comes back as just `aws-chunked`, **not** the full header value

niksis02 (changes requested):
I'd also like to see some integration tests covering the implementation, especially:
1. `x-amz-content-sha256: UNSIGNED-PAYLOAD` with `Content-Encoding: aws-chunked`, `Content-Encoding: aws-chunked,gzip` ...
2. A successful streaming upload with `Content-Encoding: aws-chunked,gzip`, where `aws-chunked`-only is stripped. `Content-Encoding: aws-chunked` upload, where the final object doesn't have any `Content-Encoding` storred.
3. A regular PUT with `Content-Encoding: aws-chunked`, where the final object has the content encoding storred.

Note that by `integration tests`, I mean the go sdk and raw http client tests under `tests/integration` directory.

niksis02:
A follow up PR has been created for this https://github.com/versity/versitygw/pull/2427

PR #2404, opened 2026-09-15
fix file hangs if it was removed; and can't be re-uploaded correctly (if file name is the same)

iRioxe:
Hello!

Thank you very much for a great project you've done!

Please take a look at my small PR. You might find it useful.

I've faced with an issue with a file removing and that might be a fix for this.

Issue description (I've used versitygw backend and your UI for diagnostic):

- I've uploaded (multipart mode) a file by using versitygw backend and than tried to remove it
- somehow I got the situation, when file "hangs" by name. What I mean by this, from the UI perspective looks like this: It is only visible inside "version viewer" and it's not visible in non-version viewer. And the only option you can do with it is to press on a bucket button (unmark it as deleted or so). But if I pressed it, than file is just gone from both viewers, BUT: if I try to upload it again from the start or try to "continue uploading" (for example, by parts)  - than the file still "hangs" as described above, i.e. it's visible only in version viewer only and etc.   (even if uploading finished correctly)
- if I just change any character in a file name (the same file content, but name is different) - everything works perfect.

Than I've concluded that somehow some meta-information or so might be not fully removable.

Additional information, that might be usefull as well:
- bucket with enabled versioning (--versioning-dir) with sidecar meta-data
- If an object is deleted without a versionId, a delete-marker sidecar file is written but the data file is left in place
- if the same key is later re-uploaded (PUT or multipart complete), the new      data/metadata are written, but the delete-marker sidecar file persists. The object therefore appears both as "deleted" and as having content. probably
- When the current version is later deleted by versionId, DeleteObject      removes the data file, finds no older versions to restore, and returns      success but leaves the sidecar directory behind.  The object path is then      stuck: no data file, but sidecar metadata (including delete-marker) remains.


versity-github:

backend/posix/posix.go
```diff
@@ -2666,6 +2666,7 @@ func (p *Posix) CompleteMultipartUploadWithCopy(ctx context.Context, input *s3.C
 		// version's path-based metadata into this (new) version's sidecar.
 		_ = p.meta.DeleteAttribute(bucket, object, objectLegalHoldKey)
 		_ = p.meta.DeleteAttribute(bucket, object, objectRetentionKey)
+		_ = p.meta.DeleteAttribute(bucket, object, deleteMarkerKey)
```
benmcclelland: This cleanup is still skipped in the failure state described by the PR. When the data file has been removed but the sidecar remains, os.Stat(objname) returns ENOENT, so the surrounding err == nil branch is not entered and the stale delete-marker attribute survives the multipart re-upload.

Would need to move this cleanup outside the existing-object snapshot branch, after createObjVersion when applicable. The same cleanup is also needed on the ordinary PutObject publish path.

backend/posix/posix.go
```diff
@@ -4165,6 +4166,7 @@ func (p *Posix) snapshotObjVersion(bucket, key string, vStatus types.BucketVersi
 		// bleed into the new version.
 		_ = p.meta.DeleteAttribute(bucket, key, objectLegalHoldKey)
 		_ = p.meta.DeleteAttribute(bucket, key, objectRetentionKey)
+		_ = p.meta.DeleteAttribute(bucket, key, deleteMarkerKey)
```
benmcclelland: snapshotObjVersion returns before reaching this line when the live data file is missing. That is the exact orphaned-sidecar state this change is intended to repair.

Also, when versioning is suspended or the current object has a null/missing version ID, the function can skip createObjVersion and therefore skip this cleanup as well. The delete-marker cleanup should be independent of whether an older data version was successfully archived.

backend/posix/posix.go
```diff
@@ -4840,12 +4842,13 @@ func (p *Posix) DeleteObject(ctx context.Context, input *s3.DeleteObjectInput) (
...
 				}
 				err = os.Remove(objpath)
-				if err != nil {
+				if err != nil && !errors.Is(err, fs.ErrNotExist) && !isErrNotDir(err) {
 					return nil, fmt.Errorf("remove obj version: %w", err)
 				}
 
 				ents, err := os.ReadDir(versionPath)
 				if errors.Is(err, fs.ErrNotExist) {
+					_ = p.meta.DeleteAttributes(bucket, object)
```
benmcclelland: This cleanup appropriate for the case where the current version was removed and no older versions remain. However, the error is ignored, so a sidecar removal failure can still return a successful DeleteObject response while leaving the object stuck.

Could this error be propagated, except for the metadata backend’s expected “nothing to remove” case?

backend/posix/posix.go
```diff
@@ -4857,6 +4860,7 @@ func (p *Posix) DeleteObject(ctx context.Context, input *s3.DeleteObjectInput) (
 				}
 
 				if len(ents) == 0 {
+					_ = p.meta.DeleteAttributes(bucket, object)
```
benmcclelland: Same concern here: ignoring a real metadata cleanup error can leave stale sidecar state while reporting successful deletion. Consider returning the cleanup error so the caller knows the object was not fully removed.

benmcclelland (changes requested):

benmcclelland:
Thank you for the pull request and for the detailed problem description. I appreciate you taking the time to investigate this.

I believe the proposed fix needs some additional changes to fully handle the orphaned sidecar metadata case. If you would like, we can convert this into a GitHub issue and take ownership of completing the fix and adding the necessary regression tests if you want to hand this off to our team.

iRioxe:
> Thank you for the pull request and for the detailed problem description. I appreciate you taking the time to investigate this.
> 
> I believe the proposed fix needs some additional changes to fully handle the orphaned sidecar metadata case. If you would like, we can convert this into a GitHub issue and take ownership of completing the fix and adding the necessary regression tests if you want to hand this off to our team.

Hi!

Thank you for your feedback. I've tried to add additional changes and pushed it into the last commit. If it's not all the changes needed or wrong, than, please, let's convert it into an issue and continue as you suggested.

Best regards

benmcclelland:
This looks good! I added a test and merged this in #2411

PR #2405, opened 2026-09-15
Hidden actual error message cause of CORS check Error on PUT/<bucket-name>?object-lock=

iRioxe:
Hello!

Thank you very much for a great project you've done!

Please take a look at my small PR. You might find it useful.

I've faced with an issue with using PUT method on object-lock. Frontend receives:

- HTTP 400 from Versity GW on PUT /113?object-lock= (S3 PutObjectLockConfiguration). Where 113 is just my bucket name.

- and the browser then reports a CORS error because the 400 response has no Access-Control-Allow-Origin header, so the browser hides the actual error message.

Inside a file s3api/router.go I found out that `applyBucketCORS` is placed after auth/checksum and for some another routes it goes before. As a short term solution and for code consistency I suggest to move applyBucketCORS to run right after BucketObjectNameValidator().


versity-github:

benmcclelland:
This looks logically correct. However, Moving applyBucketCORS before authentication causes every unauthenticated request with an Origin header to perform a backend GetBucketCors lookup before the request is authorized. When a fallback origin is configured, this lookup can also occur without an Origin header. Since this may involve filesystem or other backend I/O, an attacker can use unauthenticated requests to consume backend resources. So I just need to look into if there is a good way to lock this down, or if this is just whats required to get the expected behavior.

benmcclelland (approved):

PR #2413, opened 2026-09-18
feat: add versioning support for directory objects in posix

niksis02:
Directory objects (keys ending in `/`) are now versioned in versioning-enabled posix buckets like regular objects. The directory is kept in place across versions and carries the `version-id` and `delete-marker` attributes, while older versions are stored as empty files in the versioning directory. Only the attributes listed in `dirObjectAttrs` are copied into versions, replaced on a new `PutObject`, and restored when the latest version is deleted, so attributes the directory carries as a parent of other objects are left untouched.

`PutObject` on a directory object now snapshots the current version, starts the new version with a clean attribute set and returns a `VersionId`. `DeleteObject` creates delete markers for directory objects and, when deleting the latest version, restores the previous one onto the directory, or removes the directory (or keeps it as a plain parent of its children) when no versions are left. `ListObjectVersions` lists directory object versions and delete markers and uses the key with the trailing slash for `NextKeyMarker` and marker comparisons, `ListObjects` hides directory delete markers and `GetObject` returns the directory object version id.

`PutObject` on directory objects now applies the object lock legal hold and retention headers, and `CopyObject` and `UploadPartCopy` accept a directory object as a 0-byte copy source.

The versioning integration tests are extended to run their scenarios for directory objects as well, with new tests for directory-object-specific behavior.


backend/posix/posix.go
```diff
@@ -5033,6 +5237,93 @@ func (p *Posix) DeleteObject(ctx context.Context, input *s3.DeleteObjectInput) (
 	return &s3.DeleteObjectOutput{}, nil
 }
 
+// latestObjVersion returns the entry of the version, among the version
+// directory entries, that becomes the latest one when the latest version
+// of the object is deleted
+func latestObjVersion(ents []fs.DirEntry) fs.DirEntry {
+	return ents[len(ents)-1]
+}
```
benmcclelland: null sorts after ULID filenames, so after suspend -> create null version -> re-enable -> create versions, deleting the latest directory version restores null rather than the immediately preceding ULID version. Select by version timestamp/order, with explicit handling for null.

backend/posix/posix.go
```diff
@@ -5241,6 +5532,18 @@ func (p *Posix) GetObject(ctx context.Context, input *s3.GetObjectInput) (*s3.Ge
...
+	// If versioning is configured get the object versionId
+	if p.versioningEnabled() && versionId == "" {
+		vId, err := p.meta.RetrieveAttribute(nil, bucket, object, versionIdKey)
+		if errors.Is(err, meta.ErrNoSuchKey) {
+			versionId = nullVersionId
+		} else if err != nil {
+			return nil, err
+		}
+
+		versionId = string(vId)
```
benmcclelland: Directory objects without versionIdKey return an empty version ID instead of null since this overwrites the above "versionId = nullVersionId" for the ErrNoSuchKey case.

benmcclelland (changes requested):

benmcclelland (approved):

PR #2415, opened 2026-09-19
fix(s3api): reject a PUT whose body ends before Content-Length

### LINKED ISSUE #2414: [Bug] - posix: aborted (short) non-multipart PUT is published as a complete object
## Summary

With the `posix` backend, a `PutObject` whose body ends **before** `Content-Length` bytes have
arrived is truncated to the number of bytes actually received and then **linked into place as a
complete object**. Real S3 rejects this with `IncompleteBody` and the object never becomes visible.

This is reproducible through the Elasticsearch repository analysis API, which explicitly tests this
case. It is **not** a leak — nothing is left behind on disk — it is a *visibility* problem: an
upload that was aborted becomes a readable object.

## Environment

| | |
|---|---|
| versitygw | **v1.8.0** (latest release at time of testing) |
| Backend | `posix`, single gateway process, TLS enabled |
| Filesystem | XFS on a dedicated LV, Linux (RHEL 9 family) |
| Elasticsearch | 9.4.6, `repository-s3`, 3-node cluster |

## What Elasticsearch reports

```
repository_verification_exception:
  [backup] upload of blob was aborted, but blob was erroneously found by at least one node
  blob analysis [... length=65536 ... abortWrite=true]
```

The failing operation is a **65,536-byte single PUT** — not a multipart upload.

## Scope — what is and is not affected

I ran the same analysis again with `rare_action_probability=0`, which skips the abort operations
and exercises everything else. **It passed**, completing a full 34-minute run:

| | measured |
|---|---|
| write | 10 GB, `total_elapsed` 9.48m (97.6% of it ES's own `max_snapshot_bytes_per_sec` throttle) |
| read | 6,000 blobs / 23.9 GB |
| `read.total_wait` | **23.49 ms total across 6,000 blobs** |
| `read.max_wait` | 591.99 µs |
| largest blob | `length=2147483648` (2 GB) — multipart upload path exercised and correct |
| cleanup | `delete_elapsed: 447 ms`, no `temp-analysis-*` left behind |

So S3 semantics, multipart, performance, stability and cleanup are all fine. **The only thing that
fails is the aborted-write case.**

I also checked the staging area after the failing run:

```
$ find <bucket>/.sgwtmp -mindepth 1 -printf '%T+ %10s %p\n'
2026-09-18+11:29:39   6  <bucket>/.sgwtmp/multipart
```

Only the empty `multipart` directory — **no orphaned files, `du` 0**. That is consistent with the
failing operation being a single PUT, which never goes through `.sgwtmp` at all.

## Suspected code path

`backend/posix/posix.go`, in `PutObjectWithPostFunc` (v1.8.0, around L4325–4351):

- `io.CopyBuffer(f, rdr, buf)` is called and **the byte count is discarded** (`_`).
- `tmpfile.size` is initialised to `Content-Length` and decremented on every `Write`
  (`backend/posix/otmpfile_common.go` L30–52), so after the copy it holds the **shortfall**.
- The next block treats a non-zero shortfall as over-allocation and shrinks the file to what was
  actually written, then proceeds to link it into place:

  ```go
  if f.size > 0 {
      objsize -= f.size
      if err := f.File().Truncate(objsize); err != nil { ... }
  }
  ```

The comment explains this is meant for epilogue bytes after a browser POST-Object multipart final
boundary, where `Content-Length` legitimately exceeds the object bytes. The problem is that a
**short/aborted body reaches the same branch**, because nothing compares bytes received against
`Content-Length` — and `io.Copy` returns a `nil` error on a clean EOF, so the early-return above it
is not taken.

Note that `ErrIncompleteBody` already exists and *is* enforced on the chunked-upload paths
(`s3api/utils/signed-chunk-reader.go`, `s3api/utils/unsigned-chunk-reader.go`). The plain
(non-chunked) PUT path in the posix backend appears to be the gap.

⚠️ To be clear about confidence: the **symptom is measured**, the **code path above is my reading of
v1.8.0 source that fits the symptom** — I have not instrumented the gateway to confirm it directly.

## Reproduction

Via Elasticsearch (the same way #1435 was found):

```
POST /_snapshot/<repo>/_analyze?blob_count=2000&max_blob_size=2gb&max_total_data_size=10gb&register_operation_count=100&timeout=1800s
```

Fails as above. Adding `&rare_action_probability=0` makes the same run pass, which isolates it to
the abort operations.

Without Elasticsearch, the equivalent shape is: issue a `PutObject` with `Content-Length: N`, send
fewer than `N` bytes, then close the connection. Expected: the request fails and no object becomes
visible. Observed: an object of the truncated length is readable afterwards.

## Why this matters for this use case

Elasticsearch requires `s3`-type repositories to be fully S3-compatible and documents **risk of
snapshot data loss** otherwise, so this single failure is enough to block adoption — the aborted
write is precisely the case a backup has to get right *during* a failure. Happy to test a patch
against the same analysis run if that is useful.

Davvyy:
Fixes #2414.

## The problem

With a plain (non-`aws-chunked`) request body, a `PutObject` whose body ends **before**
`Content-Length` bytes have arrived is committed as a complete — but shorter — object. Real S3
rejects this with `IncompleteBody` and the object never becomes visible.

Nothing on that path compares bytes received against `Content-Length`:

- fasthttp's request stream reports a connection closed mid-body as a plain `io.EOF`. The
  conversion to `io.ErrUnexpectedEOF` exists only on the `ContentLength() == -1` (chunked
  transfer-encoding) branch, so a short fixed-length body looks like a clean end of stream.
- The authentication middleware wraps the body only for signed payloads (`HashReader`) or
  `aws-chunked` payloads (the chunk readers). For `UNSIGNED-PAYLOAD` it passes the stream through
  untouched.
- `io.Copy`/`io.CopyBuffer` treat `io.EOF` as a normal end, and the posix backend then shrinks the
  file to what was actually written and links it into place. That `Truncate` is intended for the
  epilogue bytes of a browser POST-Object, where `Content-Length` legitimately exceeds the object
  bytes — a short body reaches the same branch because there is nothing to tell the two apart.

`ErrIncompleteBody` already exists and is enforced by the chunk readers
(`signed-chunk-reader.go`, `unsigned-chunk-reader.go`). The plain body path is the gap.

## The fix

A small `ContentLengthReader` in `s3api/utils` that turns an EOF arriving before `Content-Length`
bytes into `ErrIncompleteBody`, applied in `PutObject` and `UploadPart` **only when the body is not
already an `aws-chunked` reader** (detected via the existing `middlewares.ChecksumReader`
interface, which both chunk readers implement).

It is in the controller rather than in a backend so every backend is covered, and so POST-Object is
untouched — that handler is separate, and its `ContentLength` is an upper bound (request length
minus boundaries), so a byte-count check there would break browser form uploads.

### Measured

All of the following were run against builds of this branch and of its parent commit
(`1246a08`), `posix` backend on a Docker volume, Elasticsearch 9.4.6 two-node cluster.

| check | parent `1246a08` | this branch |
|---|---|---|
| truncated PUT published as an object — 30 `(Content-Length, bytes sent)` combinations over HTTPS, 3 runs | 18 / 19 / 19 of 30 | **0 / 0 / 0** |
| same, plain HTTP, 3 runs | 9 / 5 / 7 of 30 | **0 / 0 / 0** |
| ES `_analyze?blob_count=100&max_blob_size=1mb&rare_action_probability=0.5`, TLS, ×5 | 5/5 fail | **5/5 pass** |
| same over plain HTTP, i.e. the signed-chunked path, ×5 | — | **5/5 pass** |
| ES `_analyze?blob_count=2000&max_blob_size=256mb&max_total_data_size=4gb&register_operation_count=100` | fail | **pass**, 264 s |
| complete-object round trips 1 B – 16 MB, SHA-256 compared, 12 objects | ok | **12/12 ok** |

`go test ./s3api/utils/ ./s3api/controllers/ ./s3api/middlewares/ ./backend/posix/` passes.
(`./s3api` itself does not link in my environment because the RDMA static libraries are not
present; that failure is identical on an unpatched tree.)

The unit test covers complete bodies, empty bodies, truncated bodies — including an EOF delivered
together with the final bytes, and one byte per `Read` — a body longer than announced, and
pass-through of a non-EOF error.

### Reasoned, not exercised

I have **not** run these paths; the entries below are from reading the code and may be wrong.

| path | reasoning |
|---|---|
| POST-Object form uploads | different handler (`bucket-post.go`), which builds its own reader and is not touched by this change |
| presigned PUT | `presign-auth.go` leaves the body unwrapped, so the new check now applies to it as well — intended, but not tested here |
| signed plain payload (`x-amz-content-sha256` = real hash) | the inner `HashReader` should return `ContentSHA256Mismatch` on a short body, a non-EOF error, which the wrapper passes through unchanged |
| Content-MD5 header | same as above (`BadDigest`) |
| HTTP chunked TE without Content-Length | already rejected earlier with `ErrMissingContentLength` |
| non-posix backends (azure, s3proxy) | the reader error may be wrapped by the SDK and surface as 500 rather than 400 |

## Note on scope

`UploadPart` has the same missing check and is fixed here. Note its posix path also lacks the
`Truncate` that `PutObject` has, while `Fallocate` extends the file to the declared size — so a
short part would have been committed at full length with a zero tail. This change prevents that by
rejecting the request; I have not changed the posix code itself.

Happy to adjust the placement or the error if you would rather have this live somewhere else.


versity-github:

s3api/controllers/object-put.go
```diff
@@ -307,6 +308,12 @@ func (c S3ApiController) UploadPart(ctx fiber.Ctx) (*Response, error) {
 	} else {
 		body = bytes.NewReader([]byte{})
 	}
+	// aws-chunked bodies are framed and length-checked by the chunk readers
+	// (the ones implementing middlewares.ChecksumReader). A plain body has
+	// nothing but Content-Length to tell a finished upload from an aborted one.
+	if _, chunked := body.(middlewares.ChecksumReader); !chunked {
+		body = utils.NewContentLengthReader(body, contentLength)
```
niksis02: This introduces a regression for plain (non-aws-chunked) PutObject/UploadPart requests that carry `X-Amz-Decoded-Content-Length`.

The controller swaps `Content-Length` for `X-Amz-Decoded-Content-Length` whenever that header is present, regardless of payload type, and that value now feeds `ContentLengthReader`. So a complete plain upload whose decoded header is larger than `Content-Length` is now rejected with `IncompleteBody`. Before this change, the same request succeeded.

AWS S3 ignores `X-Amz-Decoded-Content-Length` on plain bodies: it returns 200 and stores `Content-Length` bytes (checked against S3 directly). The length check needs to use `Content-Length` for plain bodies. The decoded length only applies to aws-chunked payloads.

niksis02 (changes requested):
Nice catch and thank you for the fix PR. Overall looks good, just a small comment to fix.

Also consider adding integration tests both for regular PUT and streaming uploads with terminated/aborted requestes.

Davvyy:
Thanks — you are right, and the reproduction is exactly as you describe.

`contentLengthStr` is replaced by `X-Amz-Decoded-Content-Length` unconditionally:

```go
decodedLength := ctx.Get("X-Amz-Decoded-Content-Length")
if decodedLength != "" {
    contentLengthStr = decodedLength
}
```

so my check was comparing the raw body against the *decoded* size. Pushed two commits:

**1. The fix** — the length check now reads the raw `Content-Length` header instead of the possibly-replaced `contentLength` variable:

```go
if _, chunked := body.(middlewares.ChecksumReader); !chunked {
    if raw, cerr := strconv.ParseInt(ctx.Get("Content-Length"), 10, 64); cerr == nil && raw > 0 {
        body = utils.NewContentLengthReader(body, raw)
    }
}
```

If the header is missing or unparseable the body is left unwrapped — an `aws-chunked` request without `Content-Length` is already rejected earlier with `ErrMissingContentLength`, so there is nothing to check at this point.

**2. Integration tests** — three, in `tests/integration/PutObject.go`:

| test | what it pins |
|---|---|
| `PutObject_aborted_plain_body` | plain PUT whose body ends early leaves no object |
| `PutObject_aborted_streaming_body` | same for an `aws-chunked` upload |
| `PutObject_plain_body_with_decoded_length` | **the regression you found** — a complete plain PUT carrying `X-Amz-Decoded-Content-Length` still returns 200 and stores `Content-Length` bytes |

The abort is driven by a reader that fails partway, so the transport tears the connection down mid-body — that is what a client that dies or cancels looks like on the wire. Sending a short body with an ordinary reader would not reproduce it: `net/http` reports the mismatch itself before anything reaches the gateway.

The third test fails without the fix in commit 1, so it guards the exact case you raised.

`gofmt`, `go vet ./s3api/... ./tests/integration/` and `go test ./s3api/utils/ ./s3api/controllers/` are clean. (`go build ./...` fails on `plugins/noop` — `function main is undeclared in the main package` — but that is identical on an unpatched tree here.)

I have not been able to run the integration suite end to end myself; if CI can be approved I would rather it verify them than take my word for it.

tests/integration/PutObject.go
```diff
@@ -1330,3 +1330,166 @@ func PutObject_false_negative_object_names(s *S3Conf) error {
...
+			Key:    &obj,
+		})
+		cancel()
+		if err == nil {
+			return fmt.Errorf("expected the aborted upload to leave no object, but %v exists", obj)
+		}
+		return nil
+	})
+}
+
```
niksis02: This doesn't exercise the streaming path. `putObjectAborted` always sends `x-amz-content-sha256: UNSIGNED-PAYLOAD`, and the gateway decides aws-chunked from that header alone, so the request takes the same plain path as the test above it.

Covering it properly needs `STREAMING-UNSIGNED-PAYLOAD-TRAILER`, an `x-amz-trailer` header and a chunk-framed body that stops mid-chunk. The code path is fine: an aborted aws-chunked upload sent by hand returns `IncompleteBody` and leaves no object. Only the test is a gap.

Conceptually, I guess this test should live in `tests/integration/unsigned_streaming_payload_trailer.go`

niksis02 (changes requested):

Davvyy:
You are right, and thank you — the test was covering nothing.

`putObjectAborted` hardcodes `x-amz-content-sha256: UNSIGNED-PAYLOAD`, so `Content-Encoding: aws-chunked` on top of it changed nothing: the request took the plain path and `PutObject_aborted_streaming_body` was a duplicate of the test above it.

Dropped it. Added `UnsignedStreamingPayloadTrailer_aborted_connection` in `tests/integration/unsigned_streaming_payload_trailer.go`, where you suggested it belongs. It sends `STREAMING-UNSIGNED-PAYLOAD-TRAILER` with `x-amz-trailer`, announces the full framed payload in `Content-Length`, and drops the connection a quarter of the way into the first chunk.

`UnsignedStreamingPayloadTrailer_incomplete_body` already covers malformed and truncated framing, but every case there is a **complete** request — short body, `Content-Length` agrees, the chunk reader rejects what it parses. The new one is the other shape: the framing is valid and the bytes simply stop arriving. That is what a client that dies or cancels looks like on the wire, and it is the shape the plain path got wrong.

## Verified by running it

Built the gateway and ran the tests against a posix backend, with the patch in three states:

| test | base (no patch) | first commit only<br>(checked against the decoded length) | as submitted |
|---|---|---|---|
| `PutObject_aborted_plain_body` | **FAIL** — object exists | PASS | PASS |
| `PutObject_plain_body_with_decoded_length` | PASS | **FAIL** — 400 | PASS |
| `UnsignedStreamingPayloadTrailer_aborted_connection` | PASS | PASS | PASS |
| `UnsignedStreamingPayloadTrailer_incomplete_body` | PASS | PASS | PASS |

So, to be precise about what the new test is worth: **it does not catch the bug** — you already said the chunk readers handle this, and the run confirms it. It is a regression guard for that path, and it now actually exercises it rather than claiming to.

The first two rows are the ones that matter, and both fail in exactly the state they are meant to catch.

`go vet ./tests/integration/... ./s3api/...` is clean and `go test ./s3api/utils/ ./s3api/controllers/` passes. `gofmt` is clean on the committed blobs — my working tree is a Windows checkout with CRLF, so `gofmt -l` flags all 544 files there; I checked the staged content instead.

benmcclelland (approved):

PR #2422, opened 2026-09-20
feat: add webui concurrent uploads with progress tracking

benmcclelland:
Queue uploads with a concurrency limit, track per-file and aggregate progress, and preserve batch bucket/prefix context.

Limit concurrent uploads to three. Display aggregate byte progress as a percentage and bytesDone / bytesTotal, completion and failure totals, and the number of active uploads in the message box. Display Queued, Uploading, and Failed files with individual progress bars; completed uploads remain in the aggregate totals without filling the live file list.

Allow queued and active uploads to be cancelled from the progress widget. Remove queued uploads immediately, abort active object and multipart part requests, and abort a multipart upload on the server after its upload ID is available.


webui/web/explorer.html
```diff
@@ -2969,35 +2969,193 @@ <h2 class="text-xl font-semibold text-charcoal">Multipart Uploads</h2>
...
+          cancelUpload(Number(button.dataset.batchId), Number(button.dataset.fileIndex));
+        };
+        uploadProgressEl.addEventListener('pointerdown', handleCancel);
+        uploadProgressEl.addEventListener('click', handleCancel);
         document.body.appendChild(uploadProgressEl);
       }
 
+      const files = Array.from(uploadBatches.values()).flatMap(batch => batch.files);
+      const bytesDone = files.reduce((sum, file) => sum + file.bytesDone, 0);
+      const bytesTotal = files.reduce((sum, file) => sum + file.size, 0);
```
niksis02: Cancelled files are hidden from the list but still counted in bytesTotal. If you cancel a large queued file while smaller ones upload, the widget stays at a low percentage and shows something like "0 B / 10 GB" until everything finishes. Could we exclude cancelled files from bytesDone and bytesTotal, and from the "N of M completed" count, so the totals match what's still being uploaded?

As you see from the screenshots, each file is 5mb, but after cancellation the progress bar still shows 10mb and `0 of 2 completed`

Image

Image

niksis02 (approved):

PR #2425, opened 2026-09-21
fix: reject DeleteObjects requests with more than 1000 keys

### LINKED ISSUE #2196: [Bug] - Fix DeleteObjects request limit (max 1000 objects)
**Describe the bug**
VersityGW currently does not enforce the maximum number of objects that can be deleted in a single DeleteObjects request. According to the AWS S3 specification, a DeleteObjects request can contain up to 1000 keys. If a client sends more than 1000 objects, Amazon S3 returns an error (InvalidRequest).

VersityGW either processes the request (which violates the specification) or returns an incorrect error, causing two Ceph s3‑tests to fail:
test_multi_object_delete_key_limit
test_multi_objectv2_delete_key_limit

**To Reproduce**
Start VersityGW with POSIX backend.
Create a bucket and upload 1001 objects.
Send a DeleteObjects request with 1001 objects in the XML body:
<Delete>
  <Object><Key>obj1</Key></Object>
  <Object><Key>obj2</Key></Object>
  ...
  <Object><Key>obj1001</Key></Object>
</Delete>
Actual response: HTTP 200 OK (the request is processed).
Expected response: HTTP 400 Bad Request with error code InvalidRequest.

Run the Ceph s3‑tests:
S3TEST_CONF=s3tests.conf pytest s3tests/functional/test_s3.py::test_multi_object_delete_key_limit -v
S3TEST_CONF=s3tests.conf pytest s3tests/functional/test_s3.py::test_multi_objectv2_delete_key_limit -v
Both tests fail with an assertion error.

**Expected behavior**
According to the AWS S3 API documentation for DeleteObjects: https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html
The request contains a list of up to 1000 keys that you want to delete.
If the request contains more than 1000 objects, Amazon S3 returns a 400 Bad Request with the error code InvalidRequest.
The server must reject such requests and not process them.

**Server Version**
./versitygw -version 
Version  : v1.5.0
Build    : fa789478
BuildTime: 2026-06-16_09:52:26

**Proposed fix**
Add a validation check in the DeleteObjects controller (s3api/controllers/bucket-post.go) to reject requests that contain more than maxDeleteObjects objects.

```diff
--- s3api/controllers/bucket-post.go.orig       2026-06-17 00:58:08.652431847 +0300
+++ s3api/controllers/bucket-post.go    2026-06-17 00:57:59.476417884 +0300
@@ -32,6 +32,13 @@
        "github.com/versity/versitygw/s3response"
 )
 
+const (
+       // maxDeleteObjects is the maximum number of objects that can be deleted in a single request.
+       // According to AWS S3 specification, the maximum is 1000 objects per DeleteObjects request.
+       // https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html
+       maxDeleteObjects = 1000
+)
+
 func (c S3ApiController) DeleteObjects(ctx *fiber.Ctx) (*Response, error) {
        bucket := ctx.Params("bucket")
        bypass := strings.EqualFold(ctx.Get("X-Amz-Bypass-Governance-Retention"), "true")
@@ -71,6 +78,21 @@
                }, s3err.GetAPIError(s3err.ErrInvalidRequest)
        }
 
+       // =====================================================================
+       // Validate the number of objects does not exceed the AWS S3 limit of 1000.
+       // See: https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html
+       // =====================================================================
+       if len(dObj.Objects) > maxDeleteObjects {
+               debuglogger.Logf("DeleteObjects request contains %d objects, exceeding the limit of %d",
+                       len(dObj.Objects), maxDeleteObjects)
+               return &Response{
+                       MetaOpts: &MetaOptions{
+                               BucketOwner: parsedAcl.Owner,
+                       },
+               }, s3err.GetAPIError(s3err.ErrInvalidRequest)
+       }
+       // =====================================================================
+
        err = auth.CheckObjectAccess(ctx.Context(), bucket, acct.Access, dObj.Objects, bypass, IsBucketPublic, c.be, false)
        if err != nil {
                return &Response{
```

**Test results**
After applying the patch, the Ceph s3‑tests pass:
$ S3TEST_CONF=s3tests.conf pytest s3tests/functional/test_s3.py::test_multi_object_delete_key_limit -v
...
s3tests/functional/test_s3.py::test_multi_object_delete_key_limit PASSED

$ S3TEST_CONF=s3tests.conf pytest s3tests/functional/test_s3.py::test_multi_objectv2_delete_key_limit -v
...
s3tests/functional/test_s3.py::test_multi_objectv2_delete_key_limit PASSED

**Additional context**
This fix aligns VersityGW with the AWS S3 specification and prevents clients from sending oversized DeleteObjects requests.
The same limit applies to both DeleteObjects and DeleteObjectsV2 (the test covers both, but the implementation is shared).

tunglambk:
Fixes #2196.

S3 caps one `DeleteObjects` request at 1000 keys and returns 400 `MalformedXML` for anything larger, as confirmed in review. The handler parsed the body and passed every key through authorization and on to the backend, so a 1001-key request was processed instead of being refused, which is what the two Ceph s3-tests in the report trip over.

The count is now checked right after the body is parsed, before authorization and before anything reaches the backend, so an over-limit batch can't be applied partially. The limit sits with the other request limits in `s3api/controllers/base.go` (`maxDeleteObjects = 1000`) and the response uses the existing `s3err.ErrMalformedXML`. Exactly 1000 keys still succeed, and an empty delete list keeps its current behavior (200 with an empty result).

`TestS3ApiController_DeleteObjects` covers 1001 keys (rejected, backend never called), exactly 1000 keys, and an empty list; the 1001-key case fails on `main` because the request reaches the backend. `DeleteObjects_key_limit` in `tests/integration/DeleteObjects.go` does the same against a live gateway.

Validation:
- `go test -race -tags=github -timeout 120s ./s3api/controllers/...` passes.
- The new `DeleteObjects_key_limit` integration case passes against a local posix gateway and fails on `main` with `expected MalformedXML, instead got nil`.
- `go test -tags=github ./...` fails the same `backend/posix` conditional-put and working-directory tests as unmodified `main`.
- `staticcheck ./s3api/controllers/... ./tests/integration/...` and `gofmt -s -l .` are clean.


tunglambk:
All the workflows are sitting at `action_required`, so a maintainer needs to approve them before they'll run. Could someone approve when convenient?

versity-github:

s3api/controllers/bucket-post.go
```diff
@@ -56,6 +56,19 @@ func (c S3ApiController) DeleteObjects(ctx fiber.Ctx) (*Response, error) {
...
+	// over-limit request before authorization and before anything reaches
+	// the backend, so a too-large batch can't be applied partially.
+	if len(dObj.Objects) > maxDeleteObjects {
+		debuglogger.Logf("delete objects: %d keys exceeds the limit of %d",
+			len(dObj.Objects), maxDeleteObjects)
+		return &Response{
+			MetaOpts: &MetaOptions{
+				BucketOwner: parsedAcl.Owner,
+			},
+		}, s3err.GetAPIError(s3err.ErrInvalidRequest)
```
niksis02: AWS S3 returns `MalformedXML` in this case.

niksis02 (changes requested):

s3api/controllers/bucket-post.go
tunglambk: Switched to `MalformedXML`; that's at the current head of the branch now.

niksis02 (approved):

benmcclelland (approved):

PR #2426, opened 2026-09-21
fix: log the client IP from a trusted proxy header when configured

### LINKED ISSUE #2243: [Bug] - Use x-real-ip or x-forwarded-for ip addresses in logs
**Describe the bug**
I use versitygw behind a reverse proxy. Versitygw log the ip of the reverse proxy, but not client ip address from x-forwarded-for or x-real-ip header.

**To Reproduce**
Run versitygw behind a reverse proxy and watch the logs.

**Expected behavior**
Versitygw should use ip from x-real-ip or x-forwarded-for header, if an option is set to use one or the other.

**Server Version**
1.6.0 on docker

**additional context**
It's possible to run 2 reverse proxies. For example a gateway and on the server. Then, x-forwarded-for header contain 2 ip addresses. Therefore, the first one should be used, because this is the client ip address.

tunglambk:
Fixes #2243.

Behind a reverse proxy the gateway logs the proxy's address, because the request log and the S3 and admin access logs all use `ctx.IP()`. This adds a `client-ip-header` option (`VGW_CLIENT_IP_HEADER`, `embedgw.Config.ClientIPHeader`) that takes the logged address from `X-Forwarded-For` or `X-Real-Ip`. It's off by default, so existing deployments log exactly what they logged before, and a header that's missing, empty or not a valid IP falls back to the socket address.

`X-Forwarded-For` is a chain, oldest hop first, so the leftmost entry is the client: the request through the two proxies in the report logs the client address, not the nearest proxy. `X-Real-Ip` carries a single address. Only the named header is read, so setting one ignores the other.

I left `ctx.IP()` alone on purpose. It's what `aws:SourceIp` evaluates in IAM policy conditions, and `auth/condition_context.go` already notes that configuring fiber's `ProxyHeader`/`TrustProxy` would make that client-controlled. The logged address is resolved separately and kept in a request local, so the change is limited to the request log and the file, webhook and admin access logs. `s3event` and the IAM API still report the socket address.

Validation:
- `go test -race -timeout 120s -tags=github ./s3api/... ./s3log/... ./embedgw/... ./internal/httpctx/...` passes.
- `go vet` and `staticcheck ./s3api/... ./s3log/... ./embedgw/... ./cmd/... ./internal/httpctx/...` are clean, and `gofmt -s -l .` reports nothing.
- `TestClientIP` fails against the old behaviour, reporting the socket address where the header cases expect `203.0.113.7`; the two-hop cases pin the leftmost-entry rule.
- `go test -p 1 ./...` passes except `backend/posix`, which fails the same way on unmodified `main` here with `invalid cross-device link`.


tunglambk:
All the workflows on this head are sitting at `action_required`, so a Versity member needs to approve the run before anything executes — the remote staticcheck and build results don't exist yet, only the local ones.

The change is opt-in and off by default (`--client-ip-header` / `VGW_CLIENT_IP_HEADER`, empty = current behaviour), so with the flag unset nothing about request handling or logging changes.

Could someone approve the run? Nothing else is blocking on my side.

versity-github:

PR #2427, opened 2026-09-21
fix: strip `aws-chunked` from stored `Content-Encoding` on streaming uploads

### LINKED ISSUE #2398: [Bug] - `Content-Encoding: aws-chunked` is persisted as object metadata and replayed on GET

niksis02:
Fixes #2398

Taking over #2400. An object uploaded with a trailing checksum came back carrying `Content-Encoding: aws-chunked`, and clients that decode content codings strictly then refused to read it although the stored bytes were intact.

The SDKs frame a PUT in `aws-chunked` to carry the trailing checksum and announce it with `Content-Encoding`, where the token describes the transport rather than the object, so S3 removes it before storing the header and keeps any other coding. The gateway stored the value verbatim and replayed it on every later GET and HEAD.

`ParseContentEncoding` now drops the token only when `x-amz-content-sha256` names a streaming payload type, so a plain PUT that sends `aws-chunked` still stores it as S3 does, and the authentication middleware rejects `aws-chunked` combined with `UNSIGNED-PAYLOAD` with an `InvalidArgument` naming `Content-Encoding`.

Covered by unit tests, REST tests, and integration tests for a framed upload, a plain PUT, and the rejected combination.


s3api/middlewares/authentication.go
```diff
@@ -132,6 +132,11 @@ func VerifyV4Signature(root RootUserConfig, iam auth.IAMService, region string,
 		if !streamBody && utils.IsStreamingPayload(hashPayload) {
 			return s3err.GetAPIError(s3err.ErrInvalidSHA256PayloadUsage)
 		}
+		// aws-chunked frames the body, so it contradicts an unsigned payload,
+		// which declares the body is sent as-is
+		if utils.IsUnsignedPaylod(hashPayload) && utils.HasAwsChunkedEncoding(ctx.Get("Content-Encoding")) {
```
benmcclelland, line 137: The new UNSIGNED-PAYLOAD rejection is bypassed for presigned uploads. https://github.com/versity/versitygw/blob/97c997cb76f998cc53b6728cba499379a0428870/s3api/middlewares/authentication.go#L51-L54 returns before the proposed guard once the presign middleware sets ContextKeyAuthenticated and presign-auth.go does exactly that and has no equivalent validation. A presigned PUT with X-Amz-Content-Sha256: UNSIGNED-PAYLOAD and Content-Encoding: aws-chunked will therefore succeed, and ParseContentEncoding will retain aws-chunked because the payload is not streaming. This recreates the invalid stored metadata path the PR intends to block. Suggest put the validation in shared request validation or add the same guard to presigned authentication.

benmcclelland:
needs a rebase to fix conflicts

benmcclelland (approved):

